PR #557, opened 2014-12-17
Standalone Mongo Orchestration

estolfo:
A first stab at making a Mongo Orchestration standalone resource available in our tests.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -90,15 +90,27 @@ def failing_delete_doc
                            { que: { field: 'test' } }
 end
 
-# Inititializes a basic scanned client to do an ismaster check.
+# Initializes a basic scanned client to do an ismaster check.
 #
 # @since 2.0.0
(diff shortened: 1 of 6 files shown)


spec/spec_helper.rb
```diff
@@ -89,12 +90,28 @@ def failing_delete_doc
...
+# Initializes a basic scanned client to do an ismaster check.
 #
 # @since 2.0.0
 def initialize_scanned_client!
   Mongo::Client.new([ '127.0.0.1:27017' ], database: TEST_DB)
 end
 
+def initialize_mo_standalone!(path = nil)
+  @standalone ||= MongoOrchestration.get(:standalone, path: path)
+end
```
estolfo: @durran What do you think about the fact that I only allow one @standalone instance? The way this is written, if you call #initialize_mo_standalone! more than once, it gives you the existing @standalone instance back if one exists. It doesn't check to see if the paths are identical.

PR #560, opened 2015-01-04
Add method to update MongoReplicaSetClient list of seeds

ebroder:
This is somewhat speculative, but would solve a problem we've recently run into. I'm happy to do polish work or cleanup as desired if the interface is viable.

This is useful when you know that the list of seeds on the connectionis no longer valid - for instance, if the old list of seeds were IP addresses referring to instances that have been rebuilt or decommissioned. Without some option like this, if all connections are closed (due to, e.g., a stepdown), the MongoDB client will attempt to connect to the old, decommissioned nodes, delaying recovery.

### DIFF
--- a/lib/mongo/mongo_replica_set_client.rb
+++ b/lib/mongo/mongo_replica_set_client.rb
@@ -176,6 +176,21 @@ module Mongo
       setup(opts.dup)
     end
 
+    # Replace the list of nodes to use as seeds in the event of a lost connection
+    #
+    # @param [Array<String>] new seeds
+    #
+    # @raise [MongoArgumentError] This is raised if the list of seeds is empty
+    def update_seeds!(seeds)
+      if seeds.length.zero?
+        raise MongoArgumentError, "Updating seeds requires at least one seed node."
+      end
+
+      @seeds = Support.normalize_seeds(seeds).freeze
+      connect(true)
+      nil
+    end
+
     def valid_opts
       super + REPL_SET_OPTS - CLIENT_ONLY_OPTS
     end


estolfo:
Hi @ebroder 
Thanks for suggesting this but I think we are going to turn down the PR.
Updating the seed list is not thread safe; this would cause problems if the client is in the midst of rediscovery when the user calls #update_seeds!

I would recommend creating a new client instead if you know your seed list has changed.

ebroder:
Thanks for the feedback! That said, I'm not entirely sure I understand why it wouldn't be threadsafe - shouldn't both assignment to `@seeds` and access to `@seeds` (in the handful of places that happens) be atomic?

In any case, if you'd prefer being explicit, it seems easy enough and harmless to grab `@connect_mutex` for the assignment call (given the relative infrequency of expected calls to `update_seeds!`, it doesn't seem like it will create a lot of contention).

The specific reason I'm interested in this API has to do with efficiency of handling seed list changes. In particular, given an existing, established connection, it's much more efficient to modify the existing seed list than it is to throw out the connection, create a new one, and go through the handshake process again.

behackett:
The driver should only use the seed list to initially discover one available replica set member. From that point on it should use the hosts/passives/arbiters lists returned by an ismaster call. The seed list changing shouldn't matter to a live client since any changes to the replset config in the cluster should be detected by the client during periodic monitoring (subsequent calls to ismaster).

estolfo:
This wouldn't be threadsafe because in a multi-threaded environment, one thread could be reading from the seed list while the user calls #update_seeds! You would end up with errors similar to these:

https://jira.mongodb.org/browse/RUBY-578
https://jira.mongodb.org/browse/RUBY-736

I understand that it's more efficient to reuse the same client but changes to the seed list and refreshes are usually infrequent. 

@behackett We have corrected this behavior in implementing SDAM in driver 2.0.

estolfo:
Hi again @ebroder 

I wanted to follow up with you again on this. You mentioned that this was speculative but I was wondering if you saw a measurable delay when the reconnecting?

If we look at the code, the pool manager doesn't necessarily go back to the original list of seeds provided by the user. When reconnecting, the pool manager updates the list of seeds here 

https://github.com/mongodb/mongo-ruby-driver/blob/1.x-stable/lib/mongo/connection/pool_manager.rb#L73

The method, #discovered_seeds gets a list of host-port pairs from the servers to which it had a successful connection.

https://github.com/mongodb/mongo-ruby-driver/blob/1.x-stable/lib/mongo/connection/pool_manager.rb#L276-L278

Can you tell me if you're seeing a significant delay when connections are closed and the client reconnects? If so, this might be indicative of another issue. I'm going to close this PR but encourage you to open a JIRA ticket under the RUBY project if you believe there is a bug or improvement needed in the driver.

https://jira.mongodb.org/browse/RUBY

Thanks

PR #571, opened 2015-01-23
RUBY-846: Rename first to not conflict with Enumerable

durran:
Part of the changes for ongoing Mongoid integration.


arthurnn:
:+1:

PR #573, opened 2015-01-23
Fix client instantiation in readme

kyzh:
I'm not sure why, so I might be wrong when reporting this as an issue.

Using `client = Mongo::Client.new`
Gives the following error `NameError: uninitialized constant Mongo::Client`

The documentation seems to indicate `Class: Mongo::MongoClient`
[Documentation source](http://api.mongodb.org/ruby/current/Mongo/MongoClient.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Here is a quick example of basic usage for the Ruby driver:
 require 'mongo'
 
 # connecting to the database
-client = Mongo::Client.new # defaults to localhost:27017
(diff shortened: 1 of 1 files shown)


arthurnn:
`Mongo::Client.new` should be valid. you are probably using a old version of the driver. 
That said, the docs are right.

durran:
You are reading the README for the master branch, which is the upcoming 2.0.0 release... Switch to the 1.x-stable branch for a README for what is currently released.

PR #574, opened 2015-01-30
RUBY-846: Mongoid Integration/Monitor Refactor (WIP)

durran:


arthurnn:
Just a heads-up .. weak-ref dont work on 1.9.3 , there are A LOT of bugs around it, so if we wanna use weak-ref we need to set ruby version to 2.0.0+

durran:
I removed all weak refs, was just a short experiment.

PR #575, opened 2015-02-03
RUBY-853 Use moving average round trip time

estolfo:
..as per server selection spec


lib/mongo/server/monitor.rb
```diff
@@ -128,8 +134,16 @@ def stop!
 
       private
 
-      def calculate_round_trip_time(start)
-        Time.now - start
+      def average_round_trip_time(start)
```
estolfo: No, I didn't think that through. How many would be appropriate to keep around?

PR #579, opened 2015-02-17
RUBY-854 Scan cluster if no server is found

estolfo:
- why was there a todo to test for compact on this line?


durran:
Looks good, travis failure was non-related.

PR #581, opened 2015-03-04
Monstrous Bulk write refactor

estolfo:


durran:
Net 730 less lines of code? Woo hoo! :+1:

estolfo:
I know, right?
There's even more deletion to come \o/

PR #585, opened 2015-03-16
Avoid sharing sockets between forked processes

eugeneius:
Support for detecting when a socket was established in a different process and automatically discarding it was added to this driver in 997d9b8ae2f5f391ef23dc4d3792e7d863675eb8, and refined in a393557eeb8c1a9a049f4d289be51e50579ebf40.

Instances of the Node class maintain their own sockets for refreshing replica set data; they need to be protected from reuse in the same way.

### DIFF
--- a/lib/mongo/connection/node.rb
+++ b/lib/mongo/connection/node.rb
@@ -73,16 +73,6 @@ module Mongo
       end
     end
 
-    def usable_socket
-      if @socket && @socket.pid != Process.pid
-        @socket.close
-        @socket = nil
-        connect
-      else
-        @socket
-      end
-    end
-
     # This should only be called within a mutex
     def close
       if @socket && !@socket.closed?
@@ -237,6 +227,16 @@ module Mongo
     end
 
     private
+
+    def usable_socket
+      if @socket && @socket.pid != Process.pid
+        @socket.close
+        @socket = nil
+        connect
+      else
+        @socket
+      end
+    end
 
     def update_max_sizes
       @max_bson_size = config['maxBsonObjectSize'] || DEFAULT_MAX_BSON_SIZE
--- a/lib/mongo/connection/node.rb
+++ b/lib/mongo/connection/node.rb
@@ -73,7 +73,7 @@ module Mongo
       end
     end
 
-    def socket
+    def usable_socket
       if @socket && @socket.pid != Process.pid
         @socket.close
         @socket = nil
@@ -98,7 +98,7 @@ module Mongo
 
     def active?
       begin
-        result = @client['admin'].command({:ping => 1}, :socket => socket)
+        result = @client['admin'].command({:ping => 1}, :socket => usable_socket)
       rescue OperationFailure, SocketError, SystemCallError, IOError
         return nil
       end
@@ -117,10 +117,10 @@ module Mongo
 
           if @client.connect_timeout
             Timeout::timeout(@client.connect_timeout, OperationTimeout) do
-              @config = @client['admin'].command({:ismaster => 1}, :socket => socket)
+              @config = @client['admin'].command({:ismaster => 1}, :socket => usable_socket)
             end
           else
-            @config = @client['admin'].command({:ismaster => 1}, :socket => socket)
+            @config = @client['admin'].command({:ismaster => 1}, :socket => usable_socket)
           end
 
           update_max_sizes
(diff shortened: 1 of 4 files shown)


estolfo:
Hi @eugeneius 

We can't use Monitor because we support ruby 1.8.7 in the 1.x series of the driver.

Is there a particular test case you can share that recreates the problem you are trying to solve? That way, we can see how to resolve it without Monitor.

Also, we have released an rc of a rewrite of the ruby driver. It's available on RubyGems as 2.0.0.rc. 
https://rubygems.org/gems/mongo/versions/2.0.0.rc
You might want to try it out to see if the issues you are having are resolved.

Thanks
Emily

eugeneius:
The test suite passes on 1.8.7 - is there a known issue with Monitor which means it can't be used? I replaced the Mutex with a Monitor because with this patch `connect` can be called inside `set_config` and both need to hold the lock; Monitor is reentrant, while Mutex isn't. It should be possible to make this patch work with a Mutex, but it might involve rearranging the class a bit.

I opened a ticket describing the issue I'm seeing: https://jira.mongodb.org/browse/RUBY-877

It only happens a few dozen times a day in production, and I haven't been able to reproduce it locally.

Even if `2.0.0.rc` fixes the problem I won't be able to use it in my application, because I'm using MongoMapper which relies on `1.x`.

lib/mongo/connection/node.rb
```diff
@@ -73,6 +73,16 @@ def connect
       end
     end
 
+    def socket
+      if @socket && @socket.pid != Process.pid
```
durran: If the process is not forked, `@socket.pid` will *always* return `nil`, which will cause the socket to be closed and reopened on *every* request, thus causing a significant negative impact on performace.

eugeneius: The `pid` property is assigned when a socket is created:

https://github.com/mongodb/mongo-ruby-driver/blob/1.12.0/lib/mongo/connection/socket/ssl_socket.rb#L28
https://github.com/mongodb/mongo-ruby-driver/blob/1.12.0/lib/mongo/connection/socket/tcp_socket.rb#L27
https://github.com/mongodb/mongo-ruby-driver/blob/1.12.0/lib/mongo/connection/socket/unix_socket.rb#L28

This is the same logic being used already in the connection pool code:

https://github.com/mongodb/mongo-ruby-driver/blob/1.12.0/lib/mongo/connection/pool.rb#L224

durran: Sorry, missed that part, was thinking of `IO#pid` in core Ruby. :)

estolfo:
My mistake, I did a quick test with 1.8.7 and you have to require 'monitor' in order to use it, whereas you don't with >= 1.9.3

I'm going to do a bit more testing but this patch seems reasonable. Thanks for opening the ticket and taking the time to investigate.

estolfo:
Ideally, we would have a test for this scenario. If you can write one, that would be great. Otherwise, I'll add one.

estolfo:
Hi Eugene

Thanks again for your contribution.

I've taken a closer look at your PR and have a few changes to propose. They can be found in this [branch](https://github.com/estolfo/mongo-ruby-driver/commits/eugeneius-detect_shared_sockets) and are being tested in jenkins [here](https://jenkins.10gen.com/job/mongo-ruby-driver-1.x-pr-sockets).

My suggestions are:

1. Changing the name of the socket method to #usable_socket. This makes it more explicit that this method needs to be called when the node's @socket is used.

2. Making the #usable_socket method private because with your change, called node.socket could mutate the node instance; the method gets the node lock and potentially calls #connect on the node. I think we should keep the accessor method #socket immutable on the node class.

3. Adding a test that gets the socket associated with a node, changes its pid, then triggers an ismaster command. The expected outcome is that the node gets a new socket because the old socket's pid does not match the process's pid.

Let me know if you'd like to update the PR yourself so I can accept it or if you want me to merge it, then make these changes on top.

Thanks again

Emily

eugeneius:
Thanks for the code review @estolfo - I've cherry-picked your changes into this branch.

estolfo:
Perfect, thanks @eugeneius !

estolfo:
oh, @eugeneius actually, can you rebase because I commit today?  https://github.com/mongodb/mongo-ruby-driver/commit/561e1a632bdcf2f7b64f9f19598abccab86ae551
Thanks

eugeneius:
Done!

estolfo:
hey @eugeneius !
ruby driver version 1.12.1 has been released with your PR. Feel free to try it out!
Thanks again

eugeneius:
Thanks @estolfo, we're upgrading now :)

PR #586, opened 2015-03-20
RUBY-879: Implement minHeartbeatFrequency at 500ms

durran:
https://jira.mongodb.org/browse/RUBY-879

### DIFF
--- a/lib/mongo/server/monitor.rb
+++ b/lib/mongo/server/monitor.rb
@@ -29,6 +29,12 @@ module Mongo
       # @since 2.0.0
       HEARTBEAT_FREQUENCY = 10.freeze
 
+      # The minimum time between forced server scans. Is
+      # minHeartbeatFrequencyMS in the SDAM spec.
+      #
+      # @since 2.0.0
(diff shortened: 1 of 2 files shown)


estolfo:
lgtm, I'm glad we're all on the same page re this now = )

ajdavis:
LGTM.

lib/mongo/server/monitor.rb
```diff
@@ -29,6 +29,12 @@ class Monitor
       # @since 2.0.0
       HEARTBEAT_FREQUENCY = 10.freeze
 
+      # The minimum time between forced server scans. Is
+      # minHeartbeatFrequencyMS in the SDAM spec.
+      #
+      # @since 2.0.0
+      MIN_SCAN_FREQUENCY = 0.5.freeze
```
arthurnn: why are you `freeze` this constant?

durran: Just habit - doesn't do anything on numbers but doesn't have any negative effects either.

PR #594, opened 2015-04-02
Test on Ruby 2.2.1 and auto test on other patched versions

nicolasleger:


.travis.yml
```diff
@@ -2,9 +2,9 @@ language: ruby
 
 rvm:
   - 1.9.3
-  - 2.0.0
-  - 2.1.5
-  - 2.2.0
+  - 2.0
```
arthurnn: 2.0.x still has the patch levels, so I think we can only do this for 2.1 and 2.2

nicolasleger: Yes you're right !

nicolasleger:
@arthurnn I put back Ruby 2.0.0.

PR #603, opened 2015-04-10
Any client :ssl* options created SSL connection... even :ssl => false

wormzer:
:ssl => false was triggering SSL connections


wormzer:
https://jira.mongodb.org/browse/RUBY-896

estolfo:
Thanks @wormzer but this is already fixed and will be in the next release.
Here is the commit:
https://github.com/mongodb/mongo-ruby-driver/commit/1b66b0ed9323de5af04d7d30c01158bf64f61085

PR #604, opened 2015-04-15
Make sure that Exception#message is not hidden

leifg:
Whenever the Mongo::URI::Invalid error is risen it is not possible to call its `message`:

    NoMethodError: private method `message' called for #<Mongo::URI::Invalid:0x7e7fc4bd>

I opted for a class method because it does not rely on any instance
variables.

### DIFF
--- a/lib/mongo/uri.rb
+++ b/lib/mongo/uri.rb
@@ -230,12 +230,12 @@ module Mongo
       #
       # @since 2.0.0
       def initialize(uri)
-        super(message(uri))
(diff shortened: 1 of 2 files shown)


estolfo:
Hi @leifg 
Thanks for opening this pull request. These error classes shouldn't actually be within the URI class but we missed moving them during a refactor. I've opened a separate pull request to fix this and match the structure of our Errors design so will close this PR. Thanks again!
https://github.com/mongodb/mongo-ruby-driver/pull/605

PR #606, opened 2015-04-20
RUBY-885: Add DBRef Support

durran:


estolfo:
looks good

jeromegn:
Shouldn't that be in https://github.com/mongodb/bson-ruby?

durran:
@jeromegn No, `DBRef` is not part of the BSON specification. http://bsonspec.org/spec.html

PR #609, opened 2015-04-23
RUBY-900 Don't add discovered servers to the cluster with Single topology

estolfo:


lib/mongo/cluster.rb
```diff
@@ -66,7 +66,7 @@ def ==(other)
     # @since 2.0.0
     def add(host)
       address = Address.new(host)
-      unless addresses.include?(address)
+      unless addresses.include?(address) || (@topology.single? && address.seed != @topology.seed)
```
estolfo: @durran This line can probably be improved, let me know if you have any other ideas.

jeromegn:
I believe that would fix it!

estolfo:
cool, thanks so much for helping out @jeromegn !

lib/mongo/cluster.rb
durran: I would probable change the logic to to make it a bit less confusing as well as extracting a method on the second part...

```ruby
if !addresses.include?(address) || direct_connection?(address)
  #...
end
```

estolfo: done, thanks

PR #610, opened 2015-04-24
RUBY-897 Print first 250 chars of query selector when logging and rescue inspect error

estolfo:
I guessed that 250 characters was a decent limit on the selector field when logging an error with a query but am open to suggestions for that limit to be decreased or increased.


durran:
Seems reasonable to me.

PR #619, opened 2015-05-15
RUBY-909 Add #parallel_scan back onto Collection::View

estolfo:


linjer:
I'm trying this out and it looks like the filter that produces the view isn't properly respected.  E.g.,

```Ruby
mongo[:users].find(firstName: 'Jerry').parallel_scan(8).each do |cursor|
  cursor.each do |u|
    u[:firstName] # This shows firstName of ALL users (i.e. no filter applied)
  end
end
```

linjer:
Followup: This was raised in the mongodb-ruby-driver JIRA [here](https://jira.mongodb.org/browse/RUBY-995) and has been resolved by being moved to `Collection`.

PR #620, opened 2015-05-18
RUBY-914 Correct special query selector format used with mongos

estolfo:


lib/mongo/operation/read_preferrable.rb
```diff
@@ -24,7 +24,8 @@ module ReadPreferrable
 
       def update_selector(context)
         if context.mongos? && read_pref = read.to_mongos
-          selector.merge(:$readPreference => read_pref)
+          { :$query => selector,
```
behackett: What if the selector already has $query in it because of some other $operator ($orderby, $maxTimeMS, etc.)?

estolfo: good point, thanks

PR #631, opened 2015-06-08
RUBY-934 Check that > 0 bytes was read from the socket or the requested length was 0

estolfo:


lib/mongo/socket.rb
```diff
@@ -123,6 +123,7 @@ def initialize(family)
     def read(length)
       handle_errors do
         data = read_from_socket(length)
+        raise IOError unless data.length == length
```
durran: This should raise the `IOError` if `data.length == 0`. There is the possibility that some data is read but not all of it, which is the reason for the loop below. If that code path isn't ever executed, which is the case with the current check, then we have the chance to be reading partial messages.

estolfo: I found that the auth tests were failing if I checked for length 0. There is also a note in the method description saying 
https://github.com/estolfo/mongo-ruby-driver/blob/RUBY-934-socket/lib/mongo/socket.rb#L111

I'm wondering if the auth tests (scram/cr/x509) tests are failing because the socket is mocked though.

durran: The docs are wrong in that case then and should be changed... As for the failures... Probably to do with the mocking... I can have a look as well.

estolfo: No worries, I'll figure it out = )

estolfo: The reason the auth tests fail is because some authentication protocols receive binary data of length 0 at the end of the conversation. This is noted in the auth specs:

```
S: {conversationId: 1, payload: BinData(0,""), done: true, ok: 1}
Note
There is an extra round trip due to an implementation decision on the server. This is accomplished by sending no bytes back to the server for what is effectively a no-op.
```

With this PR, an error would be raised when this empty binary data is attempted to be read off the socket.
I'll add a check if (data.length == 0 && length != 0)
In other words, I'll check if you read data of length 0 and explicitly asked to read data of length 0.

durran: Excellent! Looks good.

durran:
:+1:

lib/mongo/socket.rb
```diff
@@ -123,6 +123,7 @@ def initialize(family)
     def read(length)
       handle_errors do
         data = read_from_socket(length)
+        raise IOError unless (data.length > 0 || length == 0)
         while data.length < length
           data << read_from_socket(length - data.length)
```
behackett: You can have the same problem here.

estolfo: thanks

behackett:
LGTM

ablohowiak-fanhattan:
Thanks!!

PR #632, opened 2015-06-09
RUBY-945 Add #database_names and #list_databases methods to Client and #list_collections method to Database

estolfo:


lib/mongo/client.rb
```diff
@@ -223,6 +223,20 @@ def write_concern
...
+    # Get the names of all databases.
+    #
+    # @example Get the database names.
+    #   client.database_names
+    #
+    # @return [ Array<String> ] The names of the databases.
+    #
+    # @since 2.0.5
+    def database_names
+      Database.new(self, Database::ADMIN, options).command(
```
durran: Instead of constructing a new `Database` what do you think about consuming our own API and using #use here? Ex:

```ruby
use(Database::ADMIN).command(listDatabases: 1).first['databases'].collect{ |info| info['name'] }
```

estolfo: yes, that's better, thanks!

PR #637, opened 2015-06-12
RUBY-932 Update Server Discovery and Monitoring code and tests

estolfo:


lib/mongo/server/description.rb
```diff
@@ -445,6 +445,60 @@ def unknown!
...
+
+      # Is this description from the given server.
+      #
+      # @example Check if the description is from a given server.
+      #   description.self?(server)
+      #
+      # @return [ true, false ] If the description is from the server.
+      #
+      # @since 2.0.6
+      def self?(server)
```
durran: I might change this name to something else, given that `self` is reserved in Ruby and this might be a bit confusing.

estolfo: Changed it to Description#me?(server)

durran:
:+1:

lib/mongo/cluster.rb
```diff
@@ -89,14 +86,15 @@ def add(host)
     # @since 2.0.0
     def initialize(seeds, options = {})
       @addresses = []
+      @addresses_update = Mutex.new
       @servers = []
+      @servers_update = Mutex.new
```
ajdavis: I'd just put one big lock around Cluster and do all non-I/O operations (both reads and updates to all data structures) while holding the lock. If you do I/O, drop the lock, do your operation without touching any Cluster data structures, then reacquire the lock in order to update your data structures according to what you discovered from doing the I/O. So, no need to distinguish between these two mutexes, just lock the whole Cluster whenever you touch it.

lib/mongo/cluster.rb
```diff
@@ -153,9 +151,12 @@ def elect_primary!(description)
...
       log_debug([ "#{host} being removed from the cluster." ])
       address = Address.new(host)
-      removed_servers = @servers.reject!{ |server| server.address == address }
+      removed_servers = @servers.select { |s| s.address == address }
+      @servers_update.synchronize { @servers = @servers - removed_servers }
       removed_servers.each{ |server| server.disconnect! } if removed_servers
-      addresses.reject!{ |addr| addr == address }
+      @addresses_update.synchronize do
+        @addresses.reject!{ |addr| addr == address }
+      end
```
ajdavis: Is it ok that another thread can see addresses and servers go out of sync between these statements? If you just use one lock you can avoid answering that question. =)

lib/mongo/cluster.rb
```diff
@@ -213,5 +256,17 @@ def direct_connection?(address)
     def addition_allowed?(address)
       !@topology.single? || direct_connection?(address)
     end
+
+    def servers_list
+      @servers_update.synchronize do
+        @servers
+      end
```
ajdavis: Does this return a copy of the list? Doesn't it need to return a copy in order to avoid overlapping with a thread that's updating the servers while holding the servers_update lock? E.g. it seems like if one thread is in scan! and slowly iterating the list while another thread is in remove_hosts, removing servers while holding the lock, you can still get a runtime err.

lib/mongo/cluster.rb
estolfo: Thanks for your review.
I'm not going to take this approach because there is IO in a bunch of places that might not be obvious (like in the instantiation of an Address, for example) and the code would get messy if I was frequently grabbing and releasing the lock in between lines of code in Cluster.

I have taken your suggestion to only use one lock for both the servers and addresses list.

You're also right about having to copy the list. I've added that in as well.

Thanks again!

ajdavis: Glad I could help!

PR #638, opened 2015-06-15
added ruby-head

ferdinandrosario:
Added ruby-head  under allowed-failure


durran:
Thanks, but I purposely removed the allowed failures in the matrix since our matrix is already huge and it's taking a lot of resources from travis - trying to be as nice to them as possible. We already use head in our internal jenkins server so this is covered.

ferdinandrosario:
Nice. :+1:  :heart:

PR #642, opened 2015-06-24
Correct wire version range checks

estolfo:


spec/mongo/server/description/features_spec.rb
```diff
@@ -15,11 +15,20 @@
       end
     end
 
-    context 'when the server wire version range is higher' do
```
durran: I'd also keep this test and change the expectation so that an exception is not expected.

PR #644, opened 2015-06-26
RUBY-912 Use electionId from server descriptions to detect stale primaries

estolfo:


durran:
Looks good.

lib/mongo/cluster/topology/replica_set.rb
```diff
@@ -60,11 +60,14 @@ def display_name
...
-              if server.primary? && server.address != description.address
-                server.description.unknown!
+            unless detect_stale_primary!(description)
+              log_debug([ "Server #{description.address.to_s} elected as primary in #{replica_set_name}." ])
+              servers.each do |server|
+                if server.primary? && server.address != description.address
+                  server.description.unknown!
+                end
               end
+              @max_election_id = description.election_id
```
ajdavis: I think this is different from the spec, which says maxElectionId is "The largest electionId ever reported by a primary." In the [spec's pseudocode](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#user-content-updatersfromprimary) it only updates maxElectionId if the primary being elected has an electionId. Here, though, I think you'll set maxElectionId to null if a primary is elected with no electionId. Let me write up a test that shows what I mean.

PR #653, opened 2015-07-08
RUBY-883 Refactor Operations

estolfo:


lib/mongo/operation/write/create_index.rb
```diff
@@ -42,37 +42,14 @@ module Write
       #
       # @since 2.0.0
       class CreateIndex
-        include Executable
+        include GLEable
+        include WriteCommandEnablable
```
durran: Maybe for grammatical purposes `WriteCommandEnabled` might be more sane here. :)

lib/mongo/operation/write/gleable.rb
```diff
@@ -0,0 +1,47 @@
...
+# limitations under the License.
+
+module Mongo
+  module Operation
+    module Write
+
+      # This module contains common functionality for sending a GetLastError message.
+      #
+      # @since 2.1.0
+      module GLEable
```
durran: Same as above, perhaps `GLEEnabled`?

lib/mongo/operation/write/create_index.rb
estolfo: I originally had it as that but then changed it because I thought I was breaking Ruby rules. But I'd rather break Ruby rules than grammar rules.... : )

lib/mongo/operation/write/gleable.rb
estolfo: I'll just call it GLE - which is what I had originally but then changed for reasons written above.

PR #656, opened 2015-07-13
RUBY-965 add #exists? method on User::View

estolfo:


durran:
Pull request looks good, the failure is not related.

estolfo:
Do you agree, @durran that the method should be to get user info, not specifically #exists?
The driver user can then wrap the #info method in another method and get whatever they want out of the doc returned.

durran:
Yes I think it's better this way. Java and C# have helpers for the `userinfo` command as well.

PR #659, opened 2015-07-22
RUBY-983 Close leaked sockets

estolfo:


agis-:
I've tried the same process described in [RUBY-983](https://jira.mongodb.org/browse/RUBY-983) and unfortunately the problem persists.

estolfo:
hmm, ok. Thanks, I'll take another look.

PR #660, opened 2015-07-22
RUBY-986 Fallback to IPv4 if IPv6 isn't supported

agis-:
Prior to this change, if the provided server address resolved to IPv6
but the mongo server wasn't configured for IPv6, the client would
continuously attempt to connect  to the IPv6 address until the
server_selection_timeout was hit and never fallback to IPv4.

This fixes the behavior to try all the different protocol families in
the list.

Fixes https://jira.mongodb.org/browse/RUBY-986.

### DIFF
--- a/lib/mongo/address.rb
+++ b/lib/mongo/address.rb
@@ -146,12 +146,12 @@ module Mongo
     def initialize_resolver!(timeout, ssl_options)
       family = (host == 'localhost') ? ::Socket::AF_INET : ::Socket::AF_UNSPEC
       error = nil
-      ::Socket.getaddrinfo(host, nil, family, ::Socket::SOCK_STREAM).detect do |info|
+      ::Socket.getaddrinfo(host, nil, family, ::Socket::SOCK_STREAM).each do |info|
         begin
-          return FAMILY_MAP[info[4]].new(info[3], port, host).tap do |res|
-            res.socket(timeout, ssl_options).connect!
-          end
-        rescue IOError, SystemCallError => e
+          res = FAMILY_MAP[info[4]].new(info[3], port, host)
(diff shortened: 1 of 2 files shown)


estolfo:
Thanks @agis- 
Is there any chance you could provide a test as well? Thanks

agis-:
@estolfo Sure, will add them in a bit.

agis-:
@estolfo added tests!

estolfo:
Thanks!

estolfo:
Looks good, thanks so much!

This is minor, but would you mind changing the test a little to match our formatting for consistency? Putting the 'allow' portion in a before block and check that address.socket returns an object would be great:

```ruby
  describe "#socket" do

    context 'when providing a DNS entry that resolves to both IPv6 and IPv4' do

      let(:address) do
        described_class.new('localhost:27017')
      end

      before do
        allow(::Socket).to receive(:getaddrinfo).and_return(
          [ ["AF_INET6", 0, "::1", "::1", ::Socket::AF_INET6, 1, 6],
            ["AF_INET", 0, "127.0.0.1", "127.0.0.1", ::Socket::AF_INET, 1, 6]]
        )
      end

      it "attempts to use IPv6 and fallbacks to IPv4" do
        expect(address.socket(0.0)).not_to be_nil
      end
    end
  end
```

agis-:
No problem, updated. :)

estolfo:
If you could just rebase against master so we can make sure tests pass, I can merge it. Thanks!

agis-:
Rebased!

estolfo:
The travis failures are unrelated. I'll merge this now, thanks again!

PR #663, opened 2015-07-23
Don't force IPv4 on localhost connections

agis-:
An IPv6 connection to localhost is perfectly valid and should be possible.

After this change, connections are protocol-agnostic *regardless* of the
given host.

### DIFF
--- a/lib/mongo/address.rb
+++ b/lib/mongo/address.rb
@@ -144,9 +144,8 @@ module Mongo
     private
 
     def initialize_resolver!(timeout, ssl_options)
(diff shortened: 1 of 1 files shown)


behackett:
I believe this check is related to https://jira.mongodb.org/browse/PYTHON-356. I don't know how much use Ruby gets on Windows, but this was an annoying issue for PyMongo users. Note that instead of using 'localhost' you should be able to use ::1.

agis-:
@behackett Ah I see. Couldn't we check for the platform then and only skip it on Windows?

PR #665, opened 2015-07-28
RUBY-951 Ensure that driver conforms to Connection String specification

estolfo:


durran:
Yeah I figured this was going to be a pain after looking at it more with just regexes... If the tests all pass then looks good to me - quite thorough.

estolfo:
Yes, the tests were not trivial to get passing. 80% passed with the regex implementation, but to get the remaining 20% to pass, a rewrite was in order.
Thanks @jmikola for your thoroughness = )

jmikola:
> Thanks @jmikola for your thoroughness = )

PR #667, opened 2015-08-04
Found typo in `conversation.rb`

saghm:
I was looking at the implementation of the SCRAM auth for reference in writing the Rust driver authentication when I noticed this typo.


durran:
@saghm Mind rebasing against master please to get rid of the merge commit? Thanks!

PR #668, opened 2015-08-05
RUBY-981: Add support for Unix domain sockets

agis-:
This is an initial take on supporting Unix domain sockets.

Although with this patch the driver seems to work fine with unix sockets, it's far from complete. However I can go on and work on it if it's on the right track.

What do you think?

### DIFF
--- a/lib/mongo/socket.rb
+++ b/lib/mongo/socket.rb
@@ -173,10 +173,13 @@ module Mongo
     end
 
     def set_socket_options(sock)
-      encoded_timeout = [ timeout, 0 ].pack(TIMEOUT_PACK)
       sock.set_encoding(BSON::BINARY)
-      sock.setsockopt(SOL_SOCKET, SO_RCVTIMEO, encoded_timeout)
-      sock.setsockopt(SOL_SOCKET, SO_SNDTIMEO, encoded_timeout)
+
+      unless sock.is_a?(UNIXSocket) && BSON::Environment.jruby?
+        encoded_timeout = [ timeout, 0 ].pack(TIMEOUT_PACK)
+        sock.setsockopt(SOL_SOCKET, SO_RCVTIMEO, encoded_timeout)
+        sock.setsockopt(SOL_SOCKET, SO_SNDTIMEO, encoded_timeout)
+      end
     end
 
     def handle_errors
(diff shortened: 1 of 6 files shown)


lib/mongo/socket/unix.rb
```diff
@@ -56,7 +53,7 @@ def connect!
       # @since 2.0.0
       def initialize(path, timeout, family)
         @path, @timeout = path, timeout
-        super(family)
+        super(family, path)
       end
```
estolfo: Thanks so much for looking into this.

Your changes look good. My own suggestion is that we make this method, Socket::Unix#initialize look like this

```ruby
def initialize(path, timeout, family)
  @path, @timeout = path, timeout
  @socket = ::Socket.unix(path)
  set_socket_options(@socket)
end
```

instead of calling super with an extra argument. That means the changes in socket.rb wouldn't be necessary. What do you think?

estolfo:
Feel free to add tests too and to resolve the failures in tests where we connect using a fake unix socket path = )

agis-:
Hello @estolfo!

I've added some basic tests. There was an issue with JRuby not supporting the options being set on MRI in `#set_socket_options`; I'm not very happy with the solution.

WDYT?

estolfo:
Great - thanks also for getting JRuby issues resolved. I'm offline for the next week but will get back to you and merge this in as soon as I get back.
Thanks again!

lib/mongo/socket.rb
```diff
@@ -173,10 +173,13 @@ def read_from_socket(length)
     end
 
     def set_socket_options(sock)
-      encoded_timeout = [ timeout, 0 ].pack(TIMEOUT_PACK)
       sock.set_encoding(BSON::BINARY)
-      sock.setsockopt(SOL_SOCKET, SO_RCVTIMEO, encoded_timeout)
-      sock.setsockopt(SOL_SOCKET, SO_SNDTIMEO, encoded_timeout)
+
+      unless sock.is_a?(UNIXSocket) && RUBY_ENGINE == "jruby"
```
durran: Instead of `RUBY_ENGINE == "jruby"` can you use `BSON::Environment.jruby?`

agis-: done.

durran:
I ran the entire suite on this branch switching the `ADDRESSES` env variable in the test suite to use `/tmp/mongodb-27017.sock` instead. All tests pass as well with that.

PR #671, opened 2015-08-15
RUBY-999: Use constant time hash comparison

durran:


lib/mongo/auth/scram/conversation.rb
```diff
@@ -429,9 +429,15 @@ def xor(first, second)
           first.bytes.zip(second.bytes).map{ |(a,b)| (a ^ b).chr }.join('')
         end
 
+        def secure_compare(a, b)
```
behackett: Using the word "secure" is too strong of a claim (the original attempt to implement this for cpython made the same claim and was rejected). compare_digest seems better.

lib/mongo/auth/scram/conversation.rb
```diff
@@ -429,9 +429,15 @@ def xor(first, second)
...
+          check = a.bytesize ^ b.bytesize
+          a.bytes.zip(b.bytes){ |x, y| check |= x ^ y.to_i }
```
behackett: What does zip do if the string lengths aren't equal? MIght be better to take the hash from the server and compare it to itself in that case.

durran: If they lengths are not equal then null bytes are used to pad the shorter to the length of the longer.

lib/mongo/auth/scram/conversation.rb
durran: Renamed.

behackett:
LGTM

PR #674, opened 2015-08-20
RUBY-911 RUBY-963 Add Collection#with for getting a new collection with different read preference / write concern

estolfo:


lib/mongo/collection.rb
```diff
@@ -75,6 +75,55 @@ def initialize(database, name, options = {})
...
+    #
+    # @example Get a collection with changed write concern.
+    #   collection.with(:write => { w:  3 })
+
+    # @param [ Hash ] new_options The new options to use.
+    #
+    # @return [ Mongo::Collection ] A new collection instance.
+    #
+    # @since 2.1.0
+    def with(new_options)
```
durran: why not just:

```ruby
Collection.new(database, name, options.merge(new_options)
```

estolfo: I was limiting the options that could be changed to :write and :read but perhaps we should allow any options to be changed?

Also, we need to make sure to merge the client options to get the server_selection_timeout setting.

durran: You could still merge the client options in the single line as well. I think if we provide a `#with` method just like `Mongo::Client` it needs to accept any options, not just read and write options.

durran: Thinking about it a bit more, we probably don't want options to be able to be passed here that would change the cluster configuration, but we shouldn't silently ignore them... I think we should probably raise an error if anything outside of `read`, `write` or anything else that makes sense to pass here is provided.

estolfo: Raising an error sounds like the best solution.

For simplicity, I think I'm going to just merge in the client's server_selection_timeout option to the read preference. It'll be more explicit that way.

durran: Sounds good.

estolfo: I've updated the code, let me know if it looks ok.

durran:
Ship it. :+1:

PR #675, opened 2015-08-21
RUBY-951 Update YAML tests and refactor URI class

estolfo:
Yet another refactor to the URI class after changes to the spec.

### DIFF
--- a/spec/support/connection_string_tests/valid-warnings.yml
+++ b/spec/support/connection_string_tests/valid-warnings.yml
@@ -11,41 +11,41 @@ tests:
                 port: ~
         auth: ~
         options: ~
-    # -
-    #     description: "Unsupported option values are ignored"
-    #     uri: "mongodb://example.com/?fsync=ifPossible"
-    #     valid: true
-    #     warning: true
-    #     hosts:
-    #         -
-    #             type: "hostname"
-    #             host: "example.com"
-    #             port: ~
-    #     auth: ~
-    #     options: ~
-    # -
-    #     description: "Repeated option keys"
-    #     uri: "mongodb://example.com/?replicaSet=test&replicaSet=test"
-    #     valid: true
-    #     warning: true
-    #     hosts:
-    #         -
-    #             type: "hostname"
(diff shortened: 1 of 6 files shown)


spec/support/connection_string_tests/valid-warnings.yml
```diff
@@ -11,41 +11,41 @@ tests:
...
-    #     hosts:
-    #         -
-    #             type: "hostname"
-    #             host: "example.com"
-    #             port: ~
-    #     auth: ~
-    #     options: ~
-    # -
-    #     description: "Repeated option keys"
-    #     uri: "mongodb://example.com/?replicaSet=test&replicaSet=test"
```
jmikola: I think we still want to test for this error. Only readPreferenceTags can be repeated.

spec/support/connection_string_tests/valid-warnings.yml
```diff
@@ -11,41 +11,41 @@ tests:
...
+#        hosts:
+#            -
+#                type: "hostname"
+#                host: "example.com"
+#                port: ~
+#        auth: ~
+#        options: ~
+#    -
+#        description: "Repeated option keys"
+#        uri: "mongodb://example.com/?replicaSet=test&replicaSet=test"
```
jmikola: I think we do want to test for this. Only readPreferenceTags can be repeated.

spec/support/connection_string_tests/valid-warnings.yml
```diff
@@ -11,41 +11,41 @@ tests:
...
+#            -
+#                type: "hostname"
+#                host: "example.com"
+#                port: ~
+#        auth: ~
+#        options:
+#            replicaset: "test"
+#    -
+#        description: "Deprecated (or unknown) options are ignored if replacement exists"
+#        uri: "mongodb://example.com/?wtimeout=5&wtimeoutMS=10"
```
jmikola: For the record, this is ignored because Ruby driver 2.x doesn't have any deprecated options to even trigger this.

PR #677, opened 2015-08-22
RUBY-1006 Refactor View and Readable to allow for arbtirary query modifiers

estolfo:


spec/mongo/collection/view_spec.rb
```diff
@@ -18,6 +18,149 @@
...
+            { :$query => { a: 1 }, :$someMod => 100, :$orderby => { a: Mongo::Index::ASCENDING } }
+          end
+
+          let(:parsed_selector) do
+            { b: 2 }
+          end
+
+          let(:query_selector) do
+            { :$query => { a: 1 }, :$someMod => 100, :$orderby => { a: Mongo::Index::ASCENDING } }
+          end
```
estolfo: @behackett I'm assuming that if you provide a $query doc in the selector, it'll use that as the selector, not whatever else is in there. I.e., in this case, { a: 1 } is the selector, not { b: 2 }

spec/mongo/collection/view_spec.rb
```diff
@@ -545,11 +688,9 @@
             end
 
             it 'overrides the modifier value with the option value' do
-              expect(query_spec[:selector][:$query]).to eq(selector)
+              expect(query_spec[:selector][:$query]).to eq(options[:modifiers][:$query])
```
estolfo: @behackett If you have a $query in modifiers, does that get used instead of the selector?

for example:
coll.find({ a: 1 }, { :modifiers => { :$query => { b: 2 } } )

behackett: In PyMongo we don't do anything to try to correct that mistake. Since the options in modifiers are added to the criteria after the criteria is changed to {'$query': {'a': 1}} the query ends up being {'$query': {'b': 2}}.

lib/mongo/collection/view/readable.rb
```diff
@@ -461,26 +475,18 @@ def query_spec
...
+          @read_formatted ||= read.to_mongos
         end
 
         def special_selector
-          SPECIAL_FIELDS.reduce({}) do |hash, (key, method)|
-            value = send(method) || (options[:modifiers] && options[:modifiers][key])
-            hash[key] = value unless value.nil?
-            hash
-          end
+          sel = { :$query => selector }.merge(modifiers)
```
durran: This should be:

```ruby
sel = BSON::Document.new(:$query => selector).merge(modifiers)
```

In order to get the indifferent access provided by a `BSON::Document` vs. a normal hash, also, the command monitoring spec guarantees the filter will also be a `BSON::Document`.

durran:
Line 261 of `Mongo::Protocol::Query` (https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/protocol/query.rb#L261) also needs to change in this pull for command monitoring to:

```ruby
document[:filter] = filter[:$query] ? filter[:$query] : filter
```

estolfo:
ah, thanks.
I guess also the modifiers document should be initialized as a BSON::Document as well if the user hasn't specified it. Here: https://github.com/mongodb/mongo-ruby-driver/pull/677/files#diff-9da4b67a7dc4cd1487088c3f70d96f10R441

durran:
Yeah to be on the safe side I would initialize that as a `BSON::Document` as well.

PR #680, opened 2015-08-28
Improve Socket alive? behaviour

rkistner:
The root of the issue seems to be that Socket#alive? raised an error in some cases, instead of returning false. This prevents reconnection from happening.

See:
https://jira.mongodb.org/browse/RUBY-1011
https://jira.mongodb.org/browse/RUBY-1012

### DIFF
--- a/lib/mongo/socket.rb
+++ b/lib/mongo/socket.rb
@@ -56,12 +56,15 @@ module Mongo
     #
     # @since 2.0.0
     def alive?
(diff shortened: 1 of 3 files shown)


estolfo:
Thanks @rkistner for looking into this and for opening a pull request. I think we're going to take a different approach, you can see my comment on this ticket for more details: [RUBY-1012](https://jira.mongodb.org/browse/RUBY-1012)

PR #682, opened 2015-08-28
Retry for count and distinct commands if there is a connection error.

jonhyman:
We started doing failover testing against HEAD of the new driver and realized these two cases were missing from retry. We tested my branch and it works correctly.


estolfo:
Hi @jonhyman 
Thanks for catching this - would you mind rebasing against master so we can merge it in?
Thanks!
Emily

PR #687, opened 2015-09-02
RUBY-1021 Use Redacted options class to redact option fields when printing

estolfo:


durran:
I envisioned that we would only need to do this in places where the user provides the options, such as:

``Mongo::Client#initialize``
``Mongo::Client#with``

And methods like `merge` need to be implemented in order to not have to wrap every time we call `merge` in our codebase. Also, methods with signatures such as:

``method_name(options = {})`` should be changed to ``method_name(options = Options::Redacted.new)``

estolfo:
ok @durran, I've made all the suggested changes. Let me know if it looks good.

lib/mongo/options/redacted.rb
```diff
@@ -0,0 +1,156 @@
...
+      #
+      # @yieldparam [ String, Object ] The key as a string and its value.
+      #
+      # @return [ Options::Redacted ] This object.
+      #
+      # @since 2.1.0
+      def reject!
+        if block_given?
+          n_keys = keys.size
+          self.keys.each do |key|
```
durran: Don't need the `self` in front of anything but the `self[key]` and the direct reference.

lib/mongo/options/redacted.rb
```diff
@@ -0,0 +1,156 @@
...
+      #
+      # @yieldparam [ String, Object ] The key as a string and its value.
+      #
+      # @return [ Options::Redacted ] This object.
+      #
+      # @since 2.1.0
+      def select!
+        if block_given?
+          n_keys = keys.size
+          self.keys.each do |key|
```
durran: Same with `self` here.

lib/mongo/server_selector.rb
```diff
@@ -38,6 +38,11 @@ module ServerSelector
     # @since 2.0.0
     SERVER_SELECTION_TIMEOUT = 30.freeze
 
+    # Primary read preference.
+    #
+    # @since 2.1.0
+    PRIMARY = Options::Redacted.new(mode: :primary)
```
durran: Would `freeze` this.

lib/mongo/options/redacted.rb
```diff
@@ -0,0 +1,156 @@
...
+    # Class for wrapping options that could be sensitive.
+    # When printed, the sensitive values will be redacted.
+    #
+    # @since 2.1.0
+    class Redacted < BSON::Document
+
+      # The options whose values will be redacted.
+      #
+      # @since 2.1.0
+      SENSITIVE_OPTIONS = [ :password,
```
durran: Would `freeze` this.

lib/mongo/options/redacted.rb
```diff
@@ -0,0 +1,156 @@
...
+                            :pwd ]
+
+      # The replacement string used in place of the value for sensitive keys.
+      #
+      # @since 2.1.0
+      STRING_REPLACEMENT = '<REDACTED>'
```
durran: `freeze` as well here.

spec/mongo/options/redacted_spec.rb
```diff
@@ -0,0 +1,346 @@
...
+      end
+
+      it 'replaces the value with the redacted string' do
+        expect(options.to_s).to match(Mongo::Options::Redacted::STRING_REPLACEMENT)
+      end
+    end
+
+    context 'when the hash does not contain a sensitive key' do
+
+      it 'prints all the values' do
```
durran: Need an assertion here.

PR #690, opened 2015-09-09
RUBY-1025 Make sure collections info returns all results

estolfo:


lib/mongo/database/view.rb
```diff
@@ -51,6 +51,7 @@ class View
       def collection_names(options = {})
         @batch_size = options[:batch_size]
         server = next_primary
+        @limit = -1 if server.context.features.list_collections_enabled?
```
durran: Can just be `server.features.list_collections_enabled?`

durran:
Just the one comment then I think good to go.

PR #691, opened 2015-09-11
Delegate collections to the current database

namusyaka:
Hi,
I'd like to use `collections` as an instance method of `Mongo::Client`.
The use-case example is [here](https://github.com/padrino/padrino-framework/blob/592fd760619ba86af1c4de5fa0b3f5170c418ab1/padrino-gen/lib/padrino-gen/padrino-tasks/mongoid.rb#L6-L60).

Thanks.

### DIFF
--- a/lib/mongo/client.rb
+++ b/lib/mongo/client.rb
@@ -36,8 +36,8 @@ module Mongo
     # @return [ Hash ] options The configuration options.
     attr_reader :options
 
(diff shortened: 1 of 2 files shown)


estolfo:
Hi @namusyaka 
This is a good idea but would you mind adding a test to client_spec.rb ?
Thanks

namusyaka:
@estolfo Sure!

PR #692, opened 2015-09-27
Retry on mongos operation failures

durran:
This adds additional read retry scenarios on `OperationFailure`s coming from a mongos only. This is because the `ping` that is executed against a running mongos always returns ready no matter what the state of the cluster behind it is, and the normal server selection/socket retry will not cover this scenario.

1. Adds a `max_read_retries` configuration option that is only applicable to sharded clusters when specific `OperationFailure`s are raised.
2. Adds a `read_retry_interval` configuration option to set the wait time between retries.

Note that we can only look at the error messages themselves, and not the codes. The codes have changed between server versions and some are not relevant to the same scenario anymore, such as the commented 15988 - which was previously an "error querying server" in mongos but in 3.0 is a "FieldPath field names may not be empty strings", which is not a retryable error.

We don't scan the cluster in these scenarios as the mongos is up and running, and the scan would be a waste of a network call do to the `ping` reasons stated above.

### DIFF
--- a/lib/mongo/error/operation_failure.rb
+++ b/lib/mongo/error/operation_failure.rb
@@ -33,7 +33,9 @@ module Mongo
         'no master',
         'not master',
         'connect failed',
-        'error querying'
+        'error querying',
+        'could not get last error',
+        'connection attempt failed'
       ].freeze
 
       # Can the operation that caused the error be retried?
--- a/lib/mongo/error/operation_failure.rb
+++ b/lib/mongo/error/operation_failure.rb
@@ -18,6 +18,35 @@ module Mongo
     # Raised when an operation fails for some reason.
     #
     # @since 2.0.0
-    class OperationFailure < Error; end
+    class OperationFailure < Error
+
+      # These are magic error messages that could indicate a cluster
+      # reconfiguration behind a mongos. We cannot check error codes as they
+      # change between versions, for example 15988 which has 2 completely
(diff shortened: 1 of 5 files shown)


lib/mongo/error/operation_failure.rb
```diff
@@ -18,6 +18,35 @@ class Error
...
-    class OperationFailure < Error; end
+    class OperationFailure < Error
+
+      # These are magic error messages that could indicate a cluster
+      # reconfiguration behind a mongos. We cannot check error codes as they
+      # change between versions, for example 15988 which has 2 completely
+      # different meanings between 2.4 and 3.0.
+      #
+      # @since 2.1.1
+      RETRY_MESSAGES = [
```
jonhyman: `"could not get last error"` and `"connection attempt failed"` and `""` are others I think you should add here.

jonhyman:
Thank you for writing this. As mentioned, I'll do a series of failover tests on Tuesday and will report back.

lib/mongo/error/operation_failure.rb
durran: Added.

jonhyman:
Success! You need to merge in https://github.com/mongodb/mongo-ruby-driver/pull/693 to this, but then it's fantastic. We did various stepdowns and `kill -9` of primaries and secondaries and it recovered well.

durran:
Great!

PR #697, opened 2015-10-06
RUBY-1034 - Support partial indexes

aherlihy:


spec/mongo/index/view_spec.rb
```diff
@@ -196,7 +196,70 @@
...
+      it 'raises an exception' do
+        expect {
+          view.create_one(spec, partial_filter_expression: {"x": {"$asda": 3}})
+        }.to raise_error(Mongo::Error::OperationFailure)
+      end
+      it 'raises an exception' do
+        expect {
+          view.create_one(spec, partial_filter_expression: {
+                                  "$and": [{"$and": [{"x": {"$lt": 2}},
+                                                     {"x": {"$gt": 0}}]},
```
estolfo: Check out the failures on Travis - this syntax isn't valid in older versions of Ruby.
For example: https://travis-ci.org/mongodb/mongo-ruby-driver/jobs/83954684

spec/mongo/index/view_spec.rb
```diff
@@ -196,7 +196,70 @@
         expect(view.get('random_name')).to_not be_nil
       end
     end
-  end
+
+    context 'when providing an invalid partial index filter' do
+
+      let(:spec) do
+        { random:1 }
```
estolfo: We would put in a space in between the key and the value in a hash. For example:

```ruby
{ random: 1}
```

and here's an existing example:
https://github.com/aherlihy/mongo-ruby-driver/blob/1034-partial-index/spec/mongo/index/view_spec.rb#L107

spec/mongo/index/view_spec.rb
estolfo: Here's an example of the syntax required when you need to create a key with a '$'

https://github.com/aherlihy/mongo-ruby-driver/blob/1034-partial-index/spec/mongo/operation/write/update_spec.rb#L190

spec/mongo/index/view_spec.rb
```diff
@@ -196,7 +196,70 @@
...
+      end
+
+      it 'raises an exception' do
+        expect {
+          view.create_one(spec, partial_filter_expression: 5)
+        }.to raise_error(Mongo::Error::OperationFailure)
+      end
```
estolfo: We normally add spaces in between tests.

spec/mongo/index/view_spec.rb
```diff
@@ -196,7 +196,70 @@
...
+
+      it 'returns ok' do
+        expect (explain['queryPlanner']['winningPlan'].fetch(
+                   'stage', {}) == "COLLSCAN")
+      end
+
+      after do
+        view.drop_one('x_1')
+      end
```
estolfo: The after block is defined at the top of the inner-most 'context' block.

spec/mongo/index/view_spec.rb
```diff
@@ -196,7 +196,70 @@
...
+
+    end
```
estolfo: There's some indentation spacing issues.

lib/mongo/index/view.rb
```diff
@@ -251,6 +254,10 @@ def send_initial_query(server)
 
       def with_generated_names(models)
         models.dup.each do |model|
+          if model[:key].key?(:partial_filter_expression)
+            model[:partial_filter_expression] = model[:key][:partial_filter_expression]
+            model[:key].delete(:partial_filter_expression)
+          end
```
estolfo: I have to double check, but I don't believe this code is necessary. Since you added this line https://github.com/mongodb/mongo-ruby-driver/pull/697/files#diff-4104300dfb8ef7232ede5d88a9008676R59
it should probably set the partial filter expression option correctly.

spec/mongo/index/view_spec.rb
```diff
@@ -196,7 +196,70 @@
...
+                                                     {"x": {"$gt": 0}}]},
+                                           {"x": {"$exists": true}}]})
+        }.to raise_error(Mongo::Error::OperationFailure)
+      end
+    end
+
+    context 'when providing a valid partial index filter' do
+      let!(:result) do
+        view.create_one({"x": 1}, partial_filter_expression: {
+                                    "a": {"$lte": 1.5}})
```
estolfo: The ! of the let is not necessary. The difference between defining result as 
```ruby
let(:result)
```
and 
```ruby
let!(:result)
```

is that the first is lazily evaluated. In other words, it will be evaluated when you refer to it in the actually test (i.e. the 'it' block)
When you use a '!', the result is evaluated before the 'it' block runs. Sometimes this is necessary, but not in this case.
So you can just use lazy-evaluation and define the result as 

```ruby
let(:result)
```

lib/mongo/index/view.rb
estolfo: Yeah, I doubled-check and you can take out these 4 lines. The partial_filter_expression will just be passed through as part of the index spec. The name will be converted from partial_filter_expression to partialFilterExpression [here.](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/index/view.rb#L244)

spec/mongo/index/view_spec.rb
```diff
@@ -196,7 +196,70 @@
         expect(view.get('random_name')).to_not be_nil
       end
     end
-  end
+
+    context 'when providing an invalid partial index filter' do
```
estolfo: In general, the Ruby driver differs from the Python driver in that we don't rigorously check input from users, nor do we test different invalid input scenarios, unless it's our code that does the validation.
If we do test invalid user input, we usually only check one type and verify that an OperationFailure is properly raised. So here, I would only keep [this test](https://github.com/mongodb/mongo-ruby-driver/pull/697/files#diff-5fc639ad0e4b26a0c396bab4704c3fceR206) and take out the other invalid cases.

lib/mongo/index/view.rb
aherlihy: When I comment out those 4 lines and run: client[:test].indexes.create_one({"x":1, partial_filter_expression:{"a":{"$lte":1.5}}})
I get an error:
"bad index key pattern { x: 1, partial_filter_expression: { a: { $lte: 1.5 } } }: Index keys cannot be Objects or Arrays." }

Am I passing in the arguments incorrectly?

spec/mongo/index/view_spec.rb
aherlihy: Done

spec/mongo/index/view_spec.rb
aherlihy: Done

spec/mongo/index/view_spec.rb
aherlihy: What is the difference between assigning something with ':' and '=>'?

spec/mongo/index/view_spec.rb
aherlihy: Done

spec/mongo/index/view_spec.rb
aherlihy: Done

spec/mongo/index/view_spec.rb
aherlihy: Done

spec/mongo/index/view_spec.rb
estolfo: I'll explain in person when you have a minute.

spec/mongo/index/view_spec.rb
```diff
@@ -196,7 +196,70 @@
...
+
+      it 'returns ok' do
+        expect(result).to be_successful
+      end
+
+      let!(:explain) do
+        authorized_collection.find(:x => 6, :a => 1).explain
+      end
+
+      it 'returns ok' do
```
estolfo: This should have a more descriptive test description instead of "returns ok".
Something like "Applies the filter when using the index""

spec/mongo/index/view_spec.rb
```diff
@@ -196,7 +196,70 @@
...
+        expect (explain['queryPlanner']['winningPlan'].fetch(
+                   'inputStage', {})['indexName'] == "x_1")
+        expect (explain['queryPlanner']['winningPlan'].fetch(
+                   'inputStage', {})['isPartial'])
```
estolfo: I think you're using #fetch on the hash because 'inputStage' may or may not be a key. If you refer to a key in a Ruby hash that doesn't exist, it'll just return nil; it won't raise an exception.
So you can just do 

```ruby
expect (explain['queryPlanner']['winningPlan']['inputStage']).to eq('isPartial')
```
If explain['queryPlanner']['winningPlan'] doesn't have the key 'inputStage', the value of the expectation will be nil and it will be compared with the String 'isPartial'. Which would be false, failing the test.
You always want expect(ations) in rspec tests to actually inspect a value. Here you just have expect(..) and nothing else so you should add the ".to eq('isPartial')"

spec/mongo/index/view_spec.rb
```diff
@@ -196,7 +196,70 @@
...
+      end
+
+      let!(:explain) do
+        authorized_collection.find(:x => 6, :a => 1.6).explain
+      end
```
estolfo: You can change this to let(:explain) also (take out the !)

lib/mongo/index/view.rb
durran: No, it should be:

```ruby
client[:test].indexes.create_one({ "x" => 1}, partial_filter_expression: { "a" : { "$lte" => 1.5 }}})
```

The way you are doing it makes the options part of the index key, it needs to be clear that they are 2 separate parameters.

aherlihy: Makes sense to me. I'll update the tests.

aherlihy:
@estolfo updated test with version check + rebased

aherlihy:
@estolfo Updated tests to use listIndexes instead of explain.

spec/mongo/index/view_spec.rb
```diff
@@ -196,6 +196,47 @@
...
+
+      after do
+        view.drop_one('x_1')
+      end
+
+      it 'returns ok' do
+        expect(result).to be_successful
+      end
+
+      it 'creates an index', if: find_command_enabled? do
```
estolfo: You don't have to check find_command_enabled?
Putting the check at the "parent context" level is sufficient.

spec/mongo/index/view_spec.rb
```diff
@@ -196,6 +196,47 @@
...
+        expect(indexes).to_not be_nil
+      end
+
+      it 'passes partialFilterExpression correctly', if: find_command_enabled? do
```
estolfo: Same here: you don't need the find_command_enabled? check

spec/mongo/index/view_spec.rb
```diff
@@ -196,6 +196,47 @@
         expect(view.get('random_name')).to_not be_nil
       end
     end
+
```
estolfo: You could do a structure with contexts like this instad:

context 'when providing a partial index filter', if: find_command_enabled? do
   context 'when the filter is invalid' do
   context 'when the filter is valid' do

aherlihy:
@estolfo updated

estolfo:
Thanks! If you could rebase, that'd be great so I can merge it.

aherlihy:
@estolfo Rebased!

PR #701, opened 2015-10-08
Validate client options. Related to MONGOID-4157

estolfo:


durran:
Just a thought - when invalid options are provided in a URI, we log it. I think for consistency it might be good to do it here as well.

estolfo:
@durran done

durran:
:+1:

lib/mongo/client.rb
```diff
@@ -27,6 +28,40 @@ class Client
     # @since 2.1.0
     CRUD_OPTIONS = [ :database, :read, :write ].freeze
 
+    # Valid client options.
+    #
+    # @since 2.1.2
+    VALID_OPTIONS = [
+        :auth_mech,
```
durran: I'd do a 2 space indent here as it's standard formatting in Ruby, and then freeze the array.

estolfo: sorry, my ide puts in that spacing automatically. I'll update it.

PR #702, opened 2015-10-08
Update fs_bucket.rb

ericparshall:
Previously, this code was querying all IDs in the collection and storing them in an array. This is very inefficient and causes a long delay when inserting a file into fs.files, especially when the system is under a heavy load.

### DIFF
--- a/lib/mongo/grid/fs_bucket.rb
+++ b/lib/mongo/grid/fs_bucket.rb
@@ -438,7 +438,7 @@ module Mongo
       end
 
       def ensure_indexes!
-        if files_collection.find({}, projection: { _id: 1 }).to_a.empty?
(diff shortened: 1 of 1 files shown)


estolfo:
Good point, thanks! You can even just call #count on collection directly for simplicity:

```ruby
if files_collection.count == 0
```

durran:
Actually the grid fs specification states that we should be doing a `find_one` instead of a count for performance against sharded clusters:

https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#before-write-operations

estolfo:
You're right. That explains why I was using find_one and not count.
Let me check with the spec author to see what the alternatives are.

PR #703, opened 2015-10-09
Fix RUBY-1051: fails if exists a toplevel Result class

crccw:
By using const_defined? instead of defined, we can restrict the searching for Result class under current namespace.

[Review on Reviewable](https://reviewable.io/reviews/mongodb/mongo-ruby-driver/703)


estolfo:
Thanks @crccw

PR #706, opened 2015-10-12
RUBY-1038: Implement Read Concern for 3.2

durran:


behackett:
Read concern is supposed to be a Collection level option.

https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#location-specification

durran:
It's a client, database, AND collection level option (as well as operation level)... I haven't handled that yet - only the operation level ones which is why this is still marked as WIP.

behackett:
Do we allow passing write concern per operation?

durran:
Spec states: "a driver MAY allow it to be specified on a per-operation basis in accordance with the CRUD specification" - and the driver ticket has them per command as well for those we provide helpers for. (https://jira.mongodb.org/browse/DRIVERS-254)

durran:
Oh sorry - you said write concern... Yes we do on some operations.

behackett:
As long as we're being consistent.

durran:
Actually I just double checked and we pull it from collection. Good catch then - I'll change these to do the same.

behackett:
Great. Thanks!

PR #711, opened 2015-10-19
RUBY-1036 Support bypassDocumentValidation option

aherlihy:


spec/mongo/bulk_write_spec.rb
```diff
@@ -8,6 +8,14 @@
...
   after do
     authorized_collection.delete_many
+    collection_with_validator.drop()
+  end
+
+  let(:collection_with_validator) do
+    c = authorized_client[:validating,
+                          :validator => {:a => {'$exists' => true} }]
+    c.create
+    c
```
aherlihy: Is there an easier way to do this?

lib/mongo/operation/write/command/update.rb
```diff
@@ -45,11 +46,14 @@ class Update
...
-              writeConcern: write_concern.options,
-              ordered: ordered?
+            cmd = {
+                update: coll_name,
+                updates: updates,
+                writeConcern: write_concern.options,
+                ordered: ordered?
             }
+            cmd['bypassDocumentValidation'] = true if bypass_document_validation
+            return cmd
```
aherlihy: Any way to do this so I don't have to return cmd?

lib/mongo/operation/specifiable.rb
```diff
@@ -131,6 +131,11 @@ module Specifiable
       # @since 2.0.0
       READ = :read.freeze
 
+      # Whether to bypass document level validation.
+      #
+      # @since 3.2.0 (?)
```
estolfo: this would be 2.2.0

https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/version.rb#L20

lib/mongo/operation/specifiable.rb
```diff
@@ -332,6 +337,18 @@ def options
...
         spec[READ_CONCERN]
+
+      # Whether or not to bypass document level validation.
+      #
+      # @example Get the bypass_document_validation option.
+      #
+      # @return [ Boolean ] Whether to bypass document level validation.
+      #
+      # @since 3.2.0 (?)
+      end
```
estolfo: same here: 2.2.0

I think you put the comments in the read_concern method by accident.

lib/mongo/operation/write/command/update.rb
estolfo: Firstly, the return value of a Ruby method is the last expression evaluated, so you would just write

```ruby
cmd
```

as the last line.
But to answer your question, you could do

```ruby
bypass_document_validation ? cmd.merge!('bypassDocumentValidation' => true) : cmd
```

lib/mongo/collection.rb
```diff
@@ -247,6 +247,7 @@ def find(filter = nil, options = {})
     #   aggregation to run.
     # @option options [ true, false ] :use_cursor Indicates whether the command will request that the server
     #   provide results using a cursor.
+    # @option options [ Boolean ] :bypass_document_validation Whether or not to skip document level validation.
```
estolfo: You would write [ true, false ] here

lib/mongo/collection.rb
```diff
@@ -371,6 +373,7 @@ def insert_many(documents, options = {})
     #   should be executed in order.
     # @option options [ Hash ] :write_concern The write concern options.
     #   Can be :w => Integer, :fsync => Boolean, :j => Boolean.
+    # @option options [ Boolean ] :bypass_document_validation Whether or not to skip document level validation.
```
estolfo: [ true, false ]

lib/mongo/collection.rb
```diff
@@ -436,6 +439,7 @@ def parallel_scan(cursor_count)
     #
     # @option options [ true, false ] :upsert Whether to upsert if the
     #   document doesn't exist.
+    # @option options [ Boolean ] :bypass_document_validation Whether or not to skip document level validation.
```
estolfo: [ true, false ] for all the following ones

spec/mongo/bulk_write_spec.rb
estolfo: I think you can just do

```ruby
authorized_client[:validating, :validator => {:a => {'$exists' => true} }]
```

spec/mongo/bulk_write_spec.rb
```diff
@@ -425,4 +433,44 @@
...
+      it 'raises BulkWriteError' do
+        expect {
+          result
+        }.to raise_exception(Mongo::Error::BulkWriteError)
+      end
+
+      let(:result2) do
+        collection_with_validator.bulk_write(
+            requests, :bypass_document_validation => true)
+      end
```
estolfo: I would define this result2 right after result, then have the two 'it' blocks

spec/mongo/collection_spec.rb
```diff
@@ -723,12 +731,43 @@
         }.to raise_exception(Mongo::Error::BulkWriteError)
       end
     end
+
+    context 'when bypassing document validation', if: find_command_enabled? do
+
+      let(:result) do
+        collection_with_validator.insert_many([{:a => 1}, {:a => 2}])
+      end
```
estolfo: You can define these documents as:

```ruby
{ a: 1 }
```

spec/mongo/collection_spec.rb
```diff
@@ -723,12 +731,43 @@
...
+
+      it 'inserts successfully when documents are valid' do
+        expect(result.inserted_count).to eq(2)
+      end
+
+      let(:result2) do
+        collection_with_validator.insert_many([{:x => 1}, {:x => 2}])
```
estolfo: Here you can define the documents as

```ruby
[{ x: 1 }, { x: 2 }]
```

spec/mongo/collection_spec.rb
```diff
@@ -816,7 +885,10 @@
     context 'when options are provided' do
 
       let(:options) do
-        { :allow_disk_use => true }
+        {
+            :allow_disk_use => true,
+            :bypass_document_validation => true
```
estolfo: only use two spaces to indent here

spec/mongo/collection_spec.rb
```diff
@@ -1156,6 +1228,44 @@
...
+          result2
+        }.to raise_exception(Mongo::Error::OperationFailure)
+      end
+
+      let(:result3) do
+        collection_with_validator.replace_one(
+            {:a => 1}, {:x => 1}, :bypass_document_validation => true)
+      end
+
+      it 'replaces successfully when document is invalid but bypass_document_validation is true' do
```
estolfo: you don't have to say "but bypass_document_validation is true" because it's stated in the outer context. When the test runs, it'll read "when bypassing document validation it replaces successful when document is invalid"

durran:
Just a general note, commit messages should all be prefixed with the Jira ticket and a bit more descriptive. While they make some sense in the context of this pull request, once merged they will make no sense once doing a `git log` and browsing the commits... I would `rebase -i` against master and squash them into a single commit to alleviate this with a subject line at maximum 50 characters and a wrapped body at 72 characters... For example:

```
RUBY-1036: Support bypassDocumentValidation Option:

- Added bypassDocumentValidation option to related write operations
```

aherlihy:
Ok cool, I'll squash before committing. Do you generally squash your commits before or after making a PR?

spec/mongo/bulk_write_spec.rb
aherlihy: Unfortunately it doesn't work when I remove the create call. I'm not 100% sure why, but it doesn't trigger the errors (I think it just treats the collection as one without a validator). Same thing for collection_spec.rb.

Maybe I am misunderstanding how collections are created with options?

lib/mongo/operation/write/command/update.rb
aherlihy: Done

lib/mongo/operation/specifiable.rb
aherlihy: Done

lib/mongo/operation/specifiable.rb
aherlihy: Done.

lib/mongo/collection.rb
aherlihy: Done.

lib/mongo/collection.rb
aherlihy: Done.

lib/mongo/collection.rb
aherlihy: Done.

spec/mongo/bulk_write_spec.rb
aherlihy: Done.

spec/mongo/collection_spec.rb
aherlihy: Done.

spec/mongo/collection_spec.rb
aherlihy: I'm going to rephrase this tests so they're easier to understand

spec/mongo/collection_spec.rb
aherlihy: Updated.

aherlihy:
@estolfo @durran Updated.

lib/mongo/bulk_write.rb
```diff
@@ -131,7 +131,9 @@ def base_spec(operation_id)
         :coll_name => collection.name,
         :write_concern => write_concern,
         :ordered => ordered?,
-        :operation_id => operation_id
+        :operation_id => operation_id,
+        :bypass_document_validation =>
+            options[:bypass_document_validation] || false
```
estolfo: You can just make this

```ruby
!!options[:bypass_document_validation]
```

The double bang will convert the value to a boolean. If you try this out in irb (Ruby console), you'll see how it works:

```ruby
> !!nil
 => false 
> !!false
 => false 
> !!true
 => true
```

lib/mongo/collection.rb
```diff
@@ -338,6 +339,7 @@ def insert_one(document, options = {})
           :db_name => database.name,
           :coll_name => name,
           :write_concern => write_concern,
+          :bypass_document_validation => options[:bypass_document_validation] || false,
```
estolfo: same here, you can use the double bang.

lib/mongo/collection.rb
```diff
@@ -371,6 +373,7 @@ def insert_many(documents, options = {})
     #   should be executed in order.
     # @option options [ Hash ] :write_concern The write concern options.
     #   Can be :w => Integer, :fsync => Boolean, :j => Boolean.
+    # @option options [ true, false ] :bypass_document_validation Whether or not to skip document level validation.
```
estolfo: These documentation lines should be shorter. I would put them on two lines. Generally, 80 characters is the max line length in Ruby.

lib/mongo/collection/view/writable.rb
```diff
@@ -189,7 +190,9 @@ def update(spec, multi, opts)
                            Operation::UPSERT => !!opts[:upsert] },
               :db_name => collection.database.name,
               :coll_name => collection.name,
-              :write_concern => collection.write_concern
+              :write_concern => collection.write_concern,
+              :bypass_document_validation =>
+                  opts[:bypass_document_validation] || false
```
estolfo: You can use double bang here.

lib/mongo/operation/specifiable.rb
```diff
@@ -334,6 +339,17 @@ def read_concern
         spec[READ_CONCERN]
       end
 
+      # Whether or not to bypass document level validation.
+      #
+      # @example Get the bypass_document_validation option.
```
estolfo: You need to put an example here, like the following:

```ruby
#   specifiable.bypass_documentation_validation.
```

lib/mongo/operation/write/command/insert.rb
```diff
@@ -40,11 +40,14 @@ class Insert
...
-              writeConcern: write_concern.options,
-              ordered: ordered?
+            cmd = {
+                insert: coll_name,
+                documents: documents,
+                writeConcern: write_concern.options,
+                ordered: ordered?
             }
+            cmd['bypassDocumentValidation'] = true if bypass_document_validation
+            return cmd
```
estolfo: You don't need to write "return"
Ruby methods return the result of the last expression evaluated. So you can just write

```ruby
cmd
```

However, I would write the last line as it is in the following:

```ruby
def selector
  cmd = {
    insert: coll_name,
    documents: documents,
    writeConcern: write_concern.options,
    ordered: ordered?
    }
  bypass_document_validation ? cmd.merge!(BYPASS_DOCUMENT_VALIDATION => true) : cmd
end
```

In the above, I'm reusing the constant BYPASS_DOCUMENT_VALIDATION so I'm not instantiating another String. I'm also using Hash#merge! instead of explicitly setting the hash value because it returns the hash as a result of calling #merge. The code is more succinct this way.

lib/mongo/operation/specifiable.rb
```diff
@@ -131,6 +131,11 @@ module Specifiable
       # @since 2.0.0
       READ = :read.freeze
 
+      # Whether to bypass document level validation.
+      #
+      # @since 2.2.0
+      BYPASS_DOCUMENT_VALIDATION = :bypass_document_validation.freeze
```
estolfo: Maybe call this

```ruby
BYPASS_DOC_VALIDATION
```

because it's shorter.

lib/mongo/operation/write/command/update.rb
```diff
@@ -45,11 +46,13 @@ class Update
...
-              updates: updates,
-              writeConcern: write_concern.options,
-              ordered: ordered?
+            cmd = {
+                update: coll_name,
+                updates: updates,
+                writeConcern: write_concern.options,
+                ordered: ordered?
             }
+            bypass_document_validation ? cmd.merge!('bypassDocumentValidation' => true) : cmd
```
estolfo: Is there a constant for the string 'bypassDocumentValidation' that we could use here so we aren't instantiating another String? If not, I would use a Symbol anyway, as it would be a String singelton.

spec/mongo/bulk_write_spec.rb
estolfo: ok, that makes sense because in order for the collection options to be applied, you actually have to call the create collection command. Otherwise, this let block will return the collection object and just start using it, without calling create to apply the options.

In that case, I would write the block as:

```ruby
let(:collection_with_validator) do
  authorized_client[:validating, :validator => {:a => {'$exists' => true} }].tap do |c|
    c.create
  end
end
```

#tap runs a block on an object and returns the object when the block is done. It's more Ruby-like = )

spec/mongo/bulk_write_spec.rb
```diff
@@ -425,4 +433,44 @@
...
+
+      let(:requests) do
+        [
+            { insert_one: { :x => 1 } },
+            { update_one: { filter: { :a => 1 },
+                            update: { '$unset' => { :a => '' }} } },
+            { replace_one: { filter: { :a => 2 },
+                             replacement: { :x => 2 }}}
+        ]
+      end
```
estolfo: I would call this 'ops'

spec/mongo/bulk_write_spec.rb
aherlihy: I knew there would be some super ruby-like syntax for this :)

lib/mongo/bulk_write.rb
aherlihy: Done.

lib/mongo/collection.rb
aherlihy: Done.

lib/mongo/collection.rb
aherlihy: Done.

lib/mongo/collection/view/writable.rb
aherlihy: Done.

spec/mongo/bulk_write_spec.rb
estolfo: actually, this makes so much more sense as you can see the result right before the test. You can leave it like this, thanks = )

lib/mongo/operation/specifiable.rb
aherlihy: Done.

lib/mongo/operation/write/command/insert.rb
aherlihy: Done.

lib/mongo/operation/specifiable.rb
aherlihy: Done.

lib/mongo/operation/write/command/update.rb
aherlihy: Done.

spec/mongo/bulk_write_spec.rb
```diff
@@ -425,4 +433,44 @@
...
+            requests, :bypass_document_validation => true)
+      end
+
+      it 'raises BulkWriteError' do
+        expect {
+          result
+        }.to raise_exception(Mongo::Error::BulkWriteError)
+      end
+
+      it 'executes successfully when bypass_document_validation is true' do
```
estolfo: I know it seems like so many levels of contexts, but it would be more consistent to have a context 'when bypass_document_validation is true' then the it 'executes successfully' test.

spec/mongo/collection_spec.rb
```diff
@@ -4,11 +4,18 @@
...
   end
 
   let(:collection_invalid_write_concern) do
     authorized_collection.client.with(write: { w: (WRITE_CONCERN[:w] + 1) })[authorized_collection.name]
   end
+  let(:collection_with_validator) do
+    c = authorized_client[:validating,
+                          :validator => {:a => {'$exists' => true} }]
+    c.create
+    c
```
estolfo: you can use #tap here.

spec/mongo/collection_spec.rb
```diff
@@ -723,12 +731,43 @@
...
+      end
+
+      it 'inserts successfully when documents are valid' do
+        expect(result.inserted_count).to eq(2)
+      end
+
+      let(:result2) do
+        collection_with_validator.insert_many([{ x: 1 }, { x: 2 }])
+      end
+
```
estolfo: have a context here 'when documents are invalid'

spec/mongo/collection_spec.rb
```diff
@@ -723,12 +731,43 @@
...
+      it 'raises a BulkWriteError when documents are invalid' do
```
estolfo: then this test
it 'raises a BulkWriteError' do

spec/mongo/collection_spec.rb
```diff
@@ -723,12 +731,43 @@
...
+          result2
+        }.to raise_exception(Mongo::Error::BulkWriteError)
+      end
+
+      let(:result3) do
+        collection_with_validator.insert_many(
+            [{ x: 1 }, { x: 2 }], :bypass_document_validation => true)
+      end
+
+      it 'inserts successfully when documents are invalid but bypass_document_validation is true' do
```
estolfo: then a context here 
'when bypass_document_validation is true'

then a test

it 'inserts successfully'

spec/mongo/bulk_write_spec.rb
aherlihy: Done.

spec/mongo/collection_spec.rb
```diff
@@ -760,6 +799,36 @@
...
+
+      let(:result3) do
+        collection_with_validator.insert_one(
+            { x: 1 }, :bypass_document_validation => true)
+      end
+
+      it 'inserts successfully when document is invalid but bypass_document_validation is true' do
+        expect(result3.written_count).to eq(1)
+      end
+    end
```
estolfo: This is the structure I would have:

```ruby

context 'when collection has a validator' do

	context 'when the document is valid' do
	
		it 'inserts successfully' do

	context 'when the document is invalid' do

		context 'when bypass_document_validation is not set' do

			it 'raises an OperationFailure' do

		context 'when bypass_document_validation is true' do

			it 'inserts successfully' do

```

spec/mongo/collection_spec.rb
```diff
@@ -1156,6 +1225,44 @@
...
+
+      let(:result3) do
+        collection_with_validator.replace_one(
+            { a: 1 }, { x: 1 }, :bypass_document_validation => true)
+      end
+
+      it 'replaces successfully when document is invalid but bypass_document_validation is true' do
+        expect(result3.written_count).to eq(1)
+      end
+    end
```
estolfo: Try to use the example I put above to restructure the context blocks here.

spec/mongo/bulk_write_spec.rb
aherlihy: Done.

spec/mongo/collection_spec.rb
aherlihy: Done.

spec/mongo/collection_spec.rb
aherlihy: Done.

spec/mongo/collection_spec.rb
aherlihy: Done.

aherlihy:
@durran @estolfo Updated.

durran:
:+1:

spec/mongo/collection_spec.rb
```diff
@@ -760,6 +811,48 @@
...
+
+      context 'when the document is invalid' do
+
+        context 'when bypass_document_validation is not set' do
+
+          let(:result2) do
+            collection_with_validator.insert_one({ x: 1 })
+          end
+
+          it 'raises a OperationError' do
```
estolfo: This should be OperationFailure, not OperationError

spec/mongo/collection_spec.rb
```diff
@@ -1262,6 +1405,60 @@
...
+        it 'updates successfully' do
+          expect(result.modified_count).to eq(2)
+        end
+      end
+
+      context 'when the document is invalid' do
+
+        context 'when bypass_document_validation is not set' do
+
+
```
estolfo: there is an extra newline here

estolfo:
Just three more changes and then it'll be good to merge.
1 and 2: the comments I made in-line
3: General format change in multiple places: We always put a space after a { and before a } of a hash. So you would change this:

```ruby
{:a => {'$gt' => 0}}, '$inc' => {:a => 1} )
```
to

```ruby
{ :a => { '$gt' => 0 } }, '$inc' => { :a => 1 })
```
Note that we don't put the space after a ( or before a )
Don't ask me why, c'est juste comme ça = )

spec/mongo/collection_spec.rb
aherlihy: Done.

spec/mongo/collection_spec.rb
aherlihy: Done.

aherlihy:
@estolfo Updated

estolfo:
thanks @aherlihy !

PR #714, opened 2015-10-21
RUBY-1037 - Support sending write concern to findAndModify commands

aherlihy:
- Added write_concern option to find_one_and_delete, find_one_and_replace, find_one_and_update.

### DIFF
--- a/lib/mongo/collection/view/writable.rb
+++ b/lib/mongo/collection/view/writable.rb
@@ -35,6 +35,7 @@ module Mongo
           cmd[:fields] = projection if projection
           cmd[:sort] = sort if sort
           cmd[:maxTimeMS] = max_time_ms if max_time_ms
+          cmd[:writeConcern] = options[:write_concern] || collection.write_concern.options
           write_with_retry do
             database.command(cmd).first['value']
           end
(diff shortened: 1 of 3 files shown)


spec/mongo/collection_spec.rb
```diff
@@ -1710,6 +1710,16 @@
         }.to raise_exception(Mongo::Error::OperationFailure)
       end
     end
+
+    context 'when write_concern is provided', if: find_command_enabled? && standalone? do
+
+      it 'raises OperationFailure when w=2' do
```
estolfo: You can just say 'uses the write concern' because the test verifies that the write concern is set. The fact that it raises the OperationFailure is proof of this, not what we are actually testing.

estolfo:
other than that one comment, this looks good to me!

aherlihy:
@estolfo Updated

lib/mongo/collection/view/writable.rb
```diff
@@ -55,6 +56,10 @@ def find_one_and_delete
         # @option opts [ true, false ] :upsert Whether to upsert if the document doesn't exist.
         # @option opts [ true, false ] :bypass_document_validation Whether or
         #   not to skip document level validation.
+        # @option opts [ true, false ] :upsert Whether to upsert if the
+        #   document doesn't exist.
```
estolfo: I'm wondering why this is part of the commit. Have you rebased against master?

aherlihy: Oh that's weird, I rebased against master before comitting. I'll get rid of that line, I'm not sure what happened.

estolfo: Yeah, I can't figure out why it's in there. Your branch looks up-to-date with master.

aherlihy:
@estolfo Updated

estolfo:
thanks @aherlihy !

PR #718, opened 2015-10-26
RUBY-1050 Comply with write concern spec

estolfo:


lib/mongo/error/invalid_write_concern.rb
```diff
@@ -0,0 +1,34 @@
...
+    class InvalidWriteConcern < Error
+
+      # Instantiate the new exception.
+      #
+      # @example Instantiate the exception.
+      #   Mongo::Error::InvalidWriteConcern.new
+      #
+      # @since 2.2.0
+      def initialize
+        super('Invalid write concern options.')
```
behackett: It might be nice to provide a reason. In PyMongo the error message tells you why the write concern is invalid.

estolfo: thanks @behackett, I've updated it to include a more descriptive error message.

spec/mongo/operation/write/command/insert_spec.rb
```diff
@@ -56,6 +56,31 @@
...
+      let(:spec) do
+        { :documents     => documents,
+          :db_name       => db_name,
+          :coll_name     => coll_name,
+          :ordered       => true
+        }
+      end
+
+      it 'does not include write concern in the selector' do
+        expect(op.send(:selector)[:writeConcern]).to be_nil
```
durran: Should the write concern be `nil` or `{}` here? The spec makes it look like it should be the server's default write concern, which appears to be `{}` and not `nil`. Maybe verify this with craig?

https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#servers-default-writeconcern

estolfo: I think we're just not supposed to send write concern if the user doesn't specify it (and we're using write commands). If we don't send write concern, the server's default is used. Bernie verified this the other day:

emily
For write commands, we shouldn't send write concern unless the user explicitly sets one (at whatever level we allow it - client/db/collection)
Bernie
Yep

I'll double check with Craig though.

durran: I was just double checking... If that's the case then LGTM.

craiggwilson: That's correct. If the user didn't specify anything, don't send anything. Sending anything has the potential to overwrite the server's default write concern. If the user did specify anything (even what we think of as the default ({w: 1}), you must send it along. 

If you are allowing your users to specify both nil and {}, then I'd take both of these to mean the exact same thing, the server's default.

durran:
Ship it. :)

estolfo:
It's actually not quite right yet - the default write concern shouldn't be {w: 1}, it should be just sending GLE. Sending GLE with no options and sending {w: 1} aren't quite the same thing.

PR #720, opened 2015-10-28
RUBY-1065: Raise error if bulk cannot split operation

durran:


estolfo:
Just tested locally, LGTM

PR #721, opened 2015-10-28
RUBY-1064: Add maxAwaitTimeMS

durran:
- Add the ability to add the option in the view fluid interface.
- Adds maxTimeMS to the getmore command is maxAwaitTimeMS is set as an
  option.
- Ignores the option in all other cases.


estolfo:
lgtm

PR #722, opened 2015-10-29
RUBY-1066 Don't override previously-defined Hash#assert_valid_keys

TiteiKo:
Don't override ActiveSupport assert_valid_keys if defined

See https://jira.mongodb.org/browse/RUBY-1066


estolfo:
thanks @TiteiKo !

PR #725, opened 2015-11-10
RUBY-1056: Handle custom id generators

durran:
Can pass an option to client `:id_generator` that's value is an object that responds to `#generate`.


lib/mongo/collection.rb
```diff
@@ -341,7 +341,7 @@ def insert_one(document, options = {})
           :coll_name => name,
           :write_concern => write_concern,
           :bypass_document_validation => !!options[:bypass_document_validation],
-          :options => options
+          :id_generator => client.options[:id_generator]
```
estolfo: we are sure we don't need to pass :options to Operation::Write::Insert#initialize ?

spec/mongo/collection_spec.rb
```diff
@@ -825,6 +856,37 @@
...
+
+      after do
+        custom_client.close
+      end
+
+      it 'inserts with the custom id' do
+        expect(custom_collection.find.first[:_id]).to eq(1)
+      end
+    end
+
```
estolfo: did you mean to repeat the spec?

durran: Yeah - one is insert many and one is insert one.

estolfo: ok, sorry, I missed that.

PR #728, opened 2015-11-20
Benchmarking

DiannaHohensee:


Rakefile
```diff
@@ -41,3 +41,33 @@ namespace :docs do
...
+    p "MIDDLEWEIGHT BENCHMARK"
+    middleweight_benchmark!
+  end
+
+  task :heavyweight do
+    p "HEAVYWEIGHT BENCHMARK"
+    heavyweight_benchmark!
+  end
+
+  task :run_all_benchmarks => [:featherweight, :lightweight, :middleweight, :heavyweight]
```
estolfo: you can just call this :all because it's already namedspaced to :benchmark.
So you would run it like:

rake benchmark:all

profile/feather_weight_benchmark.rb
```diff
@@ -0,0 +1,83 @@
...
+# All BSON Types
+#
+# The featherweight benchmark is intended to test BSON serialization/deserialization
+#
+##
+def featherweight_benchmark!
+  #bench_helper = BenchmarkHelper.new('perftest','corpus')
+  bench_helper = BenchmarkHelper.new('foo','bar')
+  database = bench_helper.database
+  collection = bench_helper.collection
```
estolfo: I don't think you need database or collection, right?

profile/feather_weight_benchmark.rb
```diff
@@ -0,0 +1,83 @@
...
+  ##
+  flat_data = BenchmarkHelper.load_array_from_file('FLAT_BSON.txt')
+
+  first = Benchmark.bm do |bm|
+    bm.report('Featherweight::Common Flat BSON') do
+      flat_data.each do |doc|
+        BSON::Document.from_bson(  BSON::Document.new(doc).to_bson  )
+      end
+    end
+  end
```
estolfo: I would put each one of these chunks into a separate method and define the file names as constants at the top.

```ruby
FILES = ['FLAT_BSON.txt', 'DEEP_BSON.txt', 'FULL_BSON',txt]

def featherweight_benchmark!
  results = FILES.each do |file|
    data = BenchmarkHelper.load_array_from_file(file)

    Benchmark.bm do |bm|
      bm.report('..') do
       data.each do |doc|
         BSON::Document.from_bson(  BSON::Document.new(doc).to_bson  )
       end
     end
  end
end
```

not perfect, but hopefully you get what I'm saying.

profile/heavy_weight_benchmark.rb
```diff
@@ -0,0 +1,212 @@
...
+  #          Concurrency is encouraged.
+  #
+  # - Drop the 'perftest' database
+  ##
+  database.drop
+
+  # The directory name, or path to the directory, in which the LDJSON files are expected to be
+  ldjson_data_files_directory = "LDJSON_data_file_directory"
+  # Array of expected file names
+  ldjson_multi_files = Array.new(100) {|i| "LDJSON%02d.txt" % (i+1) }
```
estolfo: String interpolation is more commonly done like:

```ruby
"LDJSON#{i+1}.txt"
```

profile/heavy_weight_benchmark.rb
```diff
@@ -0,0 +1,212 @@
...
+      ldjson_multi_files.each do |file_name|
+        threads << Thread.new do
+          ldjson_multi_data = BenchmarkHelper.load_array_from_file("#{ldjson_data_files_directory}/#{file_name}")
+          collection.insert_many(ldjson_multi_data, ordered: false)
+        end
+      end
+
+      threads.each { |t| t.join }
+    end
+  end
```
estolfo: Same thing here, and for the other files as well: perhaps you can find a way to make each phase a separate method, with the differences between each phase the arguments to that method. That way, you don't have to repeat lines like 

```ruby
database.drop
```
In the method, you can just make that the first line.
The biggest advantage of this is that in the future, if more tests are added, we will only have to add a single element (like a file name) to an array constant instead of writing a whole bunch of code. It's not uncommon for these driver-wide tests are modified or added to with time.

profile/heavy_weight_benchmark.rb
DiannaHohensee: The spec uses "gridfstest%03d". If I do "LDJSON#{i+1}.txt", it'll be different. Doesn't really matter either way: should it be overruled?

profile/benchmark_helper.rb
```diff
@@ -0,0 +1,185 @@
+require 'mongo'
+require 'json'
+require 'FileUtils'
+
+class BenchmarkHelper
+  # Accessor methods for
```
estolfo: I think you're missing the description here. You can put a @ since tag as well.

profile/benchmark_helper.rb
```diff
@@ -0,0 +1,185 @@
...
+  attr_reader :database, :collection
+
+
+  # Initializes a client connection, creating a database and a collection.
+  #
+  # @example Initialize a client connection.
+  #   BenchmarkHelper.initialize_collection("testing")
```
estolfo: Needs to be updated

profile/benchmark_helper.rb
```diff
@@ -0,0 +1,185 @@
...
+  #
+  # @param [ String, Symbol ] collection_name The name of the collection.
+  # @param [ String, Symbol ] database_name The name of the database.
```
estolfo: You should put these in order in which they are passed to the method

profile/benchmark_helper.rb
```diff
@@ -0,0 +1,185 @@
...
+  #
+  # @return [ Mongo::Collection ] The collection.
```
estolfo: initialize methods don't have a return (they return self)

profile/benchmark_helper.rb
```diff
@@ -0,0 +1,185 @@
...
+    # @example Load a file into a string
+    #   BenchmarkHelper.load_string_from_file("GRIDFS_LARGE.txt")
+    #
+    # @param [ String ] data_file_name The name of the data file.
+    #
+    # @return [ String ] A string of all the file data.
+    #
+    # @since 2.2.1
+    def load_string_from_file(data_file_name)
+      file = File.open(data_file_name, "rb")
```
estolfo: is there a reason you passed File.open a block in load_array_from_file and not here? I would go with the block approach.

profile/benchmark_helper.rb
```diff
@@ -0,0 +1,185 @@
...
+
+  class << self
+
+
+    # Load a file into a string
+    #
+    # @example Load a file into a string
+    #   BenchmarkHelper.load_string_from_file("GRIDFS_LARGE.txt")
+    #
+    # @param [ String ] data_file_name The name of the data file.
```
estolfo: Maybe just call this file_name

profile/benchmark_helper.rb
```diff
@@ -0,0 +1,185 @@
...
+    #
+    # @return [ String ] A string of all the file data.
+    #
+    # @since 2.2.1
+    def load_string_from_file(data_file_name)
+      file = File.open(data_file_name, "rb")
+      contents = file.read
+      file.close
+      return contents
```
estolfo: You don't need explicit returns in Ruby. The result of the last expression evaluated will be returned from a method.

profile/benchmark_helper.rb
```diff
@@ -0,0 +1,185 @@
...
+    # Write documents from an array into the specified file.
+    # One document written per line in the file
+    #
+    # @example Write the data into the specified file
+    #   BenchmarkHelper.write_documents_to_file("foo.txt", data)
+    #
+    # @param [ String ] path The path to the file to which to write the data.
+    # @param [ Array<Hash> ] data An array of document data.
+    #
+    # @return [ nil ]
```
estolfo: if the return is nil, you don't need to note it

profile/benchmark_helper.rb
```diff
@@ -0,0 +1,185 @@
...
+    # @example Calculate the composite score
+    #   BenchmarkHelper.mmabench_composite_score(3,5,6,2)
+    #
+    # @param [ Array<Double> ] scores The numbers to be averaged.
+    #
+    # @return [ Double ] The composite score.
+    #
+    # @since 2.2.1
+    def mmabench_composite_score(scores)
+      scores.inject(0.0) { |sum, score| sum + score } / scores.size
```
estolfo: nice use of inject = )

profile/benchmark_helper.rb
```diff
@@ -0,0 +1,185 @@
...
+    # @return [ Array<Double> ] The 10th, 25th, 50th, 75th, 90th, 95th, 98th and 99th percentiles.
+    #
+    # @since 2.2.1
+    def percentile_values(scores)
+      percentiles = [10,20,50,75,90,95,98,99]
+      percentiles.map { |percentile| percentile_value(percentile, scores) }
+    end
+
+
+    def MB_per_second(data_results_array, mb)
```
estolfo: I think we're missing documentation here.

profile/feather_weight_benchmark.rb
```diff
@@ -0,0 +1,63 @@
...
+
+
+
+# Runs encode/decode benchmark on a dataset a given number of times
+#
+# @param [ Integer ] data_file_name Name of dataset file.
+# @param [ Integer ] reps Number of repetitions of the benchmark to run.
+# @param [ Integer ] label Benchmark label.
+#
+# @return [ [Array<Integer>, Double, String] ] An array of benchmark wall clock time results, the size of the dataset in MB, test label
```
estolfo: Break this up into two lines

profile/feather_weight_benchmark.rb
```diff
@@ -0,0 +1,63 @@
...
+#
+# @since 2.2.1
+def encode_decode_bson_helper(data_file_name, reps, label)
+  data = BenchmarkHelper.load_array_from_file(data_file_name)
+  data_file_size = File.size(data_file_name)
```
estolfo: I believe this returns the size of the file name. Is that what you want?

http://ruby-doc.org/core-2.1.4/File.html#method-c-size

profile/heavy_weight_benchmark.rb
```diff
@@ -0,0 +1,266 @@
...
+  bench_helper = BenchmarkHelper.new('foo','bar', 200)
+  database = bench_helper.database
+  collection = bench_helper.collection
+
+  results = []
+  results << ldjson_multi_file_import(database, collection, benchmark_reps)
+  results << ldjson_multi_file_export(database, collection, benchmark_reps)
+  results << gridfs_multi_file_upload(database, benchmark_reps)
+  results << gridfs_multi_file_download(database, benchmark_reps)
+  results
```
estolfo: The line above (L45) returns results so you don't need this line.

profile/feather_weight_benchmark.rb
DiannaHohensee: Seems to get the file size in bytes. The description of the method is just poorly worded.

2.2.1 :002 > File.size("profile.rb")
 => 999
2.2.1 :003 > File.size("TWITTER")
 => 276923862

DiannaHohensee:
@estolfo Updated with corrections

estolfo:
great, thanks!
I still have to look over two more files and will let you know when I'm done.

profile/middle_weight_benchmark.rb
```diff
@@ -0,0 +1,233 @@
...
+#
+# @since 2.2.1
+def find_many_and_empty_the_cursor(database, collection, reps)
+  database.drop
+  twitter_data = BenchmarkHelper.load_array_from_file(TWITTER)
+  data_file_size = File.size(TWITTER)
+  #twitter_data_size = twitter_data.size
+
+  #10000.times do |i|
+  #  collection.insert_one( twitter_data[i] ) if (i < twitter_data_size)
```
estolfo: if we uncomment this at some point, we should figure out if 10000 or twitter_data_size is great before doing the inserts, as the 'if' statement would factor into the time calculation.

profile/middle_weight_benchmark.rb
```diff
@@ -0,0 +1,233 @@
...
+  data_file_size = File.size(GRIDFS_LARGE)
+
+  tms_results = []
+  reps.times do
+    database.drop
+    tms_results << Benchmark.bm do |bm|
+      bm.report('Middleweight::GridFS upload') do
+        100.times do |i|
+          database.fs.insert_one(
+              Mongo::Grid::File.new(gridfs_large_data_string, :filename => "gridfstest%03d" % (i+1))
```
estolfo: Use the upload_from_stream method instead.
I would map out the filenames to an array also so the string interpolation doesn't factor into the time calculation.

profile/middle_weight_benchmark.rb
```diff
@@ -0,0 +1,233 @@
...
+  gridfs_large_data_string = BenchmarkHelper.load_string_from_file(GRIDFS_LARGE)
+  data_file_size = File.size(GRIDFS_LARGE)
+  gridfs_file = Mongo::Grid::File.new( gridfs_large_data_string, :filename => GRIDFSTEST )
+  database.fs.insert_one( gridfs_file )
+
+  tms_results = Benchmark.bm do |bm|
+    reps.times do
+      bm.report('Middleweight::GridFS download') do
+        100.times do
+          stream = StringIO.new
```
estolfo: maybe figure out a way to not have the instantiation of a new stream factor into the timing

profile/mma_bench.rb
```diff
@@ -0,0 +1,59 @@
...
+def mma_bench
+
+  # Run benchmarks
+  featherweight = featherweight_benchmark(3)
+  lightweight = lightweight_benchmark(3)
+  middleweight = middleweight_benchmark(3)
+  heavyweight = heavyweight_benchmark(3)
+
+  # FeatherBench
+  feather_bench_results = featherweight.map { |test_results| BenchmarkHelper.MB_per_second(test_results[0], test_results[1]) }
```
estolfo: I would use the do..end syntax for a block here so it's easier to read.

profile/feather_weight_benchmark.rb
estolfo: oh, haha. I think it's clear in the documentation, you're right. I just had a Amelia Bedelia moment.

estolfo:
ok @DiannaHohensee I'm done looking over the files if you want to update some things according to my comments.

profile/benchmark_helper.rb
```diff
@@ -0,0 +1,194 @@
+require 'mongo'
+require 'json'
+require 'FileUtils'
```
estolfo: I don't think you need this.

profile/middle_weight_benchmark.rb
DiannaHohensee: Actually, the insertion code here isn't part of the timing. That starts with 'Benchmark.bm do...'

profile/middle_weight_benchmark.rb
DiannaHohensee: I added an array of filenames. But wouldn't uploading from a stream take more time since it's reading the data from a file (I/O operation) and then sending the data to the DB? This way the data is already stored in a string.

DiannaHohensee:
@estolfo Made corrections, or commented.

DiannaHohensee:
So after fiddling with the dataset sizes for a bit, my conclusion is that perhaps the dataset sizes need to be run on a dedicated computer of some sort. For the LARGE_DOC dataset, the data being loaded into an array is 2.7 GB. I don't think the part that is timed takes that long, so much as loading the data: 20 document file instead of 1,000 only takes 4 seconds (timed), but a takes a long while (untimed) to load the data before the timing begins. Loading data is slow in a few other benchmarks, before the timing begins.

Other outstanding things to do: implement MongoDB extended JSON parsers, or change dataset to YAML, so featherweight benchmarks can be run; upload the final dataset files to some accessible repository because the files are too big to be put with the driver code -- would slow gem download excessively.

profile/middle_weight_benchmark.rb
estolfo: ah, of course

profile/benchmark_helper.rb
estolfo: this is making our travis ci tests fail: https://travis-ci.org/mongodb/mongo-ruby-driver/builds/93085864

PR #729, opened 2015-11-27
Proof of concept to have a connection pool per host/port pair. RUBY-1073

jonhyman:
Hey @estolfo, take a look at this. I tested it out in our staging environment and it cut down the connections to 1 per cluster host instead of 1 per database. I also did some basic tests on a production box and it worked great there as well.

This is a bit messy in terms of how it works because the authentication is stored in the `Connection` instead of injected in, so I had to add that. Also, when I run all the specs, I get a failure in this test, but if I run the entire file or spec itself, it passes. My guess is that there's some sort of state leakage now between the test run or something via `ConnectionPool.pools`, but I don't know much about that part of the code.

```
  1) Mongo::Grid::FSBucket when a read stream is opened #open_download_stream_by_name when a block is provided when revision is 0 returns the original stored file
     Failure/Error: expect(io.string).to eq('hello 1')

       expected: "hello 1"
            got: "hello 2"

       (compared using ==)
     # ./spec/mongo/grid/fs_bucket_spec.rb:741:in `block (6 levels) in <top (required)>'
```

### DIFF
--- a/lib/mongo/server/connection.rb
+++ b/lib/mongo/server/connection.rb
@@ -38,28 +38,13 @@ module Mongo
       # @since 2.1.0
       PING_BYTES = PING_MESSAGE.serialize.to_s.freeze
 
-      # @return [ Mongo::Auth::CR, Mongo::Auth::X509, Mongo::Auth:LDAP, Mongo::Auth::SCRAM ]
-      #   authenticator The authentication strategy.
-      attr_reader :authenticator
+      attr_reader :authentication_by_db
 
       def_delegators :@server,
                      :features,
                      :max_bson_object_size,
                      :max_message_size,
                      :mongos?
-
-      # Is this connection authenticated. Will return true if authorization
-      # details were provided and authentication passed.
-      #
-      # @example Is the connection authenticated?
-      #   connection.authenticated?
-      #
-      # @return [ true, false ] If the connection is authenticated.
-      #
-      # @since 2.0.0
-      def authenticated?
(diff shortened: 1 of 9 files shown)


lib/mongo/server/connection.rb
```diff
@@ -48,19 +46,6 @@ class Connection
...
 
-      # Is this connection authenticated. Will return true if authorization
-      # details were provided and authentication passed.
-      #
-      # @example Is the connection authenticated?
-      #   connection.authenticated?
-      #
-      # @return [ true, false ] If the connection is authenticated.
-      #
-      # @since 2.0.0
```
jonhyman: This wasn't used anywhere but in a test.

lib/mongo/server/connection.rb
```diff
@@ -38,9 +38,7 @@ class Connection
       # @since 2.1.0
       PING_BYTES = PING_MESSAGE.serialize.to_s.freeze
 
-      # @return [ Mongo::Auth::CR, Mongo::Auth::X509, Mongo::Auth:LDAP, Mongo::Auth::SCRAM ]
-      #   authenticator The authentication strategy.
-      attr_reader :authenticator
```
jonhyman: The authenticator reader method was only used in a test.

jonhyman:
Moved to https://github.com/jonhyman/mongo-ruby-driver/tree/feature/runner-dead-unauthorized-220

PR #730, opened 2015-11-27
Stop monitor on server garbage collection.

durran:
- Stops server monitor when Server is marked for GC.
- Scopes connection pools to cluster and fixes the Address equality as hash keys bug (was leaking pools)

### DIFF
--- a/lib/mongo/address.rb
+++ b/lib/mongo/address.rb
@@ -63,6 +63,10 @@ module Mongo
       host == other.host && port == other.port
     end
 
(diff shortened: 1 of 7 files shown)


jonhyman:
@durran are you going to support multiple authentications per cluster? This branch still creates `num_databases` connection pools to a `host:ip` cluster, since each one gets created as a new `Client` which creates a new `Cluster` (and a new `ConnectionPool`).

Just trying to figure out if I s

PR #734, opened 2015-12-15
[minor] fixing build matrix, adding ruby-head as allowed fail

brandonblack:
After all these years... I finally figured out how to configure this build matrix correctly. I remembered how wrongly I configured it here and wanted to share the solution.


brandonblack:
Nevermind. It's still a PITA with 4 different MongoDB versions. I tried.

PR #736, opened 2015-12-18
Give Grid::File ability to accept IO objects

janko-m:
Sometimes we have an IO that we don't want to read whole into memory, so we can transfer the chunking logic into IO terms and allow both strings and IOs.

### DIFF
--- a/lib/mongo/grid/file.rb
+++ b/lib/mongo/grid/file.rb
@@ -30,9 +30,6 @@ module Mongo
       # @return [ Array<Chunk> ] chunks The file chunks.
       attr_reader :chunks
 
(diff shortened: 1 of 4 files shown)


janko-m:
I actually realized that in my use case I would have to use internals too much, so I widened the change. The change now goes one step further, allowing `Grid::File` objects to accept an IO, and to internally handle this efficiently, to never load the whole file contents into memory. I was careful to keep all backwards compatibility, e.g. `Grid::File#data` still exists, but now it's loaded lazily.

estolfo:
thanks @janko-m, this is great!

PR #738, opened 2015-12-22
Don't store authenticator on connection instance.

durran:
This changes authenticator setup in the server connection to happen at
connect time, instead of connection instantiation. This fixes the
possibility of a running application with a connection open to a pre-3.0
server to get auth errors if the connection to the address was closed
and then reconnected against a server of 3.0 and higher and a different
auth mechanism.

The scenario to reproduce:
- Start an application against a MongoDB 2.6 server with MONGODB-CR
  authentication with no auth_mech option specified on the client.
- Stop the running server, but leave the application running. The first
  exception in the application will disconnect the connection.
- Upgrade the server to 3.0 and upgrade users to SCRAM-SHA-1. Restart
  the server.
- The next application call would connect to the server, and attempt to
  authenticate with MONGODB-CR instead of SCRAM-SHA-1 since the
  authenticator was previously set on the instance. An auth error would
  happen on this call.

This fixes the above scenario.

### DIFF
--- a/lib/mongo/server/connection.rb
+++ b/lib/mongo/server/connection.rb
@@ -38,28 +38,11 @@ module Mongo
       # @since 2.1.0
       PING_BYTES = PING_MESSAGE.serialize.to_s.freeze
 
-      # @return [ Mongo::Auth::CR, Mongo::Auth::X509, Mongo::Auth:LDAP, Mongo::Auth::SCRAM ]
-      #   authenticator The authentication strategy.
-      attr_reader :authenticator
-
       def_delegators :@server,
                      :features,
                      :max_bson_object_size,
                      :max_message_size,
                      :mongos?
-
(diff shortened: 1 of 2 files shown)


durran:
@jonhyman Maybe you might have run into this as well, since you are load balancing your mongos servers and were doing a 2.6 to 3.0 upgrade?

jonhyman:
We haven't yet done the upgrade to 3.0 (we were waiting on 3.0.8). We have seen random not authorized errors under high load, so in my fork, our `retryable.rb` catches not authorized and re-authenticates.

Though, my fork works the same as this patch because I set up the authentication each time `connect!` is called https://github.com/jonhyman/mongo-ruby-driver/blob/feature/runner-dead-unauthorized-220/lib/mongo/server/connection.rb#L82

PR #739, opened 2015-12-31
A couple more retries.

jonhyman:
Some more errors we've seen during failovers or crashes. I've added them to retry on my fork.

* `Mongo::Error::OperationFailure: unknown replica set d303f52f76534680956e00a707bbca89 (71)`
* `Mongo::Error::OperationFailure: dbclient error communicating with server: REDACTED (10278)`

### DIFF
--- a/lib/mongo/error/operation_failure.rb
+++ b/lib/mongo/error/operation_failure.rb
@@ -37,7 +37,9 @@ module Mongo
         'error querying',
         'could not get last error',
         'connection attempt failed',
-        'interrupted at shutdown'
+        'interrupted at shutdown',
+        'unknown replica set',
+        'dbclient error communicating with server'
       ].freeze
 
       # Can the operation that caused the error be retried?


estolfo:
Hey @jonhyman I was wondering if you could tell me the scenario in which you got the  `Mongo::Error::OperationFailure: unknown replica set d303f52f76534680956e00a707bbca89 (71)` error and how that was a temporary error (one that went away after retrying)
I want to make sure it's not a server bug. Thanks!

jonhyman:
We saw it during a step down, caused some failures but worked after the reconnect happened. It was the first time I've seen that error, though, and earlier in that day the shard had had a ton of QRs and QWs for a reason we still couldn't figure out.

behackett:
And which version of MongoDB was it?

jonhyman:
2.6.11.

estolfo:
Thanks, I double-checked with our server team and it's expected this error would make its way to the user, so not a mongos bug.

estolfo:
Hey Jon, I just wanted to let you know that there is a server ticket to track improving the error message for when a mongos is unable to connect to a particular shard-- which is the scenario in which you received the "unknown replica set" error message. Here it is, if you're interested in tracking it. https://jira.mongodb.org/browse/SERVER-22107

jonhyman:
Thanks!

PR #741, opened 2016-01-12
RUBY-1079 Can't use read preference when count documents

drobune:


estolfo:
The count command should not take a read preference as an option, it will always use the read preference of the collection. You can see our specifications [here](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#read) -- check out the CountOptions section. 

If you'd like to use a read preference for the count command, you can set the option on the collection, then call #count on it directly or on a view created from it.

I'll update our documentation here to remove the :read option, as it's a mistake. As you can see in the [documentation of the count command directly on the collection](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/collection.rb#L260-L278), there is no :read option. This is the correct documentation.

Thanks!
Emily

drobune:
make sense...
Thanks for your reply!

PR #742, opened 2016-01-15
Driver Benchmark tests

estolfo:


profile/benchmarking/multi_doc.rb
```diff
@@ -0,0 +1,179 @@
...
+          Benchmark.realtime do
+            collection.insert_many(docs)
+          end
+        end
+        client.database.drop
+        Benchmarking.median(results)
+      end
+
+      def client
+        @client ||= Mongo::Client.new(["localhost:27017"], database: 'perftest')
```
durran: Can add option `:monitoring => false` to speed up as well.

estolfo: ah, good point. Thanks

durran:
Are we commiting the json data files as well?

estolfo:
So far I wasn't committing the data files, as some are quite large and I think they're on s3. I can commit the smaller ones though.

PR #746, opened 2016-01-26
SPEC-222: SDAM Monitoring

durran:


spec/support/sdam_monitoring/standalone.yml
```diff
@@ -0,0 +1,74 @@
...
+        -
+          server_opening_event:
+            topologyId: "42"
+            address: "a:27017"
+        -
+          server_description_changed_event:
+            topologyId: "42"
+            address: "a:27017"
+            previousDescription:
+              address: "127.0.0.1:27017"
```
jyemin: This address surprises me?  Where does it come from?  Why would a driver have that as the default address?  I'd expect instead that the previousDescriptions's address would be the same as the newDescription's address, but with type "Unknown" (as it is).

spec/support/sdam_monitoring/standalone.yml
```diff
@@ -0,0 +1,74 @@
...
+              type: "Standalone"
+        -
+          topology_description_changed_event:
+            topologyId: "42"
+            previousDescription:
+              type: "Unknown"
+              setName: null
+              maxSetVersion: null
+              maxElectionId: null
+              servers: [ "127.0.0.1:27017" ]
```
jyemin: Same here.  I would expect the value to be [ "a:27017"]

spec/support/sdam_monitoring/replica_set_with_removal.yml
```diff
@@ -0,0 +1,88 @@
...
+          server_closed_event:
+            topologyId: "42"
+            address: "b:27017"
+        -
+          topology_description_changed_event:
+            topologyId: "42"
+            previousDescription:
+              type: "Unknown"
+              setName: null
+              maxSetVersion: null
```
jyemin: We should consider removing some of these fields.  While it's true they are all part of the TopologyDescription as defined in the specification, most are not useful to clients and should be considered implementation details of the SDAM implementation.  The only ones generally useful to clients are type and setName and servers. 

As for servers, the spec says that it's a list of ServerDescription, not a list of server address strings.  And that's necessary to actually implement the hasReadable/WritableServer methods.  To avoid repetition in the YAML, you can use node anchors (http://yaml.org/spec/1.2/spec.html#id2785586)

spec/support/sdam_monitoring/standalone.yml
durran: The SDAM spec states the default is one server of localhost:27017: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#topologydescription

jyemin: Two comments:

* 127.0.0.1 is not the same as localhost
* localhost:27017 is the default but the connection string doesn't allow for a default host.  It's specified as a:27017

spec/support/sdam_monitoring/standalone.yml
durran: I thought I had read in the SDAM spec as well about the default for the server description but reading it again I don't see it... So maybe leave it out? I think some implementations will default with an "empty" description and some might default with the address provided and type unknown.

spec/support/sdam_monitoring/replica_set_with_removal.yml
durran: Sounds good, I'll remove the non-important ones.

spec/support/sdam_monitoring/standalone.yml
jyemin: The address should be "a:27017" in previousDescription and type "Unknown".  I don't see how it makes sense to have an empty address for a server description changed event.  The address is part of the top-level event metadata, so both descriptions should share it.

durran:
@jyemin Updated per the comments, I've reduced the fields to the important ones and fixed the defaults.

spec/support/sdam_monitoring/replica_set_with_removal.yml
```diff
@@ -0,0 +1,64 @@
...
+          server_closed_event:
+            topologyId: "42"
+            address: "b:27017"
+        -
+          topology_description_changed_event:
+            topologyId: "42"
+            previousDescription:
+              topologyType: "Unknown"
+              setName: null
+              servers: [ "a:27017" ]
```
jyemin: This should have "b:27017" as well.

It should also be an array of ServerDescription instead of array of string

estolfo:
Replaced by: https://github.com/mongodb/mongo-ruby-driver/pull/834

PR #750, opened 2016-02-11
add ability to specify SCRAM-SHA-1 as authMechanism on uri

mattberther:


estolfo:
thanks @mattberther !

PR #758, opened 2016-04-01
RUBY-1101 SSL and Unix sockets are only connectable if they've been created

estolfo:


durran:
LGTM - Not sure how to easily test it but the code looks good.

estolfo:
It's really difficult to test because you must establish the connection first so that the Connection object has a socket associated with it. Then you must force the server to become unreachable so the SSL socket's ```@socket``` instance variable is not set. A connection exception would happen in [this line](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/socket/ssl.rb#L57), then ```@socket``` won't get set in [this line](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/socket/ssl.rb#L58).

PR #759, opened 2016-04-12
RUBY-1088: Sync Pool#with_connection

durran:
Not only the queue needs to be synchronized, the #with_connection which checks out from the queue also needs to since the code path could have entry from multiple threads at the same time.

Refactor of the `ConnectionPool` introduced this in 2.2.x

### DIFF
--- a/lib/mongo/server/connection_pool.rb
+++ b/lib/mongo/server/connection_pool.rb
@@ -23,9 +23,6 @@ module Mongo
     class ConnectionPool
       include Loggable
 
-      # @return [ Mutex ] mutex The mutex used for synchronization.
-      attr_reader :mutex
-
       # @return [ Hash ] options The pool options.
       attr_reader :options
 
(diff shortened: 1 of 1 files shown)


estolfo:
LGTM - I tested with a script that could reproduce https://jira.mongodb.org/browse/RUBY-1088 and it no longer raises exceptions with this change.

lib/mongo/server/connection_pool.rb
```diff
@@ -23,6 +23,9 @@ class Server
     class ConnectionPool
       include Loggable
 
+      # @return [ Mutex ] mutex The mutex used for synchronization.
+      attr_reader :mutex
```
estolfo: does this need to be attr_reader?

durran: Doesn't have to be - we could just reference the instance variable directly. Was just following the pattern from the other mutexes.

mperham:
-1 now only one thread can use a connection at a time.

durran:
@mperham Yeah you're right - wasn't even the core issue as well with the bug.

mperham:
You can use my connection_pool gem, it's battle hardened after 3+ years running with Sidekiq now...

PR #760, opened 2016-04-15
RUBY-741 Make connection pool queue LIFO

durran:
As per the discussion on our driver's slack channel, this changes our queue from FIFO to LIFO. Also the queue needs to be LIFO in order to support `maxIdleTimeMS` configuration and idle connection reaping down to the `minPoolSize` configuration option.

### DIFF
--- a/lib/mongo/server/connection_pool/queue.rb
+++ b/lib/mongo/server/connection_pool/queue.rb
@@ -170,7 +170,7 @@ module Mongo
         def dequeue_connection
           deadline = Time.now + wait_timeout
           loop do
-            return queue.pop unless queue.empty?
+            return queue.shift unless queue.empty?
             connection = create_connection
             return connection if connection
             wait_for_next!(deadline)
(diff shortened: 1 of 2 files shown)


matsimitsu:
Is this driver slack channel a public one?

estolfo:
Hey @matsimitsu 
It's not public, unfortunately. I also wanted to let you know that we know you've been seeing intermittent "not authorized" errors and we're close to figuring out why. We'll have an update on the related ticket soon. Sorry for the delay!

durran:
@matsimitsu Basically with the current FIFO implementation, if there is a spike in load which causes the connection pool size to hit `maxPoolSize`, the connection pool size will always stay at that size. If we make the pool LIFO, and the same spike happens but then returns to normal levels later, we can disconnect the idle connections in the pool after `maxIdleTimeMS`, thus reducing client connections and server resources.

estolfo:
replaced by https://github.com/mongodb/mongo-ruby-driver/pull/875

PR #761, opened 2016-04-25
[v 2.3.0] RUBY-1092 Validate keys when inserting documents

estolfo:
This will pass tests when [this pull request](https://github.com/mongodb/bson-ruby/pull/59) against the bson gem is merged.

### DIFF
--- a/lib/mongo/protocol/serializers.rb
+++ b/lib/mongo/protocol/serializers.rb
@@ -50,7 +50,7 @@ module Mongo
         # @param value [ String ] Header value to be serialized.
         #
         # @return [ String ] Buffer with serialized value.
-        def self.serialize(buffer, value)
+        def self.serialize(buffer, value, validating_keys = nil)
           buffer.put_bytes(value.pack(HEADER_PACK))
         end
 
@@ -76,7 +76,7 @@ module Mongo
         # @param value [ String ] The string to be serialized.
         #
         # @return [ String ] Buffer with serialized value.
-        def self.serialize(buffer, value)
+        def self.serialize(buffer, value, validating_keys = nil)
           buffer.put_cstring(value)
         end
       end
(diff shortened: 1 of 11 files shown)


lib/mongo/protocol/serializers.rb
```diff
@@ -50,7 +50,7 @@ module Header
         # @param value [ String ] Header value to be serialized.
         #
         # @return [ String ] Buffer with serialized value.
-        def self.serialize(buffer, value)
+        def self.serialize(buffer, value, validating_keys = nil)
```
estolfo: Not sure if these method signatures should be this instead:

```ruby
def self.serialize(buffer, value, validating_keys = BSON::Config.validating_keys?)
```
That would make a dependency on bson >= 4.1.0

lib/mongo/protocol/serializers.rb
```diff
@@ -158,9 +158,9 @@ module Document
         # @param value [ Hash ] Document to serialize as BSON.
         #
         # @return [ String ] Buffer with serialized value.
-        def self.serialize(buffer, value, max_bson_size = nil)
+        def self.serialize(buffer, value, max_bson_size = nil, validating_keys = nil)
           start_size = buffer.length
-          value.to_bson(buffer)
+          validating_keys.nil? ? value.to_bson(buffer) : value.to_bson(buffer, validating_keys)
```
estolfo: if the method signature is

```ruby
def self.serialize(buffer, value, validating_keys = BSON::Config.validating_keys?)
```
then this line can be simply:
```ruby
value.to_bson(buffer, validating_keys)
```

lib/mongo/protocol/serializers.rb
durran: Yes I would reaise the BSON dependency to 4.1 which would force this change to be in driver 2.3.0

estolfo: ok, thanks

durran:
I would bump the dependency on BSON as you commented and make the changes noted, then put this in 2.3 since the dependency was bumped.

estolfo:
ok, updated. The jira ticket tags this for version 2.3.0.

PR #763, opened 2016-05-05
Add dependency on 'pry'

pvalena:
'pry' is required and used by mongo_console executable[1], but is not installed by running `gem install mongo`, therefore leaving the executable unusable. 

[1] https://github.com/mongodb/mongo-ruby-driver/blob/15eb4b28462870bc8d54925e745c8da0b73543c7/bin/mongo_console#L4

### DIFF
--- a/mongo.gemspec
+++ b/mongo.gemspec
@@ -31,4 +31,5 @@ Gem::Specification.new do |s|
   s.bindir            = 'bin'
 
   s.add_dependency 'bson', '~> 4.0'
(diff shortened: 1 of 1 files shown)


estolfo:
thanks!

estolfo:
Yes, it is required by the mongo_console executable... but what you've proposed is a better solution. I've opened [a ticket](https://jira.mongodb.org/browse/RUBY-1127) in our Jira RUBY project to fix this. Thanks!

PR #765, opened 2016-05-25
[v 3.0.0] let authentication errors subclass Mongo::Error

MartinNowak:
- allows to uniformly deal with all MongoDB errors

### DIFF
--- a/lib/mongo/auth.rb
+++ b/lib/mongo/auth.rb
@@ -71,7 +71,7 @@ module Mongo
     # Raised when trying to get an invalid authorization mechanism.
     #
     # @since 2.0.0
-    class InvalidMechanism < RuntimeError
+    class InvalidMechanism < ::Mongo::Error
 
       # Instantiate the new error.
       #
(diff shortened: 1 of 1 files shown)


estolfo:
Hi @MartinNowak 
Thanks for this pull request. It looks good but I won't be able to merge it until a major release, as it's a behavior change; if anyone is catching a RuntimeError in their applications, this change would affect them.

MartinNowak:
> Thanks for this pull request. It looks good but I won't be able to merge it until a major release, as it's a behavior change; if anyone is catching a RuntimeError in their applications, this change would affect them.

Is there some possibility to inherit from both exception classes `Mongo::Error` and `RuntimeError`?
Might not make sense to let `Mongo::Error` inherit from `RuntimeError` just for this fix, does it?

estolfo:
hmm, I don't see how we'd be able to do this without it affecting behavior in users' applications. We'll have to wait until the next major release, unfortunately.

estolfo:
I'm going to close this PR for now but it can be reopened when we are working towards the 3.0.0 release. Please find the corresponding JIRA ticket [here](https://jira.mongodb.org/browse/RUBY-1274).

PR #766, opened 2016-05-31
[v 2.3.0] RUBY-1104 Add CursorReaper for periodically sending kill cursor operations

estolfo:
Todo:
- ~~Consolidate cursor ids into one op per namespace + server in CursorReaper#kill_cursors~~(not going to do this
- ~~Use read_with_one_retry method instead of read_with_retry (in sep PR right now)~~
- ~~Must add disconnect method to CursorReaper~~
- ~~Check if cursor is active at time of sending kill cursors op instead of when it is scheduled~~


estolfo:
@durran 
It'd be awesome if you could let me know if you have any comments on this. Thanks!

durran:
LGTM

PR #767, opened 2016-06-03
RUBY-1112 Check wire version for choosing auth mechanism when authenticating socket

estolfo:


estolfo:
Closing in favor of another pull request: https://github.com/mongodb/mongo-ruby-driver/pull/770

PR #768, opened 2016-06-03
Retry ismaster calls on a server's monitor socket

estolfo:
The SDAM spec says so: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#retry-ismaster-calls-once

### DIFF
--- a/lib/mongo/retryable.rb
+++ b/lib/mongo/retryable.rb
@@ -46,18 +46,19 @@ module Mongo
     # @return [ Result ] The result of the operation.
     #
     # @since 2.1.0
-    def read_with_retry(attempt = 0, &block)
+    def read_with_retry(attempt = 0, max_read_retries = nil, &block)
+      max_retries = max_read_retries || cluster.max_read_retries
       begin
         block.call
       rescue Error::SocketError, Error::SocketTimeoutError
         retry_operation(&block)
       rescue Error::OperationFailure => e
         if cluster.sharded? && e.retryable?
-          if attempt < cluster.max_read_retries
+          if attempt < max_retries
             # We don't scan the cluster in this case as Mongos always returns
             # ready after a ping no matter what the state behind it is.
             sleep(cluster.read_retry_interval)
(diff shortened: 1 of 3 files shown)


lib/mongo/retryable.rb
```diff
@@ -46,18 +46,19 @@ module Retryable
...
     # @since 2.1.0
-    def read_with_retry(attempt = 0, &block)
+    def read_with_retry(attempt = 0, max_read_retries = nil, &block)
+      max_retries = max_read_retries || cluster.max_read_retries
       begin
         block.call
       rescue Error::SocketError, Error::SocketTimeoutError
         retry_operation(&block)
       rescue Error::OperationFailure => e
         if cluster.sharded? && e.retryable?
```
durran: `OperationFailure` here would raise a `NoMethodError` on including classes that don't have a reference to `Cluster`. I think to make this module completely reusable we should remove all references to `cluster` or make another more generic retry module for the monitor and connection.

estolfo: Yeah, that makes sense. I'll update it.

estolfo: After trying to make the Retryable module more generic and avoid references to cluster, I realized it made the API and design pretty messy; in every class that included the Retryable module, or inherited it, I had to forward the methods like #max_read_retries and #read_retry_interval to cluster. I finally realized that it would be much more simple and explicit to make a new method, Retryable#read_with_one_retry and use that when calling ismaster on a Monitor's socket.
The PR is here: https://github.com/mongodb/mongo-ruby-driver/pull/769

estolfo: This would also allow me to use the #read_with_one_retry here instead of jumping through hoops to limit the retry to 1 in the existing #read_with_retry method
https://github.com/mongodb/mongo-ruby-driver/pull/767/files#diff-fcdb9df78585e64c411303e35812189cR171

PR #778, opened 2016-07-04
RUBY-1100 Disconnect a socket if it hasn't been properly authenticated.

estolfo:


estolfo:
closing because we have this instead: https://github.com/mongodb/mongo-ruby-driver/pull/785

PR #783, opened 2016-07-19
Don't freeze caller's hashes

matt-wormley:
Caller's hashes are now dup'ed before freezing for use in the database view to eliminate side effects for the client.

### DIFF
--- a/lib/mongo/collection/view.rb
+++ b/lib/mongo/collection/view.rb
@@ -132,7 +132,7 @@ module Mongo
       def initialize(collection, filter = {}, options = {})
         validate_doc!(filter)
         @collection = collection
-        parse_parameters!(BSON::Document.new(filter.freeze), BSON::Document.new(options.freeze))
+        parse_parameters!(BSON::Document.new(filter.dup.freeze), BSON::Document.new(options.dup.freeze))
       end
 
       # Get a human-readable string representation of +View+.


lib/mongo/collection/view.rb
```diff
@@ -132,7 +132,7 @@ def hash
       def initialize(collection, filter = {}, options = {})
         validate_doc!(filter)
         @collection = collection
-        parse_parameters!(BSON::Document.new(filter.freeze), BSON::Document.new(options.freeze))
+        parse_parameters!(BSON::Document.new(filter.dup.freeze), BSON::Document.new(options.dup.freeze))
```
estolfo: Thanks for pointing this out and for the proposed fix. I would prefer we freeze the @filter and @options variables here instead: https://github.com/matt-wormley/mongo-ruby-driver/blob/89d6016065c8b00ce0979da966872a0411245b64/lib/mongo/collection/view.rb#L162-L163

that way, we won't have to dup anything. What do you think?

matt-wormley: Oh for sure. I didn't look that deep but yes, certainly down lower in the code would be great.

estolfo: I've opened this PR as an alternate solution that avoid duping. Let me know what you think!
https://github.com/mongodb/mongo-ruby-driver/pull/786

PR #695, opened 2015-10-06
RUBY-1049: reauthenticate and retry when receiving a 'not authorized'…

jonhyman:
… or 'unauthorized' response from MongoDB.

PR #784, opened 2016-07-19
RUBY-1049: reauthenticate and retry when receiving unauthorized response.

jimpo:
I am re-opening https://github.com/mongodb/mongo-ruby-driver/pull/695. This is a critical bug that is causing issues for us in production, and is still unresolved as discussed [here](https://jira.mongodb.org/browse/RUBY-1100).

### DIFF
--- a/lib/mongo/error/operation_failure.rb
+++ b/lib/mongo/error/operation_failure.rb
@@ -42,6 +42,14 @@ module Mongo
         'dbclient error communicating with server'
       ].freeze
 
+      # These error messages indicate that the requests are unauthorized.
+      #
+      # @since 2.3.0
+      UNAUTHORIZED_MESSAGES = [
+        'unauthorized',
+        'not authorized'
+      ].freeze
+
       # Can the operation that caused the error be retried?
       #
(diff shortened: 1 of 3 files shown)


girak:
👍

estolfo:
Hi @jimpo 
We won't want to add this retry code into the driver because it's more of a workaround rather than an actual fix. The main problem with this change is that there is no way to distinguish between the server returning a 'not authorized' error because authentication timed out (this may be possible with older server versions, such as 2.4) and a "valid" not authorized error. The former is a temporary failure and the latter is a permanent failure that should not be retried. For example, a read-only user trying to write to a database would get a "valid" not authorized error. You wouldn't want the driver to automatically retry the write operation.
Because only your application would know whether a not authorized error is "valid", it should decide to retry.

That said, I'll open a pull request today with a proposed solution for the not authorized errors and I would welcome you taking a look and testing it out in your staging environment. Thanks!

estolfo:
Closing as we now have https://github.com/mongodb/mongo-ruby-driver/pull/785

PR #785, opened 2016-07-20
RUBY-1100 Disconnect a connection if it didn't finish connecting and authenticating

estolfo:


lib/mongo/server/connection_pool.rb
```diff
@@ -106,7 +106,10 @@ def with_connection
         connection = checkout
         yield(connection)
       ensure
-        checkin(connection) if connection
+        if connection
+          connection.disconnect! unless connection.send(:completed_connecting?)
```
jimpo: Should `completed_connecting?` be public if you are using it here?

jimpo:
Looks reasonable, thanks for getting a patch out.

One thought: on https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/server/connection.rb#L67 it looks like the connection is disconnected if there's an auth failure. Can you just check the connection state when deciding whether to check the connection into the pool or is the new `@authenticated` state necessary?

estolfo:
Hey @jimpo 
I've actually changed the code to be more resilient to threads being killed at any point during connecting. Let me know what you think.
Thanks!

estolfo:
Hi again @jimpo 
In case there's some confusion, this is an issue with the Ruby driver, not Mongoid. In other words, Mongoid 6's release isn't dependent on this and both Mongoid 5 and Mongoid 6 will benefit from a driver release with this fix.

PR #786, opened 2016-07-21
Don't freeze user args to Collection::View#initialize

estolfo:


matt-wormley:
Looks fantastic. I'm sure that fixes my issue. Thanks!

PR #787, opened 2016-07-22
Remove Server::Context

estolfo:
This will allow us to more cleanly mark a server as unknown when auth fails and update the code to comply with what [these tests](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst) verify.


estolfo:
@durran What do you think about removing the Server::Context? It makes the API more direct.

durran:
LGTM

PR #788, opened 2016-07-25
RUBY-1126 Support collation

estolfo:


lib/mongo/bulk_write.rb
```diff
@@ -182,11 +186,18 @@ def insert_one(documents, server, operation_id)
     end
 
     def update(documents, server, operation_id)
```
jyemin: collation should be applied to delete as well as update

lib/mongo/collection/view/builder/find_command.rb
```diff
@@ -94,6 +94,7 @@ def specification
 
           def find_command
             document = BSON::Document.new('find' => collection.name, 'filter' => filter)
+            document[:collation] = collation if collation
```
jyemin: Is validate_collation called at a higher level?

lib/mongo/bulk_write.rb
estolfo: Yes, I'm not quite done with this PR; it's a WIP. 
I've just commit BulkWrite delete support: https://github.com/mongodb/mongo-ruby-driver/pull/788/commits/6c94dbbe8e1b0fd91034f45967ad7aa6b407c97b

lib/mongo/collection/view/builder/find_command.rb
estolfo: yes, it's called here: https://github.com/mongodb/mongo-ruby-driver/pull/788/files/3321ddd3889c17d49a534502e137e44d16404e76#diff-cf306a3295c965d0c497491ba4d63ea6R64
before the FindCommand Builder even has a chance to build the command.

lib/mongo/bulk_write.rb
jyemin: Acknowledged. Thanks for the clarification.

lib/mongo/collection/view/builder/find_command.rb
jyemin: Acknowledged

estolfo:
@rozza
Can you take a look at this? Thank you!

rozza:
LGTM!

Its been an age since I did any ruby so somethings took some deciphering like `feature_enabled?` especially in the crud_spec - very neat.  Also `tap` is new to me and is really useful.

estolfo:
thanks @rozza = )

PR #789, opened 2016-07-27
General improvements

estolfo:
- Using the ServerSelector::PRIMARY constant instead of creating new hashes
- Creating a @primary_selector variable in cluster so that a new server selector object is not created each time a primary needs to be selected.
- Taking advantage of the fact that ServerSelector.get will return the ServerSelector object if it is passed instead of a Hash.
- Adding a check on the local_threshold option to ServerSelector object comparison
- Remove spec file testing non existent class


spec/mongo/operation/specifiable_spec.rb
```diff
@@ -50,4 +50,35 @@
...
+      end
+
+      it 'uses the read option provided' do
+        expect(specifiable.read).to eq(server_selector)
+      end
+    end
+
+    context 'when read is not specified' do
+
+      it 'returns a Primary ServerSelector spec' do
```
estolfo: todo: this behavior is inconsistent with when there is a read specified.

PR #797, opened 2016-07-29
Add docs for ruby driver manual and update .gitignore

kay-kim:
I removed the docs from the .gitignore.  Not sure if you need to update the yard docs configuration on this branch.


estolfo:
Ok, I've updated the branch so you can remove your edit to .gitignore and rebase against 2.0-stable.

PR #803, opened 2016-08-02
[for v2.4.0 / server 3.4] RUBY-1115 Support writeConcern for commands that write

estolfo:


lib/mongo/database.rb
```diff
@@ -148,12 +152,14 @@ def collections
     #
     # @return [ Hash ] The result of the command execution.
     def command(operation, opts = {})
+      if opts[:write]
```
jyemin: What's the intention here?  My understanding is that drivers should not make any special effort to include writeConcern in a generic command.

estolfo: You're right; I've fixed it. Thanks for reviewing and catching that.

PR #804, opened 2016-08-03
Fix warnings in verbose mode

375gnu:
Hi.

We run ruby in verbose mode and we see a lot of warnings, these 4 commits fix the most of them.

### DIFF
--- a/lib/mongo/protocol/message.rb
+++ b/lib/mongo/protocol/message.rb
@@ -77,6 +77,10 @@ module Mongo
       # @return [Fixnum] The request id for this message
       attr_reader :request_id
 
+      def initialize # :nodoc:
+        @request_id = nil
+      end
+
       # The default for messages is not to require a reply after sending a
       # message to the server.
       #
--- a/lib/mongo/protocol/message.rb
+++ b/lib/mongo/protocol/message.rb
@@ -110,7 +110,7 @@ module Mongo
       #
       # @return [ Message ] Instance of a Message class
       def self.deserialize(io, max_message_size = MAX_MESSAGE_SIZE, expected_response_to = nil)
-        length, request_id, response_to, op_code = deserialize_header(BSON::ByteBuffer.new(io.read(16)))
+        length, _request_id, response_to, _op_code = deserialize_header(BSON::ByteBuffer.new(io.read(16)))
 
         # Protection from potential DOS man-in-the-middle attacks. See
         # DRIVERS-276.
--- a/lib/mongo/uri.rb
+++ b/lib/mongo/uri.rb
@@ -339,7 +339,7 @@ module Mongo
     end
 
     def decode(value)
-      ::URI.decode(value)
+      ::URI::DEFAULT_PARSER.unescape(value)
     end
 
     # Hash for storing map of URI option parameters to conversion strategies
--- a/lib/mongo/uri.rb
+++ b/lib/mongo/uri.rb
@@ -193,7 +193,7 @@ module Mongo
     def initialize(string, options = {})
       @string = string
       @options = options
-      empty, scheme, remaining = @string.partition(SCHEME)
+      _, scheme, remaining = @string.partition(SCHEME)
       raise_invalid_error!(INVALID_SCHEME) unless scheme == SCHEME
       setup!(remaining)
     end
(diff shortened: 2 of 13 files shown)


lib/mongo/protocol/message.rb
```diff
@@ -226,7 +226,7 @@ def serialize_fields(buffer, max_bson_size = nil)
       # @param buffer [String] Buffer to receive the header
       # @return [String] Serialized header
       def serialize_header(buffer)
-        set_request_id unless @request_id
+        set_request_id unless defined?(@request_id)
```
estolfo: Instead of checking if ``@request_id`` is ``#defined?`` can we add something to the ``#set_request_id`` method instead?
Something like the following:

```ruby
      def set_request_id
        @request_id || @@id_lock.synchronize do
          @request_id = @@request_id += 1
        end
      end
```

lib/mongo/uri.rb
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-require 'uri'
+require 'addressable'
```
estolfo: Can you explain this please? Why should we create a dependency on another gem instead of using a core Ruby module?

mongo.gemspec
```diff
@@ -31,4 +31,5 @@ Gem::Specification.new do |s|
   s.bindir            = 'bin'
 
   s.add_dependency 'bson', '~> 4.1'
+  s.add_dependency 'addressable'
```
estolfo: I'd rather avoid adding a dependency on another gem and stick with the core Ruby URI module instead.

lib/mongo/protocol/message.rb
375gnu: @estolfo, the change you've suggested will not fix a warning "warning: instance variable @request_id not initialized".

Better fix will be to set @request_id in initialize method.

estolfo: ok. How about we set the ``@request_id`` to nil in the initialize method and use the memoize code I suggested for #set_request_id?

lib/mongo/uri.rb
estolfo: hm, I really don't want a dependency on another gem for this. Are there any other solutions?

375gnu: copy-paste code from addressable gem? just 3 lines of code.

lib/mongo/protocol/message.rb
375gnu: "Yes" for the 1st part and "you decide" for the 2nd one. I don't know when all cases set_request_id is called so I leave it to you.

lib/mongo/uri.rb
estolfo: I found that Rack is using 

```ruby
::URI::DEFAULT_PARSER.unescape
```
in its unescape_path method:
http://www.rubydoc.info/github/rack/rack/Rack%2FUtils.unescape_path

so I'll do the same.

lib/mongo/protocol/message.rb
estolfo: I think the best thing to do here is to set the ``@request_id`` to nil in the initializer. and then keep the line as ``set_request_id unless @request_id``

lib/mongo/uri.rb
estolfo: It seems like much more than 3 lines of code.

I found that Rack is using

::URI::DEFAULT_PARSER.unescape
in its unescape_path method:
http://www.rubydoc.info/github/rack/rack/Rack%2FUtils.unescape_path

so we should do the same.

375gnu:
On 10/11/16, Emily S <notifications@github.com> wrote:
> estolfo requested changes on this pull request.
>
> I just have two comments about the request_id and the URI.unescape. Would
> you mind updating the code and then I can merge these changes?


Okay, I'll push updated PR

>
>> @@ -226,7 +226,7 @@ def serialize_fields(buffer, max_bson_size = nil)
>        # @param buffer [String] Buffer to receive the header
>        # @return [String] Serialized header
>        def serialize_header(buffer)
> -        set_request_id unless @request_id
> +        set_request_id unless defined?(@request_id)
>
> I think the best thing to do here is to set the ``@request_id`` to nil in
> the initializer. and then keep the line as ``set_request_id unless
> @request_id``
>
>> @@ -12,7 +12,7 @@
>  # See the License for the specific language governing permissions and
>  # limitations under the License.
>
> -require 'uri'
> +require 'addressable'
>
> It seems like much more than 3 lines of code.
>
> I found that Rack is using
>
> ::URI::DEFAULT_PARSER.unescape
> in its unescape_path method:
> http://www.rubydoc.info/github/rack/rack/Rack%2FUtils.unescape_path
>
> so we should do the same.
>

PR #809, opened 2016-08-10
Fix socket timeout on read operations

wandenberg:
Change the implementation of `Mongo::Socket.read_from_socket` method to check if the socket_timeout was exceeded while waiting for data.
This ensure that the client don't get stuck on a long query or a mongo server overload or a network problem.

### DIFF
--- a/lib/mongo/socket.rb
+++ b/lib/mongo/socket.rb
@@ -169,7 +169,22 @@ module Mongo
     private
 
     def read_from_socket(length)
-      @socket.read(length) || String.new
+      # Block on data to read for timeout seconds
+      # using the suggested implementation of http://www.ruby-doc.org/core-2.1.3/Kernel.html#method-i-select
+      # to work with SSL connections and some OSs like Ubuntu, CentOS and Red Hat
+      time_left = timeout.to_f
+      data = String.new
+      begin
+        while data.length < length
+          data << @socket.read_nonblock(length - data.length)
+        end
+      rescue IO::WaitReadable
+        raise Timeout::Error.new("Took more than #{timeout} seconds to receive data.") if (time_left -= 0.1) <= 0
+        Kernel::select([@socket], nil, [@socket], 0.1)
+        retry
+      end
+
+      data
     end
 
     def unix_socket?(sock)
--- a/spec/mongo/server/connection_spec.rb
+++ b/spec/mongo/server/connection_spec.rb
@@ -400,6 +400,31 @@ describe Mongo::Server::Connection do
       end
     end
 
+    context 'when a socket timeout occurs' do
+      let(:client) do
+        authorized_client.with(socket_timeout: 1.5)
+      end
+
+      before do
+        authorized_collection.insert_one( a: 1)
+      end
+
+      after do
+        authorized_collection.delete_many
+      end
+
+      it 'raises a timeout when it expires' do
+        start = Time.now
+        expect {
+          Timeout::timeout(3) do
+            client[authorized_collection.name].find("$where" => "sleep(10000) || true").first
+          end
+        }.to raise_exception(Timeout::Error, "Took more than 1.5 seconds to receive data.")
+        end_time = Time.now
+        expect(end_time - start).to be_within(0.2).of(1.5)
+      end
+    end
+
     context 'when the process is forked' do
 
       let(:insert) do


durran:
The case in which the requested number of bytes is not read doesn't seem to be handled here. (For example, we request 64 bytes but only 32 are returned and we need to continue until all bytes are read for the complete message).

`IO#read_nonblock` calls the system's `read` method which does not guarantee that all bytes will actually be returned.

wandenberg:
@durran you are absolutely right.
I was focused on the `Mongo::Socket.read` function which calls `Mongo::Socket.read_from_socket` and already execute a loop to ensure that all data was read.
Can you check if the implementation is fine now?

durran:
Looks good now - we'll have to give it a good few runs on our internal infrastructure first with SSL as well just to bang on it a bit.

lib/mongo/socket.rb
```diff
@@ -169,7 +169,22 @@ def eof?
...
+      # using the suggested implementation of http://www.ruby-doc.org/core-2.1.3/Kernel.html#method-i-select
+      # to work with SSL connections and some OSs like Ubuntu, CentOS and Red Hat
+      time_left = timeout.to_f
+      data = ""
+      begin
+        while data.length < length
+          data << @socket.read_nonblock(length - data.length)
+        end
+      rescue IO::WaitReadable
+        raise Error::OperationFailure.new("Took more than #{timeout} seconds to receive data.") if (time_left -= 0.1) <= 0
```
durran: Actually I think here we would want to raise a `TimeoutError` instead of an `OperationFailure`, to be consistent with the previous behaviour (which raised a `TimeoutError` but wasn't doing it within the actual timeout set on the initial socket, at least on OSX for me locally)

wandenberg: I thought to use the SocketTimeoutError, but it is captured and the operation is retryed again. 
Use the OperationFailure was a try to not create a new exception and avoid retrying an operation that is taking too long to return multiple times.

durran: We use the `OperationFailure` error specifically for errors that come from the server, for example write concern errors or query syntax failures to name a few - `TimeoutError` is outside of this classification but won't require a new error to be created.

wandenberg:
@durran I replaced the exception with a `Timeout::Error`. Did you have opportunity to test in your internal infrastructure?

spec/mongo/server/connection_spec.rb
```diff
@@ -400,6 +400,35 @@
...
+          expect {
+            Timeout::timeout(3) do
+              client[:users].find("age" => { "$gte" => 65 }).first
+            end
+          }.to raise_exception("Took more than 1.5 seconds to receive data.")
+        end
+        expect(time).to be_within(0.2).of(1.5)
+      end
+    end
+
```
estolfo: How about doing this instead to test?

```ruby
      let(:client) do
        authorized_client.with(socket_timeout: 1.5)
      end

      it 'raises a timeout when it expires' do
        start = Time.now
        begin
          Timeout::timeout(3) do
            client[authorized_collection.name].find("$where" => "sleep(10000) || true").first
          end
        rescue
        end
        end_time = Time.now
        expect(end_time - start).to be_within(0.2).of(1.5)
      end
```

spec/spec_helper.rb
```diff
@@ -15,6 +15,10 @@
 CA_PEM = "#{SSL_CERTS_DIR}/ca.pem"
 CRL_PEM = "#{SSL_CERTS_DIR}/crl.pem"
 
+require 'timeout'
+require 'benchmark'
+require 'fileutils'
+require 'tmpdir'
 require 'mongo'
```
estolfo: We wouldn't need all these require's or the following helper methods. We can just test with the "$where" clause I put in the suggested test above.

lib/mongo/socket.rb
```diff
@@ -169,7 +169,22 @@ def eof?
...
+        while data.length < length
+          data << @socket.read_nonblock(length - data.length)
+        end
+      rescue IO::WaitReadable
+        raise Timeout::Error.new("Took more than #{timeout} seconds to receive data.") if (time_left -= 0.1) <= 0
+        Kernel::select([@socket], nil, [@socket], 0.1)
+        retry
+      end
+
+      data || String.new
```
estolfo: This looks good but I'm going to test on Jenkins so that each version of Ruby, with and without SSL, and various versions of MongoDB are tested.

spec/mongo/server/connection_spec.rb
jarthod: +1, why the `rescue` though ? I guess getting a Timeout error is as good a getting `expected 3.0 to be within 0.2 of 1.5` isn't it ?

spec/spec_helper.rb
jarthod: +1

.travis.yml
```diff
@@ -22,6 +22,7 @@ env:
   global:
     - CI="travis"
     - JRUBY_OPTS="--server -J-Xms512m -J-Xmx1024m"
+    - PATH="${PWD}/mongodb-linux-x86_64-${MONGODB}/bin:$PATH"
```
jarthod: I guess this is not required any more once you remove the helpers manually launching mongod

lib/mongo/socket.rb
```diff
@@ -169,7 +169,22 @@ def eof?
     private
 
     def read_from_socket(length)
-      @socket.read(length) || String.new
+      # Block on data to read for timeout seconds
+      # using the suggested implementation of http://www.ruby-doc.org/core-2.1.3/Kernel.html#method-i-select
+      # to work with SSL connections and some OSs like Ubuntu, CentOS and Red Hat
+      time_left = timeout.to_f
+      data = ""
```
jarthod: I think you should use `String.new` here to emphasize the fact that it has to be mutable and be ready for the Ruby 3.0 immutable string literals.

lib/mongo/socket.rb
jarthod: No need for the `|| String.new` here as data can't be nil any more.

lib/mongo/socket.rb
estolfo: +1

lib/mongo/socket.rb
estolfo: +1

spec/mongo/server/connection_spec.rb
estolfo: I usually prefer not to check String literals in tests and I didn't like the idea of the test:

```ruby
        expect {
          Timeout::timeout(3) do
            client[authorized_collection.name].find("$where" => "sleep(10000) || true").first
          end
        }.to raise_exception(Timeout::Error)
```

to be true regardless of whether the socket timeout of 1.5 was actually applied. I know we would also have a test on the length of time anyway that would fail but I wanted to only test the length of time, not the error.

I don't feel that strongly on this point so we can change it to the following if you both want:

```ruby
      it 'raises a timeout when it expires' do
        start = Time.now
        expect {
          Timeout::timeout(3) do
            client[authorized_collection.name].find("$where" => "sleep(10000) || true").first
          end
        }.to raise_exception("Took more than 1.5 seconds to receive data.")
        end_time = Time.now
        expect(end_time - start).to be_within(0.2).of(1.5)
      end
    end
```

jarthod: But you don't expect any exception here as we're not supposed to hit the 3 sec timeout, this is just to detect regression without blocking the specs, tu me that would be:
```ruby
      it 'raises a timeout when it expires' do
        start = Time.now
        Timeout::timeout(3) do
          client[authorized_collection.name].find("$where" => "sleep(10000) || true").first
        end
        expect(Time.now - start).to be_within(0.2).of(1.5)
      end
```

jarthod: Ah no you're actually checking the newly raised exception here, I though this was for the `Timeout::timeout(3)` exception, sorry. In this case I would prefer a fully qualified check to avoid any regression on the raised exception:

```ruby
     it 'raises a timeout when it expires' do
        start = Time.now
        expect {
          Timeout::timeout(3) do
            client[authorized_collection.name].find("$where" => "sleep(10000) || true").first
          end
        }.to raise_exception(Timeout::Error, "Took more than 1.5 seconds to receive data.")
        end_time = Time.now
        expect(end_time - start).to be_within(0.2).of(1.5)
      end
    end
```

But I won't push for it either, so I let @wandenberg and you decide ;)

estolfo: This is up to you, @wandenberg 
The two most important things are 1) the spec doesn't block the test suite and 2) the test only succeeds if time passed is <= 1.5

estolfo:
btw, this change is looking good: https://jenkins.10gen.com/view/Ruby/job/mongo-ruby-driver-emilys-fork/64/

estolfo:
lgtm

Note that the test failures are because the ``delete_many`` operation in the ``after`` block cannot finish in under 5 seconds because it's blocked waiting for the $where query to finish. I'm going to merge this then update the test to reduce the $where query sleep so that the ``after`` block can complete successfully.

wandenberg:
You are welcome @estolfo .
I will do another PR with a reduced timeout.

jarthod:
\o/

estolfo:
hey @jarthod Can you test if this ticket is resolved as a result of these changes? https://jira.mongodb.org/browse/RUBY-1120

PR #811, opened 2016-08-12
DOCS-7775 Update compatibility matrix for 2.3

estolfo:


pedrosnk:
Is this driver fully compatible with ruby 2.3.x (MRI) ?

estolfo:
Yes, that's what our compatibility matrix shows: https://docs.mongodb.com/ruby-driver/master/reference/driver-compatibility/

Are you finding otherwise?

pedrosnk:
Thanks a lot for the answer.

I only found this other matrix https://docs.mongodb.com/ruby-driver/master/installation/ only I was a little concerned.

estolfo:
oh! Sorry about the confusion. We'll update that as well.

PR #814, opened 2016-08-31
Description comparision should respect empty config

dongintercom:
When both descriptions' config are empty, the comparison (`==`) for descriptions returns `false` while it should return `true`.

TODO:

- [ ] add tests

@estolfo please take a look, thanks!

### DIFF
--- a/lib/mongo/server/description.rb
+++ b/lib/mongo/server/description.rb
@@ -577,6 +577,7 @@ module Mongo
       private
 
       def compare_config(other)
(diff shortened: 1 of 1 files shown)


estolfo:
@dongintercom Can you describe to me the scenario in which this came up and why this change is needed? Thanks

estolfo:
I'm going to close this pull request. Please reopen if you can provide a reason for why this change is needed.

PR #815, opened 2016-09-07
RUBY-1140 Add TLS SNI support

estolfo:


jyemin:
Seems reasonable, but having run into issues in Java you should do some manual testing for the following cases:

1. An IPv4 literal
2. An IPv6 literal

for both MRI and JRuby.

durran:
LGTM

estolfo:
This doesn't seem to work for JRuby. 

I'm setting the hostname as indicated [here](https://github.com/jruby/jruby-openssl/blob/master/src/main/java/org/jruby/ext/openssl/SSLSocket.java#L180-L181)

but the server logs show:
``
new ssl connection, SNI server name []
``

for MRI, it shows:
``
new ssl connection, SNI server name [server]
``

behackett:
If JRuby doesn't support it (it's undocumented in MRI, so no surprise) then oh well. Open a ticket against MRI to get it documented, and against JRuby to get it implemented. I couldn't implement it for all the versions of python we support either.

estolfo:
So in the Python driver, you only set the hostname for certain versions of Python?

behackett:
Yep, only versions of python that support SNI.

https://docs.python.org/2/library/ssl.html#ssl.HAS_SNI

jyemin:
There's a claim that it's implemented here: https://github.com/jruby/jruby/pull/349.  But the comment saying that SNI is supported starting with Java 7 doesn't jibe with my experience. The Java driver's SNI support relies on a Java 8 feature: https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLParameters.html#setServerNames-java.util.List-

estolfo:
I opened a ticket with the openssl library for JRuby. Hopefully I'll get some more info soon from the JRuby team directly...
https://github.com/jruby/jruby-openssl/issues/103

estolfo:
fyi, the Ruby team updated their docs: https://bugs.ruby-lang.org/issues/12740

estolfo:
Cannot support this for JRuby until they resolve the bug in openssl:
https://twitter.com/EmStolfo/status/775615814737231872

PR #817, opened 2016-09-19
RUBY-1142 Allow certs to be passed as string in the configuration

nessche:
Basically added two functions which check the option content, if it is an OpenSSL object already return the object, if it is a strong containing the PEM-encoded key/certificate, decode it into an OpenSSL object and return it, if it is a file path, read the data from the file at the given path, decode it into an OpenSSL object and return it.
Split the test key/cert combo files into separate files to make it possible for JRuby to parse them correctly without knowing in advance the type of Key to be expected.
Added an encrypted version of the client_key because I did not know the passphrase for password_protected.pem.


estolfo:
Thanks for the pull request, @nessche !
I'll take a look and get back to you.

estolfo:
Hi @nessche 

Thanks again for your pull request!
I've discussed this change internally and I'd like to hold off accepting it for a few reasons:

1. The current behavior of the driver, as you know, is to accept a file path for a cert. With your change, this code path will always pass through a begin - rescue, thus affecting the performance of previous behavior; a benchmark would show a performance regression using a file path between the previous driver and the driver with this change.

2. The drivers team is going to write a spec for Client options. I'd like to hold off on expanding any functionality of our Client options until that spec takes shape. I don't have a sense for what the timeline is, but I can update you as soon as I know.

3. This change would create a disparity between what options are possible using a URI to create a Client and those that are possible when instantiating the Client directly. I'm open to there being a difference but I'd like to limit the difference to only the extremely necessary.

I hope you don't mind that I leave this PR open for a bit without merging it. When I have more information on the above points, I will update you.

Thanks!
Emily

nessche:
Hi!

No problem at all holding off the PR. Just a few ideas to solve some of the problems highlighted above: 

1) what about defining new option names for the "pem-encoded" values and reserving the existing options for file paths? This way we would know exactly what to expect in the option, the begin/rescue cycle would be avoided and the performance should be exactly the same as the previous versions. I guess this clashes with point 2, but once the spec for the Client options is in shape it could be a viable solution.

3) I am a big fan of URI options, because they allow to change the configuration on the fly, by changing the URI, without redeploying a new version of the applications server (with e.g. a new mongoid configuration). It is for the same reason that I feel that a more dynamic mechanism for passing certs and keys to container-based deployments should be there (e.g. for using a self-signed certificate/key in staging and a proper one in production, etc...)

I am available to create a new PR according to whatever mechanism/solution we can come up with.

thanks,
marco

estolfo:
Hi Marco

Thanks for your reply!

Your proposed solution to #1 is indeed possible. Let me just give my team a little bit of time to start working on the client options specification so I can figure out how much flexibility we have with allowing additional options when instantiating the client.

On point #3: Are you aware of the client.with(...) method? That would allow you to take an existing client and create a new one with altered and/or additional options. Let me know if that's what you're looking for. I do understand what you're saying about the importance of URI options though.

Emily

nessche:
Hi Emily,

let me explain the use case that sprung my request. I have a containerized Rails app that I would like to configure only with environment variables. This way I can have exact 1:1 match between staging and production (bar the env vars, obviously). If I could pass in PEM encoded keys/certs to the ruby driver I could just have in my mongoid.yml something like
`ssl_key_pem: <%= ENV['SSL_CA_CERT'] %>`
and then set the env variable to contain whatever PEM value I want without the need of having the PEM key to be packed into the container. This way I could even have my docker image being stored in a public docker repository without any sensitive file being at risk.

I hope this clears up a bit the reason why I'd like the chance to pass in a PEM encoded string to some of the options.

marco

estolfo:
Hi @nessche 
Ok, I understand your use case. I'll have an update soon. Thanks!
Emily

estolfo:
Hi @nessche 
I've spoken to some of my colleagues and I think the best thing would be to specify different option names for each of these. Do you want to change the pull request? I'll make some suggestions inline.

lib/mongo/socket/ssl.rb
```diff
@@ -122,23 +124,18 @@ def create_context(options)
       end
 
       def set_cert(context, options)
-        context.cert = OpenSSL::X509::Certificate.new(File.open(options[:ssl_cert]))
+        context.cert = certificate_from_option(options[:ssl_cert])
```
estolfo: maybe we can call this ``ssl_cert_string`` or ``ssl_cert_raw`` for the pem-encoded key/cert
leave ``ssl_cert`` as meaning the file path for backwards compatibility
have ``ssl_cert_obj`` mean the OpenSSL::X509::Certificate object

and the same variations for ``ssl_ca_cert``

lib/mongo/socket/ssl.rb
```diff
@@ -122,23 +124,18 @@ def create_context(options)
...
       end
 
       def set_key(context, options)
-        if options[:ssl_key_pass_phrase]
-          context.key = OpenSSL::PKey::RSA.new(File.open(options[:ssl_key]),
-                                               options[:ssl_key_pass_phrase])
-        else
-          context.key = OpenSSL::PKey::RSA.new(File.open(options[:ssl_key]))
-        end
+        context.key = key_from_option(options[:ssl_key], options[:ssl_key_pass_phrase])
```
estolfo: the options for this would be:
``ssl_key`` : file path
``ssl_key_string`` or ``ssl_key_raw`` : string
``ssl_key_obj`` or ``ssl_key_object`` for the OpenSSL::PKey::RSA

``ssl_key_pass_phrase`` stays as-is

nessche:
Sounds good. I will look into this tomorrow and then post the changes to the same PR

estolfo:
great, ping me when you're ready for a code review and I'll take a look. Thanks!

nessche:
One quick question: how do you want to handle when more than one "alternative" option is specified. In other words, what if I specify both `ssl_key` and `ssl_key_string`? There are two possible ways of handling:
* raising an error
* choosing the "higher priority" option (and in this case what is the priority order among file, PEM string, and Object?)

m

nessche:
Don't look at the code yet, I still need to write couple of specs to make sure everything works fine. In any case regarding the problem above (priority vs. error), I decided to go priority all the way, and thus defined that file > string > object, because that is most likely also the order of popularity. But it can be easily changed if we find a compelling reason why it should be made in a different way.

lib/mongo/socket/ssl.rb
```diff
@@ -116,30 +118,43 @@ def connectable?
 
       def create_context(options)
         context = OpenSSL::SSL::SSLContext.new
-        set_cert(context, options) if options[:ssl_cert]
```
nessche: Removed the if branches because anyway this will be checked in the set_xxx methods

lib/mongo/socket/ssl.rb
```diff
@@ -116,30 +118,43 @@ def connectable?
...
-                                               options[:ssl_key_pass_phrase])
-        else
-          context.key = OpenSSL::PKey::RSA.new(File.open(options[:ssl_key]))
+        passphrase = options[:ssl_key_pass_phrase]
+        if options[:ssl_key]
+          context.key = passphrase ? OpenSSL::PKey.read(File.open(options[:ssl_key]), passphrase) : OpenSSL::PKey.read(File.open(options[:ssl_key]))
+        elsif options[:ssl_key_string]
+          context.key = passphrase ? OpenSSL::PKey.read(options[:ssl_key_string], passphrase) : OpenSSL::PKey.read(options[:ssl_key_string])
+        elsif options[:ssl_key_object]
+          context.key = options[:ssl_key_object]
```
nessche: We delegate the checking of the correct type of object to the underlying SSL socket. If we want we can introduce type checks also here

lib/mongo/client.rb
```diff
@@ -55,8 +55,14 @@ class Client
...
       :ssl_ca_cert,
+      :ssl_ca_cert_string,
+      :ssl_ca_cert_object,
       :ssl_cert,
+      :ssl_cert_string,
+      :ssl_cert_object,
       :ssl_key,
+      :ssk_key_string,
+      :ssl_key_object,
       :ssl_key_pass_phrase,
```
nessche: Added new options as agreed, xxx_string => PEM-encoded, xxx_object => Ruby Object

lib/mongo/socket/ssl.rb
```diff
@@ -116,30 +118,43 @@ def connectable?
...
       def set_cert(context, options)
-        context.cert = OpenSSL::X509::Certificate.new(File.open(options[:ssl_cert]))
+        if options[:ssl_cert]
+          context.cert = OpenSSL::X509::Certificate.new(File.open(options[:ssl_cert]))
+        elsif options[:ssl_cert_string]
+          context.cert = OpenSSL::X509::Certificate.new(options[:ssl_cert_string])
+        elsif options[:ssl_cert_object]
+          context.cert = options[:ssl_cert_object]
+        end
       end
```
nessche: The priority is as agreed

File > PEM-Encoded > Object

nessche:
Hi @estolfo,

the pull request is in a shape that could be reviewed. The code is dead simple and I added some documentation for the new options. Feel free to change it to make it more readable/understandable.

There is only one open point from my side and whether we should check the Ruby Objects passed in the options or we should delegate such check to the underlying SSL Socket (which will throw a TypeError if the object is not of the expected type).

marco

lib/mongo/socket/ssl.rb
```diff
@@ -23,6 +23,8 @@ class Socket
     class SSL < Socket
       include OpenSSL
 
+      ALLOWED_KEY_TYPES = [OpenSSL::PKey::RSA]
```
nessche: This was a relic from the first implementation which I had left there in case we needed type checking. But since we are not doing type checking on the objects passed in this constant is  not used and therefore I removed it completely

estolfo:
Thanks for your updates. I'm running some tests and will get back to you soon! Here is your repo master running in jenkins (igonore that the project is called emily's fork)
https://jenkins.10gen.com/view/Ruby/job/mongo-ruby-driver-emilys-fork/14/

estolfo:
Also, if you have the time to update the client options in our tutorial docs, that'd be great. They are found here: https://github.com/mongodb/mongo-ruby-driver/blob/master/docs/tutorials/ruby-driver-create-client.txt#L149

And end up here: https://docs.mongodb.com/ruby-driver/master/

nessche:
I'll take a look at the tutorial docs in couple of days (day job getting in the way of my open source contributions.... :) )

estolfo:
ok, thanks. Otherwise, I can add it in myself. I just thought it might be nice to have all work associated with this consolidated in one pull request. Let me know if you'd like me to just put it in.
Thanks!

PR #821, opened 2016-09-27
reduce the sleep on timeout test to not block operations on after block

wandenberg:
Like @estolfo said on #809 the sleep is too high blocking `after` block operations


estolfo:
sorry, I already did the update! https://github.com/mongodb/mongo-ruby-driver/commit/41a1ba7a0d5a82cfad69d2ff3e92c14498498a5d
Thanks anyway though = )

wandenberg:
perfect ;)

PR #823, opened 2016-09-29
RUBY-1120 Don't set a default timeout on a socket and fix logic

estolfo:
The time_left variable was only ever decremented if a WaitReadable error was raised.
This changes the logic so that timeout is taken into account regardless of whether bytes are successfully read or if there is a timeout when selecting.

Remove the default socket timeout, as it doesn't work in Ruby and was effectively infinite before.

### DIFF
--- a/lib/mongo/socket.rb
+++ b/lib/mongo/socket.rb
@@ -169,21 +169,25 @@ module Mongo
     private
 
     def read_from_socket(length)
-      # Block on data to read for timeout seconds
-      # using the suggested implementation of http://www.ruby-doc.org/core-2.1.3/Kernel.html#method-i-select
-      # to work with SSL connections and some OSs like Ubuntu, CentOS and Red Hat
       data = String.new
       deadline = (Time.now + timeout) if timeout
       begin
         while (data.length < length)
-          if deadline
-            raise Timeout::Error.new("Took more than #{timeout} seconds to receive data.") if (deadline - Time.now) <= 0
-          end
           data << @socket.read_nonblock(length - data.length)
         end
       rescue IO::WaitReadable
-        Kernel::select([@socket], nil, [@socket], 0.1)
-        retry
+        if deadline
+          now = Time.now
+          if deadline - now <= 0
+            raise Timeout::Error.new("Took more than #{timeout} seconds to receive data.")
+          else
+            Kernel::select([@socket], nil, [@socket], (deadline - now))
+            retry
+          end
+        else
+          Kernel::select([@socket], nil, [@socket])
+          retry
+        end
       end
 
       data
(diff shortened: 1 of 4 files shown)


lib/mongo/socket.rb
```diff
@@ -172,14 +172,16 @@ def read_from_socket(length)
       # Block on data to read for timeout seconds
       # using the suggested implementation of http://www.ruby-doc.org/core-2.1.3/Kernel.html#method-i-select
       # to work with SSL connections and some OSs like Ubuntu, CentOS and Red Hat
-      time_left = timeout.to_f
       data = String.new
+      deadline = (Time.now + timeout) if timeout
       begin
-        while data.length < length
+        while (data.length < length)
+          if deadline
```
jyemin: Seems more logical to move this block into the rescue since you should always be reading at least once.

lib/mongo/socket.rb
```diff
@@ -172,14 +172,16 @@ def read_from_socket(length)
...
+            raise Timeout::Error.new("Took more than #{timeout} seconds to receive data.") if (deadline - Time.now) <= 0
+          end
           data << @socket.read_nonblock(length - data.length)
         end
       rescue IO::WaitReadable
-        raise Timeout::Error.new("Took more than #{timeout} seconds to receive data.") if (time_left -= 0.1) <= 0
         Kernel::select([@socket], nil, [@socket], 0.1)
```
jyemin: Instead of timing out every tenth of a second, time out at Time.now - deadline (or set no timeout if !deadline).  But cache the call to Time.now from the raise statement so that the timeout of the select is guaranteed to be positive

lib/mongo/socket.rb
```diff
@@ -194,7 +196,7 @@ def unix_socket?(sock)
     def set_socket_options(sock)
       sock.set_encoding(BSON::BINARY)
 
-      unless unix_socket?(sock) && BSON::Environment.jruby?
+      if timeout && !(unix_socket?(sock) && BSON::Environment.jruby?)
         encoded_timeout = [ timeout, 0 ].pack(TIMEOUT_PACK)
         sock.setsockopt(SOL_SOCKET, SO_RCVTIMEO, encoded_timeout)
```
jyemin: What's going on in this if statement?  Are receive timeouts sometimes honored?  Does it make any sense if all reads are non-blocking?

jyemin:
As non-blocking IO is generally slower than blocking IO, it would be good to do some benchmarking to see what effects the change to non-blocking has had.

lib/mongo/socket.rb
estolfo: Since we're now using #read_nonblock, I removed this. I don't see that it would have any effect.

lib/mongo/socket.rb
estolfo: Ok, I've updated this code block with all your suggestions.

lib/mongo/socket.rb
estolfo: Done.

jyemin:
The code looks good to me now.  Let's see what the benchmarks look like though.

estolfo:
I've done some benchmarking and the performance impact is negligible. For a small performance difference, the gain is that socket timeouts are actually applied.

on master:
```ruby
SINGLE DOC BENCHMARK:: FIND ONE BY ID
find_one : 5.2999278089992

SINGLE DOC BENCHMARK:: COMMAND
command : 0.956540365004912
```
with this change:
```ruby
SINGLE DOC BENCHMARK:: FIND ONE BY ID
find_one : 5.3565079609979875

SINGLE DOC BENCHMARK:: COMMAND
command : 0.9861492830023053
```

jyemin:
LGTM

PR #826, opened 2016-10-13
DOCS-8717: document Decimal128 in Ruby driver

steveren:


estolfo:
I just realized that this PR was made against the Ruby driver and it should have been made against the bson gem, as that's where Decimal128 support is. Using Decimal128 doesn't require the driver.

@steveren Do you think it would be possible to put this into the bson gem? If you want, I can do the transfer myself. It should go here: https://github.com/mongodb/bson-ruby/tree/master/docs
Let me know what you prefer.

PR #832, opened 2016-10-28
RUBY-1151 Allow command document size to have a limit of maxBsonObjectSize + 16K

estolfo:


estolfo:
closing this because OP_COMMAND will be implemented soon and this will be done as part of it.

PR #833, opened 2016-11-02
Fix round trip time units mismatch

jarthod:
Hi, it's me again with another bug in the ruby driver (I think) ☺

I tried using the `nearest` for my geo-redundant RS (2 servers with 1ms latency, 1 with 180ms) and noticed that it simply didn't work, it selects randomly any of my servers, including the 180ms one.

I dug a bit in the code and found out that there seems to be some mismatch in the units, some variables are supposed to contains ms but contains sec, so when the threshold (0.015s) was added to the slowest latency (ex: 0.001s) it is first multiplied by 1000, leading to a 15.001 sec threshold ^^

I fixed this by removing the multiplication and fixing all the specs stubbing ms whereas the real code works with seconds:
```ruby
production> Mongoid.default_client.cluster.servers.map &:average_round_trip_time
=> [0.00290229132, 0.18428593504000002, 0.0012378256400000004]
```

### DIFF
--- a/lib/mongo/server/description.rb
+++ b/lib/mongo/server/description.rb
@@ -239,7 +239,7 @@ module Mongo
       #
       # @param [ Address ] address The server address.
       # @param [ Hash ] config The result of the ismaster command.
-      # @param [ Float ] average_round_trip_time The moving average time (ms) the ismaster
+      # @param [ Float ] average_round_trip_time The moving average time (sec) the ismaster
       #   call took to complete.
       #
       # @since 2.0.0
--- a/lib/mongo/server/description/inspector.rb
+++ b/lib/mongo/server/description/inspector.rb
@@ -61,7 +61,7 @@ module Mongo
         #
         # @param [ Description ] description The old description.
         # @param [ Hash ] ismaster The updated ismaster.
-        # @param [ Float ] average_round_trip_time The moving average round trip time (ms).
+        # @param [ Float ] average_round_trip_time The moving average round trip time (sec).
         #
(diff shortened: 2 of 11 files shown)


estolfo:
Hi @jarthod 

Thanks so much for reporting this and for the fix! It's always nice getting pull requests from you because I know the driver will be better as a result = )

You're right that the unit of the local_threshold variable wasn't properly respected. I believe this happened because the URI specifies localThresholdMS as milliseconds and the client options specify it as seconds. In any case, your changes look good but we should probably also change the local_threshold specified [here](https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/mongo/client_spec.rb#L157-L176) to be 0.01 to emphasize that the units is seconds. If you want to make that change as well, that'd be great.
Thanks!
Emily

jarthod:
Yep, makes sense, I just changed that too ☺

estolfo:
thank you!

jarthod:
:beers:

estolfo:
I've also merged it into [2.3-stable](https://github.com/mongodb/mongo-ruby-driver/tree/2.3-stable) so that the next release of the 2.3 series will have this fix as well. Thanks again!

jarthod:
Great, thanks :smiley:

estolfo:
I've also merged it into [2.3-stable](https://github.com/mongodb/mongo-ruby-driver/tree/2.3-stable) so that the next release of the 2.3 series will have this fix as well. Thanks again!

PR #834, opened 2016-11-03
RUBY-1096 Implement SDAM Monitoring spec

estolfo:
Ref: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst


durran:
LGTM :shipit:

estolfo:
Thank you for reviewing @durran !

PR #837, opened 2016-11-28
RUBY-1136 Allow ODM name and version to be included in app_metadata in a platform client option

estolfo:


estolfo:
@bjori, I've included an embedded doc at the key 'odm' as part of the driver metadata sent during the connection handshake. I've reread the spec and I'm not sure if it's permitted to add an extra embedded doc in the driver embedded doc?
The handshake doc would look like this if the odm name and version are specified:

```
{ driver: { name: "mongo-ruby-driver", version: "2.4.0.rc1", odm: { name: "mongoid", version: "6.0.2" } }, os: { type: "darwin", name: "darwin14.5.0", architecture: "x86_64" }, platform: "2.2.3, x86_64-darwin14, x86_64-apple-darwin14.5.0" }
```

bjori:
This should go into the `platform` field. It was deliberately not part of the `client` document.
The spec doesn't have the words `you MUST NOT add other fields`, but it does indicated the `client` document to be complete and fully documented, which was supposed to be sufficient to indicated other fields are not allowed

estolfo:
thanks for your response, @bjori 
I wrote you a direct message about concatenating the ODM name and version to the driver name and version Strings because @rozza mentioned that's how Spark ended up doing it:

```
{ "driver" : { "name" : "mongo-java-driver|mongo-spark", "version" : "unknown|2.0.0" }, "os" : { "type" : "Darwin", "name" : "Mac OS X", "architecture" : "x86_64", "version" : "10.11.6" }, "platform" : "Java/Oracle Corporation/1.8.0_31-b13|Scala 2.10 / Spark 2.0.0", "application" : { "name" : "appName" } }
```

rozza:
With my example the driver is Spark and it wraps the java driver hence both names.  The platform has JVM information and Spark version information as it differ depending on installation.

estolfo:
Resolution: added an option, ``platform`` to a client that appends the value to the app metadata platform info.

PR #852, opened 2017-02-10
Retry read operations on socket error

reidmorrison:
This change is based on code originally implemented in [mongo_ha](https://github.com/reidmorrison/mongo_ha). mongo_ha has been thoroughly tested in production and is now ready to be merged into the driver directly.

It improves the recoverability of reads and writes when using a replica-set, or standalone servers. In addition it handles network connectivity loss for writes when directly connected to a replica-set. This is important since a primary/master switch forces the server to immediately drop all connections, which results in significant errors on clients if the connection is not automatically re-established.


reidmorrison:
This PR also moves the selection of the server in a few remaining cases into the retry block so that it can recover from a primary change in a replica-set.

All tests pass except for 1 transient error, please retry just that one via Travis.

nmuller:
This would be great to add to the main driver branch.
Without this we cannot safely stepdown without restarting all systems and reconnecting afterwards, which I'd really like to see in the main driver.

reidmorrison:
@estolfo have you had a chance to review these changes?

Please let me know if you have any questions.

estolfo:
Hi @reidmorrison 

Thank you very much for this pull request! I apologize for the delay in getting back to you. I have not finished reviewing it but will have comments very soon. Thanks for your patience,
Emily

estolfo:
Hi @reidmorrison
Thank you again for this pull request! After looking it over and discussing with some colleagues, I'd like to merge in the read retry component but not the write retry. Not including server selection inside #count and #distinct's read_with_retry blocks was an oversight (as we are including the selection in the block for find operations) so thank you for proposing that fix.

For the write retries, the main two reasons I hesitate to merge in the changes are 1) The next version of the server (3.6) will have a proper fix for retriable writes. I hesitate to add in related behavior now and then alter it to work with the server behavior in the next version. 2) We don't want to do things that can't work right across all topologies. Although there is some special handling for sharded clusters in the retryable module already, I'd rather not elaborate on this behavior and wait for the server implementation.

That said, would you like to re-propose a pull request with the read retry changes? I'm happy to merge those in.
Thanks again for all your work on this!
Emily

reidmorrison:
Thank you @estolfo, I removed the write retry changes, as well as any write retry changes in the tests.

lib/mongo/retryable.rb
```diff
@@ -46,18 +36,22 @@ module Retryable
...
+      attempt = 0
       begin
-        block.call
-      rescue Error::SocketError, Error::SocketTimeoutError
-        retry_operation(&block)
+        attempt += 1
+        yield
+      rescue Error::SocketError, Error::SocketTimeoutError => e
+        raise(e) if attempt > cluster.max_read_retries
+        retry_reconnect(e)
```
estolfo: I made a comment below in the retry_reconnect method about sleeping when retrying writes and the same is true here: we weren't sleeping upon Socket errors when retrying reads before so I would hesitate to change the behavior. We were only sleeping when the cluster was sharded, instead of scanning.

lib/mongo/retryable.rb
```diff
@@ -83,11 +77,10 @@ def read_with_retry(attempt = 0, &block)
...
     #
     # @since 2.2.6
-    def read_with_one_retry(&block)
-      block.call
-    rescue Error::SocketError,
-           Error::SocketTimeoutError
-      block.call
+    def read_with_one_retry
+      yield
+    rescue Error::SocketError, Error::SocketTimeoutError => e
```
estolfo: Do we need the e variable?

lib/mongo/retryable.rb
```diff
@@ -46,18 +36,22 @@ module Retryable
     # @return [ Result ] The result of the operation.
     #
     # @since 2.1.0
-    def read_with_retry(attempt = 0, &block)
```
reidmorrison: Instead of using recursion to retry the block it now uses the ruby `retry` method. It also passes the exception to the `retry_reconnect` method so that it can log the exception to make the retry visible. It also logs the apparent slowdown during reconnection.

lib/mongo/retryable.rb
reidmorrison: Excellent catch, will remove the sleep in the `log_retry` method.

lib/mongo/error/operation_failure.rb
```diff
@@ -20,29 +20,37 @@ class Error
...
         'socket exception',
         "can't connect",
-        'no master',
-        'not master',
-        'could not contact primary',
         'connect failed',
         'error querying',
         'could not get last error',
         'connection attempt failed',
         'interrupted at shutdown',
```
reidmorrison: Thank you, good catch.

lib/mongo/error/operation_failure.rb
```diff
@@ -20,29 +20,37 @@ class Error
...
+      # These are magic error messages that could indicate a master change.
+      #
+      # @since 2.4.2
+      WRITE_RETRY_MESSAGES = [
+        'no master',
+        'not master',
+        'could not contact primary',
+        'Not primary',
+        'interrupted at shutdown'
+      ]
```
reidmorrison: Will do

lib/mongo/error/operation_failure.rb
```diff
@@ -20,29 +20,37 @@ class Error
...
-        'could not contact primary',
         'connect failed',
         'error querying',
         'could not get last error',
         'connection attempt failed',
         'interrupted at shutdown',
         'unknown replica set',
         'dbclient error communicating with server'
-      ].freeze
+      ]
```
reidmorrison: Will freeze the list.

lib/mongo/retryable.rb
```diff
@@ -107,23 +100,29 @@ def read_with_one_retry(&block)
     # @return [ Result ] The result of the operation.
     #
     # @since 2.1.0
-    def write_with_retry(&block)
+    def write_with_retry
+      retried = false
```
reidmorrison: Great idea. Using a constant for the retries to be consistent with the read retries.

lib/mongo/retryable.rb
```diff
@@ -46,18 +36,22 @@ module Retryable
...
+        attempt += 1
+        yield
+      rescue Error::SocketError, Error::SocketTimeoutError => e
+        raise(e) if attempt > cluster.max_read_retries
+        log_retry(e)
+        retry
       rescue Error::OperationFailure => e
         if cluster.sharded? && e.retryable?
-          if attempt < cluster.max_read_retries
+          if attempt <= cluster.max_read_retries
```
estolfo: Why don't we make these lines:

```ruby
raise(e) if attempt > cluster.max_read_retries
sleep(cluster.read_retry_interval)
retry
```

for consistency with the block above.

Also, do we intend to only log the retry if there is a SocketError? Would it not be useful and consistent to also log the retry if there is an OperationFailure?

lib/mongo/retryable.rb
```diff
@@ -107,23 +100,28 @@ def read_with_one_retry(&block)
...
-        if e.message.include?(NOT_MASTER) || e.message.include?(COULD_NOT_CONTACT_PRIMARY)
-          retry_operation(&block)
+        raise(e) if attempt > Cluster::MAX_WRITE_RETRIES
+        if e.write_retryable?
+          log_retry(e)
+          retry
         else
-          raise e
+          raise(e)
         end
```
estolfo: This method looks good now. It's consistent with the previous behavior.

lib/mongo/retryable.rb
estolfo: Hi @reidmorrison 
I wanted to check that you saw my above comment?

reidmorrison: Sorry, been very busy. Will push a change shortly.

estolfo:
Thanks so much @reidmorrison for all your work on this!

PR #855, opened 2017-03-06
RUBY-1201 Set max pool size to the min size, if min is greater than max default or provided value

estolfo:


estolfo:
https://evergreen.mongodb.com/version/58bd3cfa3ff1226ea1005051

estolfo:
Replace by: https://github.com/mongodb/mongo-ruby-driver/pull/861

PR #856, opened 2017-03-29
RUBY-1208 Verify that time remaining is positive or nil when checking if socket connect timed out

estolfo:
Ruby's Kernel#select can take a positive value or nil as the last argument (the socket timeout), but not a negative value.

### DIFF
--- a/lib/mongo/socket.rb
+++ b/lib/mongo/socket.rb
@@ -177,7 +177,7 @@ module Mongo
         end
       rescue IO::WaitReadable
         select_timeout = (deadline - Time.now) if deadline
-        unless Kernel::select([@socket], nil, [@socket], select_timeout)
+        unless (select_timeout.nil? || select_timeout > 0) && Kernel::select([@socket], nil, [@socket], select_timeout)
           raise Timeout::Error.new("Took more than #{timeout} seconds to receive data.")
         end
         retry
--- a/lib/mongo/socket.rb
+++ b/lib/mongo/socket.rb
@@ -177,7 +177,7 @@ module Mongo
         end
       rescue IO::WaitReadable
         select_timeout = (deadline - Time.now) if deadline
-        unless (select_timeout.nil? || select_timeout > 0) && Kernel::select([@socket], nil, [@socket], select_timeout)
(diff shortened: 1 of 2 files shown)


estolfo:
https://evergreen.mongodb.com/version/58e4d6592fbabe2bb6005819#/1

alloy:
We’re being hit by this issue, so I was wondering if there’s any ETA for this to be released as a patch release or if we should switch to using unreleased code for now?

If the suggestion is to use unreleased code for now, should we consider the master branch stable or would it be a better idea to cherry-pick just this PR onto the v2.4.1 tag?

estolfo:
Hi @alloy 
Thanks for letting me know this is an issue for you. I was hoping to get in an outstanding pull request before releasing 2.4.2, but it sounds like getting this fix out is more urgent. That said, I'll aim to release 2.4.2 of the driver early next week.
I would not recommend using unreleased code for anything but testing. Have you had a chance to see if this commit fixes the issue for you?
Thanks
Emily

alloy:
Thanks for the fast response! 🙌 

I have not yet, this specific task I’m debugging is a lengthy one to run. I’ll copy over all data to a local machine and see if I can reproduce it there and then try with just this fix applied 👍 

----

For background on how I found this PR, going through the trace tells me that it’s about [this line](https://github.com/mongodb/mongo-ruby-driver/blob/v2.4.1/lib/mongo/socket.rb#L180):

```
11:47:54 ArgumentError: time interval must be positive
11:47:54 /home/jenkins/.rvm/gems/ruby-2.2.2/gems/mongo-2.4.1/lib/mongo/socket.rb:180:in `select'
```

And indeed if I pass a negative argument to `Kernel::select` I get this exact same exception:

```
irb(main):001:0> select(nil, nil, nil, -1)
ArgumentError: time interval must be positive
```

estolfo:
ok, let me know if this commit fixes it for you!

alloy:
Alas I haven’t been able to reproduce it locally yet, there’s probably a unique mix of load required to trigger it. If anything changes I’ll let you know.

estolfo:
Hi @alloy

I've released 2.4.2 with this fix. Let me know if you have any other questions/issues!
Emily

alloy:
Thanks! 🙌  Updated in our app, so we’ll see if these issues go away.

PR #857, opened 2017-04-07
RUBY-1211 Add Keep-Alive to Mongo::Socket::TCP.

mikeantonelli:
https://jira.mongodb.org/browse/RUBY-1211

I wanted to get some eyes on this to get some feedback - would consumers expect this to be enabled by default or should this be configured? If the answer is the later, should it be an explicit connection option that is passed through or via something like `ENV['MONGO_TCP_KEEPALIVE']`?


estolfo:
Thanks @mikeantonelli 
This should have been in the driver originally, so I'll merge this in.
I've tested it [here](https://evergreen.mongodb.com/version/5900604d2fbabe5e32025a89)

PR #858, opened 2017-04-07
RUBY-1211 Add Keep-Alive to Mongo::TCPSocket

mikeantonelli:
Back patch of https://github.com/mongodb/mongo-ruby-driver/pull/857 to `1.x-stable`.

### DIFF
--- a/lib/mongo/connection/socket/tcp_socket.rb
+++ b/lib/mongo/connection/socket/tcp_socket.rb
@@ -40,6 +40,7 @@ module Mongo
         begin
           sock = Socket.new(info[4], Socket::SOCK_STREAM, 0)
           sock.setsockopt(Socket::IPPROTO_TCP, Socket::TCP_NODELAY, 1)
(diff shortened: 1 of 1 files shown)


mikeantonelli:
Thanks for merging @estolfo! Would it be possible to get a 1.12.6 release?

estolfo:
Sure, I'll plan to do it next week.
Thanks again!

mikeantonelli:
@estolfo do you think you'd have some time to cut a 1.12.6 release for this change?

estolfo:
Hi @mikeantonelli 

Because this series (1.x) of the Ruby driver is no longer actively developed and only security fixes will motivate a release, it is not tested in our testing environment. Releasing this series without proper testing doesn't meet our standard for QA, so I would suggest upgrading to the 2.x series.
2.x is compatible back to version 2.4 of the MongoDB server.
Is there a reason you are still using the 1.x series?
Thanks
Emily

PR #861, opened 2017-04-26
RUBY-1201 raise error when instantiating client and min pool size is greater than provided or default max size

estolfo:


perlun:
For reference: [RUBY-1201](https://jira.mongodb.org/browse/RUBY-1201).

PR #862, opened 2017-05-02
MONGOID-4435 Suport mode key as a String in read preference

estolfo:


mariokostelac:
I am wondering is it better than changing the cause of bug - cloning options object. There could be other properties suffering the same way.

estolfo:
I agree, in theory that'd be better but the options are copied [here](https://github.com/mongodb/mongoid/blob/master/lib/mongoid/criteria/queryable.rb#L81), which then leads to this [line](https://github.com/mongodb/mongoid/blob/master/lib/mongoid/criteria/queryable/smash.rb#L97).

Changing that line would cause issues in many other parts of the codebase so either we do a big refactor or we just support mode as a string in the driver...

However, I'll look at this again and see if there's something else I can do in Mongoid.

mariokostelac:
Correct me if I am wrong here. [\_\_deep_clone\_\_](https://github.com/mongodb/mongoid/blob/71c29a80599087008ecfbd057d3d049c2153c7ce/lib/mongoid/criteria/queryable/smash.rb#L21-L27) calls [store](https://github.com/mongodb/mongoid/blob/71c29a80599087008ecfbd057d3d049c2153c7ce/lib/mongoid/criteria/queryable/options.rb#L70-L72), which calls evolve -> storage_pair.
If that is the case, I am wondering if `store` is correct method to be called during the \_\_deep_clone\_\_ operation. 

What is the intention of `evolve` method?

estolfo:
The ``evolve`` method assures that certain objects are serialized to the appropriate BSON types when used in queries. For example, a string like ``59162367f1286e0b9918f721`` will be converted to a BSON::ObjectId Ruby object so that it can properly be serialized to a BSON object id. 

I still haven't found a decent solution for this. I'm leaning towards rewriting how options are deep-copied.

mariokostelac:
@estolfo Do you want me to create a PR for mongoid implementing new __deep_clone__ method?

estolfo:
Hi @mariokostelac 
Please feel free to do so, that'd be awesome! I didn't like the solution of changing the driver for this issue.

mariokostelac:
@estolfo I think you made the right call.

estolfo:
Hey @mariokostelac 
I've pushed a fix for this to Mongoid master and the 7.0-dev branch. I just wrote a #__deep_copy__ method for the criteria options object. Let me know if you get a chance to test it. Thanks!

mariokostelac:
@estolfo I've checked code and tests and it seems to be working ok for tests we have. How can we be sure that all values will have `__deep_copy__` method? It seems that we're not really preventing other things being added into criteria options. Is that some implicit contract inside mongoid codebase?

estolfo:
Hi @mariokostelac 
When a criteria is created, its options are of the class https://github.com/mongodb/mongoid/blob/master/lib/mongoid/criteria/queryable/options.rb and when the criteria is cloned, the options are deep-copied here: https://github.com/mongodb/mongoid/blob/master/lib/mongoid/criteria/queryable.rb#L81
Each value is deep-copied here: https://github.com/mongodb/mongoid/blob/master/lib/mongoid/criteria/queryable/options.rb#L102

So you can be sure that the options is entirely deep-copied when a criteria is cloned. Does that answer your question?
Let me know if I can help clarify anything else.
Emily

mariokostelac:
I think not. We're doing something like:
```ruby
options.each_key do |k, v|
   cloned[k] = v.__deep_copy__
end
```

What happens if v does not respond to __deep_copy__? Or we can be sure that all possible values in options hash have that method implemented?

estolfo:
If your object doesn't implement __deep_copy__, it will fall back to this: https://github.com/mongodb/mongoid/blob/master/lib/mongoid/criteria/queryable/extensions/object.rb#L117

mariokostelac:
Uh, I was missing that. Thanks for putting effort into explaining it to me. Can we backport it to 5.2.0? I can do that tomorrow if you're fine with that.

estolfo:
Hi @mariokostelac 
That would be possible but the changes would need to go into the Origin gem (which was incorporated into the Mongoid codebase with version 6). I've pushed a commit to origin, if you want to test it out: https://github.com/mongoid/origin/commit/80f4924e009fb363e81f7eb33d0cfe0c4413d326
Let me know if it works for you.
Thanks
Emily

estolfo:
Hi @mariokostelac 
Did you get a chance to test out the update to the origin gem?
If you have the time to test it out, I'd be happy to hear if it works for you. Thanks!
Emily

PR #864, opened 2017-05-07
Prevent excessive memory allocation in Mongo::Socket#read_from_socket.

skaes:
PROBLEM: After upgrading the driver gem to version 2.4.1, I noticed a
large increase in the number of bytes allocated per Rails request.
After a lot of debugging I found the problem in the Mongo::Socket class.
The current code calls @socket.read_nonblock(n), where n is
the number of bytes to read. This leads to read_nonblock allocating
a buffer of at least capacity n, on the heap. For large n, say 1MB,
this will result in a large number of differently sized heap chunks
allocated and immediately discarded, because read_nonblock usually
returns only a few kilobytes of data per call.

SOLUTION: read data from the socket in smaller chunks. I chose 4096
bytes, which brought memory allocation back to what it was before the
upgrade.

### DIFF
--- a/lib/mongo/socket.rb
+++ b/lib/mongo/socket.rb
@@ -173,7 +173,9 @@ module Mongo
       deadline = (Time.now + timeout) if timeout
       begin
         while (data.length < length)
-          data << @socket.read_nonblock(length - data.length)
+          bytes_to_read = length - data.length
+          bytes_to_read = 4096 if bytes_to_read > 4096
+          data << @socket.read_nonblock(bytes_to_read)
         end
       rescue IO::WaitReadable
         select_timeout = (deadline - Time.now) if deadline


skaes:
Here's a image showing the difference in memory allocation for my application with/without patch. You can see that the patch also positively affects the number of ruby object allocations. 

screen shot 2017-05-07 at 08 06 49

estolfo:
Hi @skaes 

Thank you for your pull request and for bringing this issue to our attention.
I don't see any information in Ruby documentation or [source](https://github.com/ruby/ruby/blob/0130bb01baed404c0e3c75bd5db472415a6da1d3/ext/stringio/stringio.c#L1452) regarding number of bytes returned from read_nonblock so I'm wondering if you could share some code so I can reproduce this? Can you also tell me what operating system you're on and what version of Ruby?
Thanks

skaes:
Hi @estolfo,

the number of bytes returned from read_nonblock depends on how many bytes are available in the OS buffers, which in turn depends on the speed of the network and the speed with which the mongodb server can send the answer to a given query.

Your source code link points to the wrong file, the code  used is [here](https://github.com/ruby/ruby/blob/0130bb01baed404c0e3c75bd5db472415a6da1d3/io.c#L2649).
which calls https://github.com/ruby/ruby/blob/0130bb01baed404c0e3c75bd5db472415a6da1d3/io.c#L2363 which calls https://github.com/ruby/ruby/blob/0130bb01baed404c0e3c75bd5db472415a6da1d3/string.c#L1988. As you can see, the string is made big enough to hold the number of bytes to be read.

I have debugged and tested the new code on OS X and Ubuntu 14.04. The ruby version is 2.3.3.

I have no simple reproducible test case, because simulating network effects is non trivial.

estolfo:
Hi @skaes 

Thanks for the additional info. I know it's difficult to create a reproducible test case for this but can you tell me what operation you were doing just so I have more context?
I consulted with our Python team and they pointed out that Tornado, an asynchronous web framework, does something similar; however they use 64KB chunks. The value is "read_chunk_size" found here:
https://github.com/tornadoweb/tornado/blob/master/tornado/iostream.py#L159
and it's used here:
https://github.com/tornadoweb/tornado/blob/master/tornado/iostream.py#L1052

Did you test out reading in 64KB chunks, by any chance? I'm curious what your graph would look like when using 64KB.
Thanks
Emily

estolfo:
By the way, I've posted [a question](https://groups.google.com/forum/#!topic/python-tornado/xIah7U9C03Y) to the Tornado google group to find out why Tornado uses 64KB as a default read chunk size.

estolfo:
Additionally, I've opened a ticket with MRI (CRuby) and their response may shed some light on how we should alter this code: https://bugs.ruby-lang.org/issues/13597

estolfo:
btw, I've opened a ticket with our project in Jira to track this: https://jira.mongodb.org/browse/RUBY-1217

estolfo:
Please see my comment here: https://github.com/mongodb/mongo-ruby-driver/pull/874#issuecomment-319009611

PR #871, opened 2017-06-08
RUBY-1125 Document how connection pooling works for Ruby driver

estolfo:


estolfo:
@ajdavis 
I can't select you as a reviewer for some reason but it'd be great if you took a look at this. Thanks in advance.

docs/tutorials/ruby-driver-create-client.txt
```diff
@@ -412,3 +412,63 @@ Details on Timeout Options
...
+for connections.
+
+The client instance opens one additional connection per server in your MongoDB topology for monitoring the server's
+state.
+
+The size of each connection pool is capped at ``max_pool_size``, which defaults to 5. When a thread in your application
+begins an operation on MongoDB, it tries to retrieve a connection from the pool to send that operation on. If there
+are some connections available in the pool, it checks out a connection from the pool and uses it for the operation.
+If there are no connections available and the size of the pool is less than the ``max_pool_size``, a new connection
+will be created. If all connections are in use and the pool has reached its maximum, the thread pauses and waits for a
```
ajdavis: I propose "the thread waits for a connection", not "pauses and waits".

docs/tutorials/ruby-driver-create-client.txt
```diff
@@ -412,3 +412,63 @@ Details on Timeout Options
...
+connection to be returned to the pool by another thread.
+
+The number of seconds the thread will wait is configurable. This setting, called ``wait_queue_timeout``, is defined in
+seconds. It determines how long a thread will wait for a connection to be returned to the pool, when there are no
+connections available. If this timeout is reached, a ``Timeout::Error`` is raised. The default is 1 second.
+
+You can also set the number of connections initialized when the pool is created with the ``min_pool_size``
+setting. This is helpful if your application experiences load spikes and you want to avoid a slow-down when
+connections are created at the time of the spike. The default ``min_pool_size`` is 1.
```
ajdavis: "want to avoid the latency of creating new connections at the beginning of a spike"

docs/tutorials/ruby-driver-create-client.txt
```diff
@@ -412,3 +412,63 @@ Details on Timeout Options
...
+
+For example, a client connected to a 3-node replica set opens 3 monitoring sockets. It also opens as many sockets as
```
ajdavis: Awkward transition, how about "Here is an example of estimating the number of connections a multi-threaded application will open."

docs/tutorials/ruby-driver-create-client.txt
```diff
@@ -412,3 +412,63 @@ Details on Timeout Options
...
+needed to support a multi-threaded application's concurrent operations on each server, up to ``maxPoolSize``. If the
+application only uses the primary (the default), then only the primary connection pool grows and the total connections
+is at most 8 (5 connections for the primary pool + 3 monitoring connections). If the application uses a read
+preference to query the secondaries, their pools also grow and the total connections can reach 18 (5 + 5 + 5 + 3).
```
ajdavis: Does the default min pool size of 1 apply to the secondary pools? I.e., does the driver open an additional connection proactively to each secondary, distinct from the monitors' connections?

docs/tutorials/ruby-driver-create-client.txt
```diff
@@ -412,3 +412,63 @@ Details on Timeout Options
...
+    client = Mongo::Client.new(["localhost:27017"], max_pool_size: 200)
+
+Any number of threads are allowed to wait for connections to become available, and they can wait the default (1 second)
+or the ``wait_queue_timeout`` setting:
+
+.. code-block:: ruby
+
+    client = Mongo::Client.new(["localhost:27017"], wait_queue_timeout: 0.5)
+
+When ``#close`` is called on a client by any thread, all connections are closed:
```
ajdavis: Are checked-out sockets closed immediately, too, or are they closed as they're returned to the pool? (Maybe this level of detail would be excessive here, not sure.)

docs/tutorials/ruby-driver-create-client.txt
estolfo: I didn't get into the details, but the driver actually wraps sockets in a ``Connection`` object. So a connection is not quite synonymous with a socket; a Connection opens a socket when it is used. So to answer your question, yes, a Connection object is created for each secondary but the socket is initialized to nil and no actual "connection" is made.
https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/server/connection.rb#L129-L138

docs/tutorials/ruby-driver-create-client.txt
estolfo: Only non-checked out connections are closed.

docs/tutorials/ruby-driver-create-client.txt
estolfo: done

docs/tutorials/ruby-driver-create-client.txt
estolfo: done

docs/tutorials/ruby-driver-create-client.txt
estolfo: done

ajdavis:
LGTM!

estolfo:
thanks so much @ajdavis

PR #872, opened 2017-06-12
gemspec: Remove relic has_doc attribute

agis:
`has_doc` is always set to `true` so there's no point in setting it (https://github.com/rubygems/rubygems/blob/d8d235f1eeb6da0fb289ce0415ad049153f9adc9/lib/rubygems/specification.rb#L1411).

### DIFF
--- a/mongo.gemspec
+++ b/mongo.gemspec
@@ -27,7 +27,6 @@ Gem::Specification.new do |s|
 
   s.executables       = ['mongo_console']
   s.require_paths     = ['lib']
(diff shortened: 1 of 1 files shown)


estolfo:
It's nice to hear from you @agis after so long!
Thanks for this, I see that this was made the always true 6 years ago https://github.com/rubygems/rubygems/commit/8876b4c0b38307142c4f44800f5ad9b85e5ea013#diff-05477d38d75c98664b0e7dec6eb73b45L232

PR #873, opened 2017-06-17
IPv6 fallback issue

jarthod:
Hello, this MR is not mergeable, it's only to demonstrate the issue:
Basically when IPv6 is down but IPv4 is ok, mongo ruby driver doesn't work.
I got the issue in production and was quite surprised as the mongo server and CLI were still working perfectly (fallbacking to IPv4) but the ruby mongo driver was dead so there's something wrong in the IPv6 failover.

I had a look at the code and the issue seems to be:
- It seems you're using the `socket_timeout` to try the connexion, which by default is undefined (infinite) so the fallback can't really work, I guess you should use the `connect_timeout ` which is more appropriate and defaults to 10s.
- Even if the connexion times out, the error is not catched to go to the next IP, because it's a `Error::SocketTimeoutError` and you're only catching `Error::SocketError` (and it's not an ancestor)
- Finally, there was a test for that but it wasn't doing anything because using an `Address` instance which already have a `@resolver` initialized, it was just a dead test.

In this MR I added the exception and fixed the spec to show you the issue (it's one of my server IPs, packets are filtered which leads to timeout instead of connexion closed). With the timeout set to 1 the fallback seems to be working ok, but with your current value of 0 it waits indefinitely.

Of course you don't want to keep that and you need to fix the default timeout because otherwise it'll only work in specs ;) And I'm not sure how to simulate a timeout locally to leave this test without my IPs.

### DIFF
--- a/lib/mongo/address.rb
+++ b/lib/mongo/address.rb
@@ -173,7 +173,7 @@ module Mongo
           res = FAMILY_MAP[info[4]].new(info[3], port, host)
           res.socket(timeout, ssl_options).connect!.close
           return res
-        rescue IOError, SystemCallError, Error::SocketError, Mongo::Error::SocketTimeoutError => e
+        rescue IOError, SystemCallError, Error::SocketError, Error::SocketTimeoutError => e
           error = e
         end
       end
--- a/lib/mongo/address.rb
+++ b/lib/mongo/address.rb
@@ -173,7 +173,7 @@ module Mongo
           res = FAMILY_MAP[info[4]].new(info[3], port, host)
           res.socket(timeout, ssl_options).connect!.close
           return res
-        rescue IOError, SystemCallError, Error::SocketError => e
+        rescue IOError, SystemCallError, Error::SocketError, Mongo::Error::SocketTimeoutError => e
           error = e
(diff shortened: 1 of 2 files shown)


estolfo:
Hi @jarthod 
Thanks for bringing this up and for the PR demonstrating the issue. I've opened a ticket to track it: https://jira.mongodb.org/browse/RUBY-1222

jarthod:
Ok thanks :)

estolfo:
Hi @jarthod 
I've been working on this here: https://github.com/mongodb/mongo-ruby-driver/pull/876/files

I still need to create some tests still but basically, I pushed the connect_timeout value down to the Address object, initialized with the server options. That connect_timeout is used when the resolver is defined and thus allows it to fallback to ipv4.
The address object also now wraps connecting the socket so it can pass its connect_timeout value (determined from the server options upon initialization). 
The socket itself keeps its original timeout, which is connect_timeout for Monitor/Connection and socket_timeout for the Server/Connection.

estolfo:
replaced by https://github.com/mongodb/mongo-ruby-driver/pull/876/files

PR #875, opened 2017-06-27
RUBY-741 Implement socket max idle time

estolfo:
* Implement maxIdleTime
* Create socket reaper for closing stale sockets in a given connection pool.
* Create PeriodicExecutor that executes reaper methods at regular intervals in a background thread. Currently reapers are CursorReaper and SocketReaper.

### DIFF
--- a/lib/mongo/client.rb
+++ b/lib/mongo/client.rb
@@ -44,6 +44,7 @@ module Mongo
       :id_generator,
       :local_threshold,
       :logger,
+      :max_idle_time,
       :max_pool_size,
       :max_read_retries,
       :min_pool_size,
@@ -167,6 +168,8 @@ module Mongo
     # @option options [ Integer ] :server_selection_timeout The timeout in seconds
     #   for selecting a server for an operation.
     # @option options [ String ] :password The user's password.
+    # @option options [ Integer ] :max_idle_time The maximum seconds a socket can remain idle
+    #   since it has been checked in to the pool.
     # @option options [ Integer ] :max_pool_size The maximum size of the
     #   connection pool.
     # @option options [ Integer ] :min_pool_size The minimum size of the
--- a/lib/mongo/cluster.rb
+++ b/lib/mongo/cluster.rb
@@ -13,7 +13,9 @@
 # limitations under the License.
 
 require 'mongo/cluster/topology'
-require 'mongo/cluster/cursor_reaper'
+require 'mongo/cluster/reapers/socket_reaper'
+require 'mongo/cluster/reapers/cursor_reaper'
+require 'mongo/cluster/periodic_executor'
 require 'mongo/cluster/app_metadata'
 
 module Mongo
@@ -189,9 +191,11 @@ module Mongo
       ) if @servers.size > 1
 
       @cursor_reaper = CursorReaper.new
-      @cursor_reaper.run!
+      @socket_reaper = SocketReaper.new(self)
+      @periodic_executor = PeriodicExecutor.new(@cursor_reaper, @socket_reaper)
+      @periodic_executor.run!
 
-      ObjectSpace.define_finalizer(self, self.class.finalize(pools, @cursor_reaper))
+      ObjectSpace.define_finalizer(self, self.class.finalize(pools, @periodic_executor))
     end
 
     # Finalize the cluster for garbage collection. Disconnects all the scoped
@@ -206,10 +210,9 @@ module Mongo
     # @return [ Proc ] The Finalizer.
     #
     # @since 2.2.0
-    def self.finalize(pools, cursor_reaper)
+    def self.finalize(pools, periodic_executor)
       proc do
-        begin; cursor_reaper.kill_cursors; rescue; end
-        cursor_reaper.stop!
+        periodic_executor.stop!
         pools.values.each do |pool|
           pool.disconnect!
         end
@@ -371,8 +374,7 @@ module Mongo
     #
     # @since 2.1.0
     def disconnect!
-      begin; @cursor_reaper.kill_cursors; rescue; end
-      @cursor_reaper.stop!
+      @periodic_executor.stop!
       @servers.each { |server| server.disconnect! } and true
     end
 
@@ -387,7 +389,7 @@ module Mongo
     def reconnect!
       scan!
       servers.each { |server| server.reconnect! }
-      @cursor_reaper.restart! and true
+      @periodic_executor.restart! and true
     end
 
     # Add hosts in a description to the cluster.
--- a/lib/mongo/cluster/periodic_executor.rb
+++ b/lib/mongo/cluster/periodic_executor.rb
@@ -1,0 +1,106 @@
+# Copyright (C) 2014-2017 MongoDB, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module Mongo
+
+  class Cluster
+
+    # A manager that calls #execute on its executors at a regular interval.
+    #
+    # @api private
+    #
+    # @since 2.5.0
+    class PeriodicExecutor
+
+      # The default time interval for the periodic executor to execute.
+      #
+      # @since 2.5.0
+      FREQUENCY = 5
+
+      # Create a periodic executor.
+      #
+      # @example Create a PeriodicExecutor.
+      #   Mongo::Cluster::PeriodicExecutor.new(reaper, reaper2)
+      #
+      # @api private
+      #
+      # @since 2.5.0
+      def initialize(*executors)
+        @thread = nil
+        @executors = executors
+      end
+
+      # Start the thread.
+      #
+      # @example Start the periodic executor's thread.
+      #   periodic_executor.run!
+      #
+      # @api private
+      #
+      # @since 2.5.0
+      def run!
+        @thread && @thread.alive? ? @thread : start!
+      end
+      alias :restart! :run!
+
+      # Stop the executor's thread.
+      #
+      # @example Stop the executors's thread.
+      #   periodic_executor.stop!
+      #
+      # @api private
+      #
+      # @since 2.5.0
+      def stop!
+        begin; flush; rescue; end
+        @thread.kill && @thread.stop?
+      end
+
+      # Trigger an execute call on each reaper.
+      #
+      # @example Trigger all reapers.
+      #   periodic_executor.execute
+      #
+      # @api private
+      #
+      # @since 2.5.0
+      def execute
+        @executors.each(&:execute) and true
+      end
+
+      # Execute all pending operations.
+      #
+      # @example Execute all pending operations.
+      #   periodic_executor.flush
+      #
+      # @api private
+      #
+      # @since 2.5.0
+      def flush
+        @executors.each(&:flush) and true
+      end
+
+      private
+
+      def start!
+        @thread = Thread.new(FREQUENCY) do |i|
+          loop do
+            sleep(i)
+            execute
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/mongo/cluster/reapers/socket_reaper.rb
+++ b/lib/mongo/cluster/reapers/socket_reaper.rb
@@ -1,0 +1,59 @@
+# Copyright (C) 2014-2017 MongoDB, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module Mongo
+
+  class Cluster
+
+    # A manager that calls a method on each of a cluster's pools to close stale
+    #  sockets.
+    #
+    # @api private
+    #
+    # @since 2.5.0
+    class SocketReaper
+
+      # Initialize the SocketReaper object.
+      #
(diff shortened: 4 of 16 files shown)


estolfo:
Hi @ShaneHarvey and @ajdavis 
I can't add either of you as an official reviewer but I would be grateful if you would take a look at this. Thank you!

lib/mongo/client.rb
```diff
@@ -157,6 +158,8 @@ def hash
     # @option options [ Integer ] :server_selection_timeout The timeout in seconds
     #   for selecting a server for an operation.
     # @option options [ String ] :password The user's password.
+    # @option options [ Integer ] :max_idle_time The maximum seconds a socket can remain
+    #   idle in the connection pool after it has been checked out and checked back in.
```
ajdavis: This is a little confusing, perhaps "maximum seconds a socket can remain idle since it has been checked in to the pool."

lib/mongo/cluster/periodic_executor.rb
```diff
@@ -0,0 +1,93 @@
+# Copyright (C) 2014-2016 MongoDB, Inc.
```
ajdavis: 2017

lib/mongo/cluster/periodic_executor.rb
```diff
@@ -0,0 +1,93 @@
...
+      #
+      # @example Start the periodic executor's thread.
+      #   periodic_executor.run!
+      #
+      # @api private
+      #
+      # @since 2.5.0
+      def run!
+        @thread && @thread.alive? ? @thread : start!
+      end
```
ajdavis: Could many *other* threads all call run! at once? If so you need a lock here to ensure you don't start many threads.

lib/mongo/cluster/reapers/socket_reaper.rb
```diff
@@ -0,0 +1,53 @@
+# Copyright (C) 2014-2015 MongoDB, Inc.
```
ajdavis: 2017

lib/mongo/server/connection_pool/queue.rb
```diff
@@ -165,12 +164,45 @@ def wait_timeout
...
+        #   option is set.
+        #
+        # @example Close the stale sockets
+        #   queue.close_stale_sockets!
+        #
+        # @since 2.5.0
+        def close_stale_sockets!
+          if max_idle_time
+            mutex.synchronize do
+              queue.reverse_each do |connection|
```
ajdavis: Nice.

lib/mongo/server/connection_pool/queue.rb
```diff
@@ -165,12 +164,45 @@ def wait_timeout
...
+                if last_checkin = connection.last_checkin
+                  connection.disconnect! if (Time.now - last_checkin) > max_idle_time
```
ajdavis: Does disconnect! remove the connection from the queue? Is it ok to continue iterating in reverse_each as connections are being removed?

lib/mongo/cluster.rb
```diff
@@ -359,8 +363,8 @@ def servers
     #
     # @since 2.1.0
     def disconnect!
```
ajdavis: Is this called by the Ruby garbage collector? If so I'd like to discuss whether a deadlock is possible akin to PYTHON-799. If this isn't called during garbage collection, nevermind. =)

lib/mongo/server/connection_pool/queue.rb
```diff
@@ -165,12 +164,45 @@ def wait_timeout
           @wait_timeout ||= options[:wait_queue_timeout] || WAIT_TIMEOUT
         end
 
+        # The maximum seconds a socket can remain idle in the connection pool after it
+        #   has been checked out and checked back in.
```
ShaneHarvey: What Jesse said: "maximum seconds a socket can remain idle since it has been checked in to the pool."

lib/mongo/server/connection_pool/queue.rb
```diff
@@ -165,12 +164,45 @@ def wait_timeout
...
+        #
+        # @since 2.5.0
+        def close_stale_sockets!
+          if max_idle_time
+            mutex.synchronize do
+              queue.reverse_each do |connection|
+                if last_checkin = connection.last_checkin
+                  connection.disconnect! if (Time.now - last_checkin) > max_idle_time
+                end
+                break if queue.index(connection) == min_size
```
ShaneHarvey: Why break at the min size? Shouldn't this method check all the sockets in the pool?

lib/mongo/server/connection_pool/queue.rb
```diff
@@ -88,7 +87,7 @@ def disconnect!
         # @since 2.0.0
         def enqueue(connection)
           mutex.synchronize do
-            queue.unshift(connection)
+            queue.unshift(connection.record_checkin!)
             resource.broadcast
           end
         end
```
ShaneHarvey: When the queue is initialized with `@queue = Array.new(min_size) { create_connection }`  will those connections have a checkin time?

lib/mongo/server/connection_pool/queue.rb
estolfo: No, {{connection.disconnect!}} only disconnects the socket.
Connection objects are wrappers for sockets; they create and connect sockets or disconnect them. They are only removed from the queue when they are checked out.

lib/mongo/server/connection_pool/queue.rb
estolfo: No. ``create_connection`` calls these methods:
https://github.com/estolfo/mongo-ruby-driver/blob/master/lib/mongo/cluster.rb#L269
https://github.com/estolfo/mongo-ruby-driver/blob/master/lib/mongo/server/connection_pool.rb#L132

which eventually creates a Connection, with ``last_checkin`` initialized to nil [here](https://github.com/estolfo/mongo-ruby-driver/blob/RUBY-741-max-idle-time/lib/mongo/server/connection.rb#L142).

Also, ``last_checkin`` is set to nil when the connection is [disconnected](https://github.com/estolfo/mongo-ruby-driver/blob/RUBY-741-max-idle-time/lib/mongo/server/connection.rb#L91).

lib/mongo/server/connection_pool/queue.rb
estolfo: If I want to guarantee a minimum pool size, I shouldn't disconnect any more than (size - min_size) connections, no?

lib/mongo/client.rb
estolfo: done

lib/mongo/cluster/periodic_executor.rb
estolfo: done

lib/mongo/cluster/reapers/socket_reaper.rb
estolfo: done

lib/mongo/server/connection_pool/queue.rb
estolfo: done

lib/mongo/cluster/periodic_executor.rb
estolfo: [This is called](https://github.com/mongodb/mongo-ruby-driver/pull/875/files#diff-e1d286ad6aafcf552750e39f66afcf57R184) only during the #initialize method of a cluster, which is called when a client is created. It's the same design as our [SDAM monitor](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/server/monitor.rb#L120-L127).

I don't believe many other threads would be able to call #run! at once, as initializing a client won't return until the cluster/monitor/periodic_executor have all been initialized and started.
Do you still see a potential issue with this?

lib/mongo/cluster.rb
estolfo: hmm I don't think you could get into a deadlock, disconnect! is not called by the garbage collector. I need to read the python ticket a few more times because I don't think I fully understand the scenario.

This did make me realize that I should called ``periodic_executor.stop!`` before calling #execute on it in the finalize method called by the garbage collector [here](https://github.com/mongodb/mongo-ruby-driver/pull/875/files#diff-e1d286ad6aafcf552750e39f66afcf57R203).

lib/mongo/server/connection_pool/queue.rb
ajdavis: What PyMongo does here, which I think is correct, is to close *all* sockets that have been idle too long. Then, if the pool has shrunk smaller than minPoolSize as a consequence, PyMongo proactively opens enough new sockets to reach minPoolSize again.

lib/mongo/server/connection_pool/queue.rb
ShaneHarvey: I think all connections in the pool, even the initial ones, must have `last_checkin` set. Otherwise, those connections will never be disconnected according to maxIdleTime until they are checked out and checked back in.

lib/mongo/server/connection_pool/queue.rb
estolfo: Yes, I saw that PyMongo does that but I didn't do it because it would require doing two loops through all connections while the lock is held. I don't know if the benefit of closing, then reconnecting the min number of sockets is worth it? Maybe it is up to the user then to make sure min_size isn't ridiculously high so that this method doesn't hold the lock for so long?

lib/mongo/server/connection_pool/queue.rb
estolfo: The Ruby driver doesn't connect the Connection objects until they are checked out of the pool the first time. So they wouldn't need to be disconnected until they are checked out and checked back in (and at that point, they would have a last_checkin value). Does that make sense?

[Here](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/server/connection.rb#L135) is where the Connection object is initialized.
And then, when it is used for the first time, [this method](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/server/connectable.rb#L93-L103) is called that connects it and creates a socket for the Connection.

lib/mongo/server/connection_pool/queue.rb
estolfo: UPDATE: Putting the reconnect in was a lot simpler than I thought and didn't require iterating over the list twice. I've updated the code to call ``connect!`` on the connection after it has been disconnected if its index is < min_pool_size.

lib/mongo/server/connection_pool/queue.rb
ShaneHarvey: Okay, so this does not conflict with max idle time. Thanks for explaining.

lib/mongo/server/connection_pool/queue.rb
```diff
@@ -165,12 +164,48 @@ def wait_timeout
...
+        def close_stale_sockets!
+          if max_idle_time
+            mutex.synchronize do
+              queue.reverse_each do |connection|
+                if last_checkin = connection.last_checkin
+                  if (Time.now - last_checkin) > max_idle_time
+                    connection.disconnect!
+                    if queue.index(connection) < min_size
+                      connection.connect!
+                    end
```
ShaneHarvey: Does `connection.connect!` block? In Python we have a rule: no blocking I/O while holding a lock. Does Ruby follow a similar mantra?

lib/mongo/server/connection_pool/queue.rb
estolfo: thank you for your review!

lib/mongo/server/connection_pool/queue.rb
estolfo: That's a good point but don't you risk a socket being checked out and messing up the pool size in [these lines](https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/pool.py#L727-L730) ? More specifically, what if a socket is checked out while one is being connected (L730) and you're already at max_pool_size? Wouldn't you end up with max_pool_size + 1?

ShaneHarvey: Yep, it's a classic [TOCTOU](https://en.wikipedia.org/wiki/Time_of_check_to_time_of_use) bug. We discussed fixing it in [PYTHON-1193](https://jira.mongodb.org/browse/PYTHON-1193), but the issue went away. I'll probably revive the fix... I suspect a future change will make it more likely to be hit.

Holding the lock is simpler and avoids that possibility so it's up to you.

ShaneHarvey: Nice! LGTM.

lib/mongo/cluster.rb
```diff
@@ -194,10 +198,10 @@ def initialize(seeds, monitoring, options = Options::Redacted.new)
     # @return [ Proc ] The Finalizer.
     #
     # @since 2.2.0
-    def self.finalize(pools, cursor_reaper)
+    def self.finalize(pools, periodic_executor)
       proc do
-        begin; cursor_reaper.kill_cursors; rescue; end
-        cursor_reaper.stop!
+        periodic_executor.stop!
+        begin; periodic_executor.execute; rescue; end
```
saghm: It looks like these two lines will call CursorReaper#kill_cursors twice, once through the `finalize` alias in the call to PeriodicExecutor#stop! and then again through the call to the `execute` alias through PeriodicExecutor#execute. It doesn't look like calling it twice causes any issues, although since it does require acquiring a lock on a mutex, I'm wondering if it might have some performance cost.

lib/mongo/server/connection_pool/queue.rb
```diff
@@ -165,12 +164,57 @@ def wait_timeout
...
+
+        # Close sockets that have been open for longer than the max idle time, if the
+        #   option is set.
+        #
+        # @example Close the stale sockets
+        #   queue.close_stale_sockets!
+        #
+        # @since 2.5.0
+        def close_stale_sockets!
+          if max_idle_time
```
saghm: Would it be more readable to `return unless max_idle_time` instead of wrapping everything in an if block? I'm not sure if this style is frowned upon or not, but I think it might help a bit due to the level of nesting; I personally had a little trouble telling if the `to_connect.select ...` block was within the if block or not.

lib/mongo/cluster.rb
estolfo: Good catch, thanks!
I changed the #stop! method on the periodic executor to only stop the thread and created another method, #flush that calls #flush on the executors to execute pending operations. What do you think?

lib/mongo/server/connection_pool/queue.rb
estolfo: Sure, I changed the method to return unless ``max_idle_time``. Does it seem clearer now?

lib/mongo/cluster.rb
saghm: Sounds good to me

lib/mongo/server/connection_pool/queue.rb
saghm: Yep, thank for indulging me!

PR #877, opened 2017-07-03
RUBY-1223 Ignore lastWrite when comparing server descriptions

rinmu:
This pull request solves the problem discussed at the following URL.

https://groups.google.com/forum/#!msg/mongodb-user/moEu4STgzUg/HMnZhj2mDQAJ

Since lastWrite changes like localTime, I think it is correct to ignore them when comparing config.


estolfo:
Hi @rinmu 
Thank you very much for this pull request!

PR #880, opened 2017-07-21
Cache SCRAM-SHA-1 ClientKey

saghm:
This fixes RUBY-1213

### DIFF
--- a/lib/mongo/auth/scram/conversation.rb
+++ b/lib/mongo/auth/scram/conversation.rb
@@ -110,6 +110,13 @@ module Mongo
         # @since 2.0.0
         def continue(reply)
           validate_first_message!(reply)
+
+          # The salted password needs to be calculated now; otherwise, if the
+          # client key is cached from a previous authentication, the salt in the
+          # reply will no longer be available for when the salted password is
+          # needed to calculate the server key.
+          salted_password
+
           Protocol::Query.new(
             user.auth_source,
             Database::COMMAND,
@@ -247,9 +254,9 @@ module Mongo
         #
         # @since 2.0.0
         def client_key
-          key = @client_key ||= hmac(salted_password, CLIENT_KEY)
-          user.client_key ||= key
-          key
+          @client_key ||= hmac(salted_password, CLIENT_KEY)
+          user.client_key ||= @client_key
+          @client_key
         end
 
         # Client proof algorithm implementation.
--- a/lib/mongo/auth/scram/conversation.rb
+++ b/lib/mongo/auth/scram/conversation.rb
@@ -188,7 +188,7 @@ module Mongo
         def initialize(user)
           @user = user
           @nonce = SecureRandom.base64
-          @client_key = user.client_key
+          @client_key = user.send(:client_key)
         end
 
         private
@@ -255,7 +255,7 @@ module Mongo
         # @since 2.0.0
         def client_key
           @client_key ||= hmac(salted_password, CLIENT_KEY)
-          user.client_key ||= @client_key
+          user.send(:client_key=, @client_key) unless user.send(:client_key)
           @client_key
         end
 
--- a/lib/mongo/auth/scram/conversation.rb
+++ b/lib/mongo/auth/scram/conversation.rb
@@ -181,6 +181,7 @@ module Mongo
         def initialize(user)
           @user = user
           @nonce = SecureRandom.base64
+          @client_key = user.client_key
         end
 
         private
@@ -246,7 +247,9 @@ module Mongo
         #
         # @since 2.0.0
         def client_key
-          @client_key ||= hmac(salted_password, CLIENT_KEY)
+          key = @client_key ||= hmac(salted_password, CLIENT_KEY)
+          user.client_key ||= key
+          key
         end
 
         # Client proof algorithm implementation.
--- a/lib/mongo/auth/user.rb
+++ b/lib/mongo/auth/user.rb
@@ -49,10 +49,6 @@ module Mongo
       # @return [ Array<String> ] roles The user roles.
       attr_reader :roles
 
-      # Sets the client key for the user.
-      # @return [ String ] The client key for the user.
-      attr_accessor :client_key
-
       # Determine if this user is equal to another.
       #
       # @example Check user equality.
@@ -135,9 +131,10 @@ module Mongo
       # @option options [ String ] :password The user's password.
       # @option options [ Symbol ] :auth_mech The authorization mechanism.
       # @option options [ Array<String>, Array<Hash> ] roles The user roles.
+      #@option options [ String ] :client_key The user's client key cached from a previous authentication on the same connection.
       #
       # @since 2.0.0
-      def initialize(options, client_key = nil)
+      def initialize(options)
         @database = options[:database] || Database::ADMIN
         @auth_source = options[:auth_source] || @database
         @name = options[:user]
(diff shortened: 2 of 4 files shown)


estolfo:
I believe there's another area that needs some caching: In the #authenticate method, a new user object [is created each time](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/server/connection.rb#L193). We would need to cache the user object as well if we want to benefit from the client key being cached on the user object.

lib/mongo/server/connection.rb
```diff
@@ -190,9 +190,9 @@ def handshake!
 
       def authenticate!
         if options[:user] || options[:auth_mech]
-          user = Auth::User.new(Options::Redacted.new(:auth_mech => default_mechanism).merge(options))
+          @user ||= Auth::User.new(Options::Redacted.new(:auth_mech => default_mechanism).merge(options))
```
saghm: From looking at the other authentication mechanisms, none of them seem to cache any state from the authentication conversation, so I think it should be fine to cache the user without checking that the mechanism is SCRAM.

estolfo: We unfortunately can't cache the user and must set it each time we connect because the auth mechanism could have changed. This commit describes the scenario: https://github.com/mongodb/mongo-ruby-driver/commit/49f534668d1a4d32c44d300557da59fbd4765d0f

saghm: Thanks for the info; I've updated it to properly handle this.

Additionally, there was an issue with the salted password not being able to be calculated when the client key was cached because it was no longer needed until generating the server key, and by that time a new reply had come since the one that contained the salt. I fixed this by explicitly calling `salted_password` in `SCRAM::Conversation#continue`.

lib/mongo/auth/user.rb
```diff
@@ -49,6 +49,10 @@ class User
       # @return [ Array<String> ] roles The user roles.
       attr_reader :roles
 
+      # Sets the client key for the user.
+      # @return [ String ] The client key for the user.
+      attr_accessor :client_key
```
estolfo: Let's keep this method private so that if we need to change how the client key is cached as some point, we're not bound to a public API.

You can make this method private, then access it with ``user.send(:client_key)`` or you can access it with ``#instance_variable_get`` on the user.

lib/mongo/server/connection.rb
```diff
@@ -190,9 +190,12 @@ def handshake!
 
       def authenticate!
         if options[:user] || options[:auth_mech]
-          user = Auth::User.new(Options::Redacted.new(:auth_mech => default_mechanism).merge(options))
+          user = Auth::User.new(Options::Redacted.new(:auth_mech => default_mechanism).merge(options), @client_key)
```
estolfo: since everything is already passed to the user initialize method as part of the options hash, if we want to be consistent, we could make this:

```ruby
user = Auth::User.new(Options::Redacted.new(:auth_mech => default_mechanism, :client_key => @client_key).merge(options))
```

and then extract the client key in the #initialize method of the user:
```ruby
@client_key = options[:client_key]
```

lib/mongo/auth/scram/conversation.rb
```diff
@@ -247,6 +255,8 @@ def client_final
         # @since 2.0.0
         def client_key
           @client_key ||= hmac(salted_password, CLIENT_KEY)
+          user.client_key ||= @client_key
+          @client_key
```
estolfo: I don't think you need to return @client_key explicitly. It will either return the ``user.client_key`` value, if there is one, or the ``@client_key`` value after being assigned to ``user.client_key``.

lib/mongo/auth/user.rb
```diff
@@ -133,14 +137,15 @@ def hashed_password
...
-      def initialize(options)
+      def initialize(options, client_key = nil)
         @database = options[:database] || Database::ADMIN
         @auth_source = options[:auth_source] || @database
         @name = options[:user]
         @password = options[:password] || options[:pwd]
         @mechanism = options[:auth_mech] || :mongodb_cr
         @auth_mech_properties = options[:auth_mech_properties] || {}
         @roles = options[:roles] || []
+        @client_key = client_key
```
estolfo: See my comment in ``server/connection.rb``: You can just add the ``client_key`` to the options hash passed to the ``#initialize`` method, as all the arguments are in there anyway.

spec/mongo/auth/scram_spec.rb
```diff
@@ -67,8 +67,17 @@
...
       expect(login['ok']).to eq(1)
+      expect(root_user.client_key).not_to be_nil
+    end
+
+    it 'raises an exception when an incorrect client key is set', if: list_command_enabled? do
+      root_user.client_key = "incorrect client key"
+      expect {
+        cr.login(connection)
+      }.to raise_error(Mongo::Auth::Unauthorized)
+      root_user.client_key = nil
```
estolfo: let's put this line in an ``after`` block for clarity.

You can put this above the two ``it`` blocks:

```ruby
after do
  root_user.client_key = nil
end
```

estolfo:
Sorry for the fragmented review. Github was having issues yesterday so I couldn't get all my comments to be applied. Do you know if I can edit or add to a review? I didn't seem to be able to do that so I've put in a few "single comments" above.

saghm:
No worries about the multiple comments! I don't think there's a way to edit past comments into a review; from my limited understanding of Github's review functionality, it's basically just a way to bulk send a bunch of comments without multiple emails and tack on an "approved/changes requested".

lib/mongo/auth/scram/conversation.rb
saghm: After changing User#client_key to be private, I can't use `||=` anymore and instead need to do `user.send(:client_key=, @client_key) unless user.send(:client_key)`, so I've left it in.

lib/mongo/auth/user.rb
saghm: Done.

lib/mongo/auth/user.rb
saghm: Done.

lib/mongo/server/connection.rb
saghm: Done.

spec/mongo/auth/scram_spec.rb
saghm: Done.

lib/mongo/auth/user.rb
```diff
@@ -154,6 +156,12 @@ def initialize(options)
       def spec
         { pwd: hashed_password, roles: roles }
       end
+
+      private
+
+      # Sets the client key for the user.
+      # @return [ String ] The client key for the user.
+      attr_accessor :client_key
```
estolfo: you can make this ``attr_reader`` if you want, and then the line above ``user.instance_variable_set(:@client_key, client_key)`` to avoid the awkwardness of ``user.send(:client_key=, @client_key)``

lib/mongo/auth/user.rb
```diff
@@ -131,6 +131,7 @@ def hashed_password
       # @option options [ String ] :password The user's password.
       # @option options [ Symbol ] :auth_mech The authorization mechanism.
       # @option options [ Array<String>, Array<Hash> ] roles The user roles.
+      #@option options [ String ] :client_key The user's client key cached from a previous authentication on the same connection.
```
estolfo: minor nitpick: spacing after the ``#`` and usually we don't let lines get as long as this one.

PR #881, opened 2017-07-24
Separate concept of user-defined read preference and server selection

estolfo:
The driver made no distinction between a read preference set in options and a server selector. A server selector must use the user-defined read preference or default to a primary. The read preference is either set in options or is nil.

These changes are necessary when implementing OP_MSG so that we aren't sending a read preference to the server if the user hasn't defined one. In the case that the user hasn't defined one on the client or collection, we must not send the read preference so that the default server-side can be used.

### DIFF
--- a/lib/mongo/collection/view/readable.rb
+++ b/lib/mongo/collection/view/readable.rb
@@ -134,8 +134,9 @@ module Mongo
           cmd[:maxTimeMS] = opts[:max_time_ms] if opts[:max_time_ms]
           cmd[:readConcern] = collection.read_concern if collection.read_concern
           read_pref = opts[:read] || read_preference
+          selector = ServerSelector.get(read_pref || server_selector)
           read_with_retry do
-            server = ServerSelector.get(read_pref || server_selector).select_server(cluster, false)
+            server = selector.select_server(cluster, false)
             apply_collation!(cmd, server, opts)
             Operation::Commands::Command.new({
                                                :selector => cmd,
@@ -170,8 +171,9 @@ module Mongo
           cmd[:maxTimeMS] = opts[:max_time_ms] if opts[:max_time_ms]
           cmd[:readConcern] = collection.read_concern if collection.read_concern
           read_pref = opts[:read] || read_preference
+          selector = ServerSelector.get(read_pref || server_selector)
           read_with_retry do
-            server = ServerSelector.get(read_pref || server_selector).select_server(cluster, false)
+            server = selector.select_server(cluster, false)
             apply_collation!(cmd, server, opts)
             Operation::Commands::Command.new({
                                                :selector => cmd,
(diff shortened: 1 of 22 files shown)


estolfo:
@saghm Would you mind reviewing this? I can't add you as an official reviewer for some reason.

saghm:
Will do; it appears that I don't currently have write access to the repo, so I suspect that's necessary for me to be set as a reviewer. I'll look into getting that changed.

lib/mongo/collection/view/readable.rb
```diff
@@ -133,15 +133,15 @@ def count(opts = {})
           cmd[:limit] = opts[:limit] if opts[:limit]
           cmd[:maxTimeMS] = opts[:max_time_ms] if opts[:max_time_ms]
           cmd[:readConcern] = collection.read_concern if collection.read_concern
-          preference = ServerSelector.get(opts[:read] || read)
+          read_pref = opts[:read] || read_preference
           read_with_retry do
-            server = preference.select_server(cluster, false)
+            server = ServerSelector.get(read_pref || server_selector).select_server(cluster, false)
```
saghm: Would it still make sense to call`ServerSelector#get` above the retry loop (and store the result in a separate variable from `read_pref`) here and below in `distinct`? I'm not sure how the block is used internally in read_with_retry, but I'm guessing it should be fine to use the same variable in each iteration.

lib/mongo/grid/stream/read.rb
```diff
@@ -167,7 +167,8 @@ def ensure_readable!
           end
 
           def view
-            @view ||= fs.chunks_collection.find({ :files_id => file_id }, options).read(read_preference).sort(:n => 1)
+            @view ||= (opts = options.merge(read: read_preference) if read_preference
```
saghm: I'm having a little trouble parsing exactly what this is doing.  Is line 171 a separate statement being grouped with the previous one for the purpose of having its return value stored in view? If so, I think it might be clearer to move the assignment of `opts` above the line 170 and then change line 170 to `@view ||= fs.chunks_collection.find(...)`.

lib/mongo/collection/view/readable.rb
estolfo: Good idea. The only code we want to retry is the server selection, not the instantiation of a server selector. I'll change it, as well as in #distinct.

lib/mongo/grid/stream/read.rb
estolfo: I wanted that whole block of code only executed once so I put it all in a memoized ()
``options.merge(read: read_preference)`` creates a new object and since the read object is immutable, we only have to do it once when the ``@view`` is created. Do you have another suggestion for how we can format this method to be more readable and so that the lines are only executed once?

saghm: The only thing that comes to mind is putting those two lines in a separate function and then conditionally-assigning the result of calling it to view. That may or may not be worth it depending on whether you want keep all of the logic together rather than separating it out, though.

estolfo: I think we can just keep all the logic in one place, esp since calling a separate method is less efficient.

PR #882, opened 2017-07-24
RUBY-1227 Handle min_pool_size being a present key but nil value

estolfo:


spec/mongo/client_spec.rb
```diff
@@ -444,6 +444,25 @@
...
+          {
+              :min_pool_size => nil,
+              :max_pool_size => nil
+          }
+        end
+
+        it 'does not set either option' do
+          expect(client.options[:max_pool_size]).to be_nil
+          expect(client.options[:min_pool_size]).to be_nil
+        end
```
perlun: You could even change this to something like:

```ruby
it 'does not raise an exception when creating a new client'
  expect { client }.to_not raise_error
end
```

...if you like. That will make it even more clear that this test catches the (incorrect) exception being thrown. But both ways are fine, it's just a matter of "what do we want the test to verify".

estolfo: I think I'll leave it as verifying that neither option value is set, just to be sure that we don't end up with options erroneously being set. I'd like to be a little more specific than just making sure an exception is not raised.

PR #883, opened 2017-07-31
RUBY-1226 Implement Drivers Sessions API and support for causally consistent reads

estolfo:


lib/mongo/collection.rb
```diff
@@ -184,11 +190,11 @@ def create
       if (options[:collation] || options[Operation::COLLATION]) && !server.features.collation_enabled?
         raise Error::UnsupportedCollation.new
       end
-      Operation::Commands::Create.new({
-                                        selector: operation,
-                                        db_name: database.name,
-                                        write_concern: write_concern
-                                      }).execute(server)
+      Operation::Commands::Create.new(
```
saghm: Are operations like creating and deleting collection supposed to also be done in the context of the session? I can't find anything specific in the spec about them, so I'm not sure.

lib/mongo/database.rb
```diff
@@ -155,11 +158,11 @@ def collections
     def command(operation, opts = {})
       preference = ServerSelector.get(opts[:read] || ServerSelector::PRIMARY)
       server = preference.select_server(cluster)
-      Operation::Commands::Command.new({
+      Operation::Commands::Command.new(
```
saghm: I'm also not sure if commands should be executed in the context of a session or not

lib/mongo/monitoring/event/command_failed.rb
```diff
@@ -55,14 +57,15 @@ class CommandFailed
         # @param [ Float ] duration The duration the command took in seconds.
         #
```
saghm: The `reply` parameter should probably be documented here as well.

lib/mongo/monitoring/event/command_failed.rb
```diff
@@ -79,15 +82,16 @@ def initialize(command_name, database_name, address, request_id, operation_id, m
         # @return [ CommandFailed ] The event.
```
saghm: Ditto above comment about documenting `reply`

lib/mongo/collection.rb
estolfo: That's a good question - one that I asked myself as well.
The spec says 'The server reports the operation time in the response to CRUD commands.'
The ``with_session`` method extracts the operationTime from the operation response and given that the spec says the value is reported in CRUD commands, I've only wrapped CRUD commands in the ``with_session`` method. I'll keep my eye on the spec though, and test other commands, to see if operationTime is reported in more than just ones of the CRUD type.

lib/mongo/database.rb
estolfo: same comment as above - I think only CRUD commands report operationTime.

lib/mongo/monitoring/event/command_failed.rb
estolfo: thanks!

lib/mongo/database.rb
xdg: It's not just CRUD.  Drivers are supposed to look for operationTime in *all* database responses, including errors.  If that's not updated in the spec, it's supposed to be.

estolfo: Thanks, I've written a comment in the google doc next to the text saying it reports it from CRUD commands.

PR #884, opened 2017-08-01
RUBY-1224 Implement arrayFilters option for update operations

saghm:
Fixes RUBY-1224


lib/mongo/bulk_write/combineable.rb
```diff
@@ -28,6 +28,10 @@ module Combineable
       # @return [ true, false ] has_collation Whether one or more operations has a collation defined.
       attr_reader :has_collation
 
+      #return [true, false] has_array_filters Whether one or more operations specifies an array
```
estolfo: Format should be ``# @return`` so that yard documentation can pick up the ``@return`` tag properly.
Also, we usually put a space before/ after the ``[`` and ``]``

lib/mongo/collection.rb
```diff
@@ -582,9 +586,11 @@ def find_one_and_delete(filter, options = {})
     # @option options [ true, false ] :upsert Whether to upsert if the document doesn't exist.
     # @option options [ true, false ] :bypass_document_validation Whether or
     #   not to skip document level validation.
-    # @option options [ Hash ] :write_concern The write concern options.
+    # @option options [ Hash ] :write_concern he write concern options.
```
estolfo: small delete mistake here

lib/mongo/collection.rb
```diff
@@ -582,9 +586,11 @@ def find_one_and_delete(filter, options = {})
...
     #   Defaults to the collection's write concern.
-    # @option options [ Hash ] :collation The collation to use.
+    # @option options [ Hash ] :collation The ollation to use.
```
estolfo: same here - the c went missing = )

lib/mongo/collection/view/writable.rb
```diff
@@ -107,6 +109,7 @@ def find_one_and_update(document, opts = {})
           cmd[:maxTimeMS] = max_time_ms if max_time_ms
           cmd[:bypassDocumentValidation] = !!opts[:bypass_document_validation]
           cmd[:writeConcern] = write_concern.options if write_concern
+          cmd[:arrayFilters] = opts[:array_filters] if opts[:array_filters]
```
estolfo: Don't we need to check if the server supports array filters in the findAndModify methods?

lib/mongo/collection/view/writable.rb
```diff
@@ -182,6 +185,8 @@ def replace_one(replacement, opts = {})
         # @option opts [ true, false ] :upsert Whether to upsert if the
         #   document doesn't exist.
         # @option opts [ Hash ] :collation The collation to use.
+        # @option options [ Array ] :array_filters A set of filters specifying to which array elements
```
estolfo: This should be ``opts``

lib/mongo/collection/view/writable.rb
```diff
@@ -197,6 +202,8 @@ def update_many(spec, opts = {})
         #
         # @param [ Hash ] spec The update statement.
         # @param [ Hash ] opts The options.
+        # @option options [ Array ] :array_filters A set of filters specifying to which array elements
```
estolfo: This should be put below with the other options and should be ``opts``

lib/mongo/collection/view/writable.rb
```diff
@@ -234,13 +241,14 @@ def update(spec, multi, opts)
...
             apply_collation!(update_doc, server, opts)
+            apply_array_filters!(update_doc, server, opts)
 
             Operation::Write::Update.new(
               :update => update_doc,
               :db_name => collection.database.name,
               :coll_name => collection.name,
               :write_concern => collection.write_concern,
-              :bypass_document_validation => !!opts[:bypass_document_validation]
+              :bypass_document_validation => !!opts[:bypass_document_validation],
```
estolfo: I don't think we need the comma here

lib/mongo/error/unsupported_array_filters.rb
```diff
@@ -0,0 +1,51 @@
...
+    # @since 2.5.0
+    class UnsupportedArrayFilters < Error
+
+      # The default error message describing that array filters are not supported.
+      #
+      # @return [ String ] A default message describing that array filters are not supported by the server.
+      #
+      # @since 2.5.0
+      DEFAULT_MESSAGE = "The array_filters option is not a supported feature of the server handling this operation. " +
+          "Operation results may be unexpected."
```
estolfo: We should freeze this string.

lib/mongo/error/unsupported_array_filters.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      # The error message describing that array filters cannot be used when write concern is unacknowledged.
+      #
+      # @return [ String ] A message describing that array filters cannot be used when write concern is unacknowledged.
+      #
+      # @since 2.5.0
+      UNACKNOWLEDGED_WRITES_MESSAGE = "The array_filters option cannot be specified when using unacknowledged writes. " +
+        "Either remove the array_filters option or use acknowledged writes (w >= 1)."
```
estolfo: We should freeze this string.

lib/mongo/error/unsupported_array_filters.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      # Create the new exception.
+      #
+      # @example Create the new exception.
+      #   Mongo::Error::UnsupportedCollation.new
```
estolfo: This should be ``UnsupportedArrayFilters``

lib/mongo/operation/write/update.rb
```diff
@@ -60,9 +61,16 @@ class Update
         def write_command_op
           s = spec.merge(:updates => [ update ])
           s.delete(:update)
+
+          puts s
```
estolfo: You can take this out.

lib/mongo/collection/view.rb
```diff
@@ -183,13 +185,26 @@ def new(options)
         View.new(collection, filter, options)
       end
 
+      def apply_array_filters!(doc, server, opts = {})
```
estolfo: We can probably add this method and the ``validate_array_filters!`` method onto the Writable module instead, as they're only ever used in that module.

lib/mongo/error/unsupported_array_filters.rb
```diff
@@ -0,0 +1,51 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module Mongo
+  class Error
+
+    # Raised if the array filters option is specified for an operation but the server
+    # selected does not support collations.
```
estolfo: 'does not support array filters'

lib/mongo/operation/write/update.rb
```diff
@@ -60,9 +61,16 @@ class Update
...
+
           Command::Update.new(s)
         end
 
+        def has_array_filters?
+          update[:array_filters] || update[Operation::ARRAY_FILTERS]
```
estolfo: Do we need to catch when the update document has the key ``arrayFilters``?

estolfo:
Hey @saghm 
Let me know when you've finished addressing the comments and I should review again. Thanks!

saghm:
Everything is passing on Evergreen now (mod some of the usual jruby failures), so it's ready for review again

estolfo:
ok thanks!

lib/mongo/operation/write/update.rb
```diff
@@ -63,6 +64,10 @@ def write_command_op
           Command::Update.new(s)
         end
 
+        def has_array_filters?
+          update[:array_filters] || update[:arrayFilters] || update['arrayFilters'] || update[Operation::ARRAY_FILTERS]
```
estolfo: We only need to check for ``update[Operation::ARRAY_FILTERS]`` (after changing the constant to ``arrayFilters`` because that's the only key at which we'll add the option from Collection::View::Writable and (should add in) BulkWrite.

Regarding, bulk writes, we have to add the array_filters here: 
https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/bulk_write/transformable.rb#L120
https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/bulk_write/transformable.rb#L106
https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/bulk_write/transformable.rb#L92

lib/mongo/bulk_write/validatable.rb
```diff
@@ -43,6 +43,10 @@ def validate(name, document)
         if document.respond_to?(:keys) && (document[:collation] || document[Operation::COLLATION])
           @has_collation = true
         end
+
+        if document.respond_to?(:keys) && (document[:array_filters] || document[Operation::ARRAY_FILTERS])
```
estolfo: You only need to check if the document has the key ``:array_filters`` or ``'array_filters'``

Because I've suggested changing ``Operation::ARRAY_FILTERS`` to ``arrayFilters`` below, you shouldn't use this constant.

lib/mongo/collection/view.rb
```diff
@@ -131,6 +131,8 @@ def hash
       # @option options :sort [ Hash ] The key and direction pairs used to sort the
       #   results.
       # @option options [ Hash ] :collation The collation to use.
+      # @option options [ Array ] :array_filters A set of filters specifying to which array elements
+      #   an update should apply.
```
estolfo: You don't need to document ``array_filters`` here, as the methods ``update_one``, ``update_many``, ``find_one_and_replace``, ``find_one_and_update`` pass the options from API method on collection to the corresponding method in writable.

For example, calling ``collection.update_many({..}, {..}, array_filters => [..])``, creates a ``Collection::View``, then immediately calls ``#update_many`` on the view. It passes the options directly to ``Collection::View::Writable``
[see here](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/collection.rb#L513)

Let me know if that's not clear.

lib/mongo/collection/view/writable.rb
```diff
@@ -244,6 +253,19 @@ def update(spec, multi, opts)
             ).execute(server)
           end
         end
+
+        def apply_array_filters!(doc, server, opts = {})
+          if filters = opts[:array_filters] || opts[:arrayFilters] || opts['array_filters'] || opts['arrayFilters']
```
estolfo: You only need to check for ``'array_filters'`` and ``:array_filters``
and we should use a constant somewhere for the string

lib/mongo/operation/write/update.rb
```diff
@@ -71,12 +76,16 @@ def message(server)
           flags = []
           flags << :multi_update if update[Operation::MULTI]
           flags << :upsert if update[Operation::UPSERT]
+
+          options = flags.empty? ? {} : { flags: flags }
+          options[:array_filters] = update[:array_filters] if has_array_filters?
```
estolfo: You don't need to do any of this, as array filters can only be used with server version 3.6, with which only write commands will be used (not the ``Protocol::Update`` message).

lib/mongo/operation/write/update.rb
```diff
@@ -71,12 +76,16 @@ def message(server)
...
+
           Protocol::Update.new(
             db_name,
             coll_name,
             update[Operation::Q],
             update[Operation::U],
-            flags.empty? ? {} : { flags: flags }
+            options
```
estolfo: this change isn't necessary either.

lib/mongo/protocol/update.rb
```diff
@@ -58,7 +58,7 @@ def initialize(database, collection, selector, update, options = {})
         @selector    = selector
         @update      = update
         @flags       = options[:flags] || []
-        @upconverter = Upconverter.new(collection, selector, update, flags)
+        @upconverter = Upconverter.new(collection, selector, update, flags, options[:array_filters])
```
estolfo: None of the changes in this file are necessary, as array filters will only be used with write commands.

lib/mongo/collection/view/writable.rb
saghm: I put the constant at the top of this module; let me know if there's somewhere else that would be better suited for it.

estolfo: that's fine, thanks!

lib/mongo/bulk_write/transformable.rb
```diff
@@ -90,6 +90,7 @@ module Transformable
           Operation::UPSERT => doc.fetch(:upsert, false)
         }.tap do |d|
           d[Operation::COLLATION] = doc[:collation] if doc[:collation]
+          d[Operation::ARRAY_FILTERS] = doc[:array_filters] if doc[:array_filters]
```
estolfo: Sorry, this was my mistake to tell you to merge in array_filters here. Array filters aren't relevant to a replace_one, right? If not, you don't need this line. That was my error, sorry!

lib/mongo/collection.rb
```diff
@@ -483,6 +483,8 @@ def parallel_scan(cursor_count, options = {})
     # @option options [ true, false ] :bypass_document_validation Whether or
     #   not to skip document level validation.
     # @option options [ Hash ] :collation The collation to use.
+    # @option options [ Array ] :array_filters A set of filters specifying to which array elements
+    #   an update should apply.
```
estolfo: Sorry again, I think we don't need this here, as array filters aren't relevant to a replace_one. We do, however need the document on the update_one method on line 533.

spec/mongo/collection_spec.rb
```diff
@@ -3487,5 +3487,213 @@ def generate
         expect(result).to be_nil
       end
     end
+
+    context 'when arrayFilters is provided' do
```
estolfo: Would you mind moving these tests to be under their respective methods?

So for #update_one, you'd put the block under the context for the method here:
https://github.com/saghm/mongo-ruby-driver/blob/304438edc9680f214b8d67ffae2951c50b602035/spec/mongo/collection_spec.rb#L2441

for #update_many, you'd put it in this context:
https://github.com/saghm/mongo-ruby-driver/blob/304438edc9680f214b8d67ffae2951c50b602035/spec/mongo/collection_spec.rb#L2169

and #find_one_and_update:
https://github.com/saghm/mongo-ruby-driver/blob/304438edc9680f214b8d67ffae2951c50b602035/spec/mongo/collection_spec.rb#L2900

PR #887, opened 2017-09-05
RUBY-1236 Parse localLogicalSessionTimeoutMinutes from isMaster response

estolfo:


estolfo:
Hi @ajdavis 
Would you mind looking this over? Thank you

lib/mongo/cluster.rb
```diff
@@ -442,6 +442,21 @@ def addresses
...
+    #
+    # @example Get the logical session timeout in minutes.
+    #   cluster.logical_session_timeout
+    #
+    # @return [ Integer, nil ] The logical session timeout.
+    #
+    # @since 2.5.0
+    def logical_session_timeout
+      servers.inject(nil) do |min, server|
+        break unless timeout = server.logical_session_timeout
```
ajdavis: This returns nil if any server's timeout is nil?

estolfo: Yes. The ``break`` ends the iteration returning nil from the block.

ajdavis: Great, looks good to me.

estolfo: thanks

PR #888, opened 2017-09-07
RUBY-1228 Change Streams

estolfo:


saghm:
I looked over this, and unfortunately I don't really have any useful feedback to provide. Everything seems fine to me.

estolfo:
ok, thank you @saghm !

lib/mongo/cursor.rb
```diff
@@ -185,8 +190,12 @@ def exhausted?
     end
 
     def get_more
-      read_with_retry do
+      if @options[:disable_retry]
         process(get_more_operation.execute(@server))
+      else
+        read_with_retry do
```
ShaneHarvey: I don't think getmores should be retryed with the same cursor.  The spec says to create a new cursor on a retryable error.

lib/mongo/collection/view/change_stream.rb
```diff
@@ -0,0 +1,138 @@
...
+        #   stream.each do |document|
+        #     p document
+        #   end
+        #
+        # @return [ Enumerator ] The enumerator.
+        #
+        # @since 2.5.0
+        #
+        # @yieldparam [ BSON::Document ] Each change stream document.
+        def each
```
ShaneHarvey: I think there might be a few spec violations in this method. First, the spec says:
> A driver SHOULD attempt to kill the cursor on the server on which the cursor is opened during the resume process, and MUST NOT attempt to kill the cursor on any other server.

Does `close` ensure that the kill cursor request goes to the original server?

Second, `create_cursor` retrys once and this method retrys as well. That means it will retry on consecutive network errors. The spec says to only retry on the first network error. Any errors that happen during the retry should be fatal. I think instead of `(@cursor || create_cursor!)` it should be something like `(@cursor || create_cursor_no_retry!)`.

lib/mongo/collection/view/change_stream.rb
```diff
@@ -0,0 +1,138 @@
...
+        # @option options [ Integer ] :batch_size The number of documents to return per batch.
+        # @option options [ BSON::Document, Hash ] :collation The collation to use.
+        #
+        # @since 2.5.0
+        def initialize(view, pipeline, options = {})
+          @view = view
+          @change_stream_filters = pipeline
+          @options = options.freeze
+          @resume_token = @options[:resume_after]
+          create_cursor!
```
ShaneHarvey: I thought I remember reading that the initial aggregation command should not be retryed but I don't see it in the spec. I'll open a spec question.

ShaneHarvey: Waiting for an answer: https://jira.mongodb.org/browse/SPEC-947

estolfo: Thanks for pointing this out. I chose to retry it because we historically retry read operations in the driver. 

([here](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/collection/view/iterable.rb#L39) is where we retry finds, aggregations, and [here](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/collection/view/readable.rb#L138) is where we retry a count, for example)

and I thought it would be inconsistent to not retry the initial aggregation for a change stream. I'll see how the discussion goes in the spec ticket, thanks for opening it.

lib/mongo/collection/view/change_stream.rb
estolfo: > Does close ensure that the kill cursor request goes to the original server?

Yes. A cursor has a reference to the server (``@server`` variable). The kill cursors method called on the cursor in line 105, [sends the operation to that server](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/cursor.rb#L201).

> Second, create_cursor retrys once and this method retrys as well. That means it will retry on consecutive network errors. The spec says to only retry on the first network error. Any errors that happen during the retry should be fatal. I think instead of (@cursor || create_cursor!) it should be something like (@cursor || create_cursor_no_retry!). Do you agree or am I misunderstanding the code?

Yes, you're right. I shouldn't be retrying in the create_cursor method. I altered the logic [a few commits ago](https://github.com/mongodb/mongo-ruby-driver/pull/888/commits/5cb322d8bf2f6a26a217aadc5ea7b98457007bf1). I believe it was correct before so I'll change it back.

lib/mongo/cursor.rb
ShaneHarvey: Thanks for the explanation.

lib/mongo/collection/view/change_stream.rb
ShaneHarvey: Thanks, this makes a lot more sense now that I know Ruby retries read operations already.

lib/mongo/collection/view/change_stream.rb
```diff
@@ -0,0 +1,138 @@
...
+        #   new change stream.
+        # @option options [ Integer ] :max_await_time_ms The maximum amount of time for the server to wait
+        #   on new documents to satisfy a change stream query.
+        # @option options [ Integer ] :batch_size The number of documents to return per batch.
+        # @option options [ BSON::Document, Hash ] :collation The collation to use.
+        #
+        # @since 2.5.0
+        def initialize(view, pipeline, options = {})
+          @view = view
+          @change_stream_filters = pipeline
```
ShaneHarvey: Do you need to make a deep copy of the pipeline to ensure the user does not modify it after calling watch?
```ruby
pipeline = []
change_stream = coll.watch(pipeline, {})
pipeline << {'$project': {_id: 0}}
# Will the next cursor be created with [{'$changeStream': ...}, {'$project': {_id: 0}}]?
```
Maybe I'm being too paranoid...

estolfo: haha, that is possible, I guess. Good call though, it wouldn't hurt to copy it.

lib/mongo/collection/view/change_stream.rb
```diff
@@ -0,0 +1,143 @@
...
+        end
+
+        def closed?
+          !!@cursor
+        end
+
+        private
+
+        def cache_resume_token(doc)
+          unless @resume_token = doc[:_id]
```
ShaneHarvey: Similarly, do you need to duplicate the _id to prevent a user from modifying it?

estolfo: No, because ObjectIds are immutable and we are caching the ObjectId itself, not the document. If we were caching the document, then we might have to duplicate it. Thanks for checking though.

ShaneHarvey: The `_id` here is a change stream resume token which is a document.

estolfo: Oh right, I forgot. hm, I guess I should duplicate it then. Thanks!

PR #892, opened 2017-11-17
Update change stream example comments

saghm:


kay-kim:
@saghm  -- just wondering if this can be merged

saghm:
Oops, sorry this fell off my radar! It looks like a couple things failed in Evergreen; although I'm pretty sure they're unrelated, I think I should take a closer look and figure out what's going on.

Worst case scenario, Emily is back next week and can help me debug any issues that are still going on.

saghm:
Given that these changes are so trivial, we can just merge this, as it would take me longer than we have until Emily comes back to get everything on the Evergreen build green again.

estolfo:
thanks @saghm 
I'm back and can take care of getting Evergreen builds green. Thanks = )

PR #894, opened 2017-12-04
Perform a cluster.scan! on OperationFailure in read_with_retry

mdehoog:
We noticed that performing failovers in production leaves lots of persistent connections to the secondary, causing `Mongo::Error::OperationFailure: not master and slaveOk=false` errors. `Retryable.read_with_retry` performs a `cluster.scan!` both `Error::SocketError, Error::SocketTimeoutError`, but not on an `Error::OperationFailure`. Is there a particular reason for this?

We monkey-patched this in production and all our failover issues went away (after making this change too: https://github.com/mongodb/mongoid/pull/4456). 

Related: https://jira.mongodb.org/browse/MONGOID-4305.
Same as https://github.com/reidmorrison/mongo_ha/blob/master/lib/mongo_ha/retryable.rb#L17-L20.

### DIFF
--- a/lib/mongo/retryable.rb
+++ b/lib/mongo/retryable.rb
@@ -48,14 +48,10 @@ module Mongo
         cluster.scan!
         retry
       rescue Error::OperationFailure => e
-        if cluster.sharded? && e.retryable?
-          raise(e) if attempt > cluster.max_read_retries
-          log_retry(e)
-          sleep(cluster.read_retry_interval)
-          retry
-        else
-          raise e
-        end
+        raise(e) if !e.retryable? || (attempt > cluster.max_read_retries)
+        log_retry(e)
+        cluster.sharded? ? sleep(cluster.read_retry_interval) : cluster.scan!
+        retry
       end
     end
 
--- a/spec/mongo/retryable_spec.rb
+++ b/spec/mongo/retryable_spec.rb
@@ -95,45 +95,30 @@ describe Mongo::Retryable do
 
     context 'when an operation failure occurs' do
 
-      context 'when the cluster is not a mongos' do
+      context 'when the operation failure is not retryable' do
 
-        before do
-          expect(operation).to receive(:execute).and_raise(Mongo::Error::OperationFailure).ordered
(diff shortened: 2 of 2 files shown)


estolfo:
Hi @mdehoog 

Can you elaborate on what you mean when you say "performing failovers"? Do you mean you are intentionally stepping-down a primary in production?

The ``#cluster.scan!`` method connects to all servers and refreshes the state of each. This is necessary and useful when there is a Socket error indicating that there is a network or connectivity issue. It's an operation that has the potential to be very slow, as it connects to each server synchronously.

An OperationFailure indicates that an operation reached the server but failed. It's unnecessary to refresh all servers' state when an OperationFailure is encountered, as the failure is not an indication of a network or connectivity issue. As mentioned above, refreshing all servers' state has the potential to be very slow so it would be a big performance hit to run ``#cluster.scan!`` upon every OperationFailure.

Nevertheless, if you are finding that the driver is taking a while to recognize that there has been a failover, you can adjust the ``heartbeat_frequency`` client parameter. This parameter affects how frequency the driver refreshes its view of each server state. See the documentation [here](https://docs.mongodb.com/ruby-driver/master/tutorials/ruby-driver-create-client/#ruby-options) for other timeouts and options you can use for your driver client.

I would recommend adjusting the client options to see if that has an effect on how your application behaves during a failover. Otherwise, you might want to consider catching and handling the errors in your application code, especially if you are forcing the failover yourself and can predict when they'll happen.
Thanks
Emily

mdehoog:
We use mLab's Mongo hosting, and we sometimes failover (stepping-down the primary) as required for maintenance (e.g. vertically scaling a Mongo box). When failing-over, we don't receive a `SocketError`, but rather an `OperationFailure` is raised. We haven't seen `OperationFailure`s anywhere else apart from these failovers. It could possibly be mLab specific, but this change has certainly helped fixing our issues.

estolfo:
I'm going to close this as I would encourage users to adjust their client settings instead of calling #cluster.scan! on read failures. We are also working on a retryable reads specification so you can expect this behavior to change soon anyway.

PR #895, opened 2017-12-07
RUBY-1230 Causally consistent reads

estolfo:


lib/mongo/session.rb
```diff
@@ -182,8 +187,38 @@ def advance_cluster_time(new_cluster_time)
...
+    def causal_consistency_doc(read_concern)
+      if operation_time && causal_consistency?
+        (read_concern || {}).merge(:afterClusterTime => operation_time)
+      else
+        read_concern
+      end
+    end
+
+    def causal_consistency?
+      @causal_consistency ||= (!@options.key?(:causal_consistency) || @options[:causal_consistency])
```
saghm: I'm not sure this line is doing what it's supposed to; if `:causal_consistency` isn't in `@options` (i.e. `@options.key?(:causal_consistency)` is false), this would set `@causal_consistency` to true. If I'm understanding the intended logic correctly, I think this should be `@options.key?(:causal_consistency) && @options[:causal_consistency]`.

estolfo: If the option is not specified, it should be true. You can see the first paragraph [here](https://github.com/mongodb/specifications/blob/master/source/causal-consistency/causal-consistency.rst#causalconsistency)

So:
options = {} # causal_consistency is true
options = { :causal_consistency => true } # causal_consistency is true
options = { :causal_consistency => false } # causal_consistency is false
options = { :causal_consistency => nil } # causal_consistency is false

And I'm realizing that there's a bug with my code for the last scenario, so I'll fix it.

saghm: Hmm, yeah, it's a little unfortunate that the default is true here, since it seems weird to have a lack of a `:causal_consistency` key being different than a nil value, but having `nil` and `false` be different doesn't seem much better. I guess for the sake of the users this will have to be documented carefully.

PR #896, opened 2017-12-12
Update change stream examples test

estolfo:
These tests were previously failing frequently on Evergreen because the operations launched in the thread didn't seem to always execute. I've changed the test to create two threads and force the main thread to wait until each completes in succession.


estolfo:
no worries, it's Ruby's fault = )

PR #897, opened 2017-12-13
RUBY-1229 retry writes

estolfo:


lib/mongo/retryable.rb
```diff
@@ -98,12 +98,58 @@ def read_with_one_retry
...
+
+    def legacy_write_with_retry(server = nil)
       attempt = 0
       begin
         attempt += 1
-        yield(server_selector.call)
+        yield(server || cluster.next_primary)
       rescue Error::OperationFailure => e
+        server = nil
         raise(e) if attempt > Cluster::MAX_WRITE_RETRIES
```
saghm: If I understand the purpose of this line correctly, it's to abstract away the exact value of MAX_WRITE_RETRIES instead of assuming some value (e.g. 1). If that's the case, should this be in a loop? If MAX_WRITE_RETRIES is greater than 1, this function will still only retry once.

estolfo: Can you tell me why you think it'll only retry once? If {{attempt}} is less than {{Cluster::MAX_WRITE_RETRIES}} then the error will not be raised. Then, if the error is not {{write_retryable?}}, execution will continue until line 158, when the block will be retried.
Keep in mind that this is the old behavior before we wrote the retryable writes spec. It's still used for methods like update_many and delete_many and writes to servers < 3.6.)

saghm: Ah, the Github diff viewer had hidden the part with the retry, and I didn't notice that the were lines missing. Sorry about that!

estolfo: oh, that explains it! No worries

estolfo:
thank you @saghm !

PR #898, opened 2017-12-18
RUBY-1245 Support SRV protocol for connection strings

estolfo:


lib/mongo/uri.rb
```diff
@@ -46,6 +46,39 @@ class URI
...
+
+    # Error details for an invalid scheme.
+    #
+    # @since 2.1.0
+    INVALID_SCHEME = "Invalid scheme. Scheme must be '#{MONGODB_SCHEME}' or '#{MONGODB_SRV_SCHEME}'".freeze
+
+    # MongoDB URI format specification.
+    #
+    # @since 2.0.0
+    FORMAT = 'mongodb://[username:password@]host1[:port1][,host2[:port2]' +
```
saghm: Does the scheme need to reflect the possible `+srv` here?

estolfo: yes it should, thank you.

estolfo: I've updated the SRVProtocol class to create its own InvalidURI errors so it can pass in its own format info.

lib/mongo/uri/srv_protocol.rb
```diff
@@ -78,12 +75,18 @@ def client_options
                                     "[#{VALID_TXT_OPTIONS.join(', ')}].".freeze
 
       MISTMATCHED_DOMAINNAME = "Parent domain name in SRV record result (%s) does not match " +
-                                 "hostname (%s)".freeze
+                                 "that of the hostname (%s)".freeze
+
+      FORMAT = 'mongodb+srv://[username:password@]host[/[database][?options]]'.freeze
```
saghm: I think it should begin with `mongodb[+srv]` since the "+srv" is optional

estolfo: But for the SRV protocol, the +srv is required. There is this FORMAT variable for the mongodb scheme in the uri class: https://github.com/mongodb/mongo-ruby-driver/pull/898/files#diff-e7786d37b967fce59a4cd10176f3a85aR74
These are used in the error message for an InvalidURI error; at the time the error is instantiated, I know whether I need to pass in the FORMAT of the SRV protocol or the mongodb scheme.

PR #909, opened 2018-01-24
RUBY-1278 Resync connection string spec tests for percent encoding in userinfo

estolfo:


estolfo:
Perhaps superseded by later commits to the specifications repo. Waiting on more info before adding this change.

PR #925, opened 2018-04-13
RUBY-1297 Update server selection tests for read preferences with sha…

saghm:
…rded clusters

### DIFF
--- a/spec/support/server_selection.rb
+++ b/spec/support/server_selection.rb
@@ -149,7 +149,7 @@ module Mongo
         #
         # @since 2.0.0
         def candidate_servers
-          @candidate_servers.select { |s| s['type'] != 'Unknown' }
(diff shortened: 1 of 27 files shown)


spec/support/server_selection.rb
```diff
@@ -149,7 +149,7 @@ def in_latency_window
         #
         # @since 2.0.0
         def candidate_servers
-          @candidate_servers.select { |s| s['type'] != 'Unknown' }
+          @candidate_servers.select { |s| s['type'] != 'Unknown' && s['type'] != 'PossiblePrimary' }
```
estolfo: This is really minor but in terms of style, it might be easier to write it like this:

```ruby
@candidate_servers.select { |s| !['Unknown', 'PossiblePrimary'].include?(s['type'] ) }
```

That way, if we ever need to add a certain server type to exclude, we can just add it to the array.

saghm: Done

PR #932, opened 2018-05-29
Add an example for connecting directly to a replica set member

p-mongo:


p-mongo:
@saghm ready for review

PR #934, opened 2018-05-30
Scram sha 256 dev

saghm:
Not ready yet for review; posted to make it easier for @p-mongo to avoid merge conflicts


.evergreen/.evg.yml
```diff
@@ -320,7 +320,7 @@ functions:
         working_dir: "src"
         script: |
           ${PREPARE_SHELL}
-          AUTH=${AUTH} SSL=${SSL} MONGODB_URI="${MONGODB_URI}" TOPOLOGY=${TOPOLOGY} RVM_RUBY="${RVM_RUBY}" .evergreen/run-tests.sh
+          AUTH=${AUTH} SSL=${SSL} MONGODB_URI="${MONGODB_URI}" TOPOLOGY=${TOPOLOGY} RVM_RUBY="${RVM_RUBY}" MONGODB_VERSION=${VERSION} .evergreen/run-tests.sh
```
saghm: The reason we're setting the version as an environmental variable is due to a quirk in the way that saslprep for SCRAM-SHA-256; because the method `String#unicode_normalize!` wasn't added until Ruby 2.2, we would need to add a dependency to be able to salsprep in older Ruby versions. Rather than add a dependecy for old versions of Ruby to be able to authenticate with SCRAM-SHA-256 in the newest version of the database when SCRAM-SHA-1 still works fine, we decided that it was better to just raise an informative error and wait and see if anyone asks for the feature to be added. Because we create a number of clients in the tests, we need to specify that they use SCRAM-SHA-1 when the Ruby version is below 2.2 and the server version is at least 4.0, so we add the MongoDB server version the tests are running against in order to be able to specify this.

lib/mongo/auth/scram256.rb
```diff
@@ -0,0 +1,70 @@
+# Copyright (C) 2014 MongoDB Inc.
```
saghm: This file was copied from lib/mongo/auth/scram.rb and modified as necessary.

lib/mongo/server/connection.rb
```diff
@@ -237,14 +237,26 @@ def handshake!
           min_wire_version = response[Description::MIN_WIRE_VERSION] || Description::LEGACY_WIRE_VERSION
           max_wire_version = response[Description::MAX_WIRE_VERSION] || Description::LEGACY_WIRE_VERSION
           features = Description::Features.new(min_wire_version..max_wire_version)
-          @auth_mechanism = (features.scram_sha_1_enabled? || @server.features.scram_sha_1_enabled?) ? :scram : :mongodb_cr
+
+          if response["ok"] == 1
+            @auth_mechanism = if response['saslSupportedMechs']
```
saghm: I'm not super fond of the way this logic is expressed and would welcome feedback on it. The basic algorithm is "use SCRAM-SHA-256 if the server responds with it in the list of supported SASL mechanisms, use SCRAM-SHA-1 if the server returns a list of mechanisms without SCRAM-SHA-256 or if no mechanisms are returned by the server but SCRAM-SHA-1 is enabled, and use MONGODB-CR otherwise".

spec/mongo/auth/scram/conversation_spec.rb
```diff
@@ -1,12 +1,13 @@
...
-describe Mongo::Auth::SCRAM::Conversation do
+describe Mongo::Auth::SCRAM::Conversation, if: scram_sha_1_enabled? do
 
   let(:user) do
     Mongo::Auth::User.new(
       database: Mongo::Database::ADMIN,
       user: 'user',
-      password: 'pencil'
+      password: 'pencil',
+      auth_mech: 'SCRAM-SHA-1'
```
saghm: SCRAM-SHA-1 is explicitly set here and elsewhere that we're explicitly testing it so we don't accidentally use SCRAM-SHA-256 with newer server versions.

spec/mongo/auth/scram256/conversation_spec.rb
```diff
@@ -0,0 +1,198 @@
+require 'spec_helper'
```
saghm: This was copied from spec/mongo/auth/scram/conversation_spec.rb and modified as necessary.

spec/mongo/auth/scram256_spec.rb
```diff
@@ -0,0 +1,99 @@
+require 'spec_helper'
```
saghm: This was copied from spec/mongo/auth/scram_spec.rb and modified as necessary.

spec/mongo/auth/scram_spec.rb
```diff
@@ -43,15 +43,16 @@
...
+          auth_mech: 'SCRAM-SHA-1'
         )
       end
 
       let(:cr) do
         described_class.new(user)
       end
 
-      it 'raises an exception', if: list_command_enabled? do
+      it 'raises an exception', if: scram_sha_1_enabled? do
```
saghm: `scram_sha_1_enabled?` is defined as an alias to `list_command_enabled?` (because they were added in the same server version), so it seemed reasonable to just use it explicitly here to avoid confusion for future readers .

spec/mongo/cluster/app_metadata_spec.rb
```diff
@@ -112,7 +112,17 @@
         end
 
         it 'truncates the document to be just an ismaster command and the compressors', unless: compression_enabled? do
-          expect(app_metadata.ismaster_bytes.length).to eq(Mongo::Server::Monitor::Connection::ISMASTER_BYTES.length + 26)
+          sasl_supported_mechs_size = 0
```
saghm: Given that we sometimes query the server for the supported authentication mechanisms, we need to make sure that when we assert the length of the ismaster document that we include that info.

lib/mongo/auth/scram256/conversation.rb
```diff
@@ -0,0 +1,509 @@
+# Copyright (C) 2014 MongoDB Inc.
```
saghm: This was copied from lib/mongo/auth/scram/conversation.rb and modified as necessary.

spec/mongo/auth/scram256/negotiation_spec.rb
```diff
@@ -0,0 +1,562 @@
+require 'spec_helper'
+require 'cgi'
+
+module Mongo
+  class Server
```
saghm: This is monkey-patched to facilitate testing that SCRAM-SHA-256 is used when both SCRAM-SHA-1 and SCRAM-SHA-256 are available for a given user.

saghm:
Ready for review

.evergreen/.evg.yml
p-mongo: Can you please add this and some of the other comments to the code so that they are available to whoever might read the source in the future?

spec/mongo/auth/scram/conversation_spec.rb
p-mongo: This one

spec/mongo/auth/scram256/negotiation_spec.rb
p-mongo: This one

spec/mongo/cluster/app_metadata_spec.rb
p-mongo: This one

saghm:
Comments added as requested

lib/mongo/auth/scram256/conversation.rb
```diff
@@ -0,0 +1,509 @@
...
+          raise Error::InvalidNonce.new(nonce, rnonce) unless rnonce.start_with?(nonce)
+        end
+
+        def validate!(reply)
+          raise Unauthorized.new(user) unless reply.documents[0][Operation::Result::OK] == 1
+          @reply = reply
+        end
+
+        def digest
+          @digest ||= DIGEST
```
p-mongo: I am quite certain that this is not correct. The definition of DIGEST in SCRAM isn't correct either but it is not used anywhere (def digest in SCRAM does not use DIGEST constant) hence is effectively harmless.

DIGEST as defined in SCRAM and copied by you to SCRAM256 is a frozen instance of a digest class. Except freeze on these instances doesn't seem to do anything - they can be mutated with `update` even when frozen. And since you *are* using DIGEST here, what this code does is it uses the same digest instance for all digests in the application (original SCRAM does not have this problem as its digest method does not use DIGEST constant).

I have this repaired in https://github.com/p-mongo/mongo-ruby-driver/commit/e1875021d44b9a35a0657076c3d1db2cc3c59226 as part of DRYing the code.

lib/mongo/auth/scram256/conversation.rb
```diff
@@ -0,0 +1,509 @@
...
+        end
+
+        # Get the iterations from the server response.
+        #
+        # @api private
+        #
+        # @since 2.6.0
+        def iterations
+          @iterations ||= payload_data.match(ITERATIONS)[1].to_i.tap do |i|
+            next unless i < MIN_ITER_COUNT
```
p-mongo: This is a very strange construct, can the block body be rephrased to `if i < MIN_ITER_COUNT then raise ...` ?

saghm:
I'll take a look at the branch on your fork. In the future, let's not do parallel development as part of a code review, as it can get messy when multiple people are working on the same changes at the same time. If necessary, we can discuss offline about how to coordinate if for some reason the person assigned the work isn't able to do part of it.

lib/mongo/auth/scram256/conversation.rb
saghm: I think we should hold off on refactoring the conversation algorithm for another time (and more broadly, not bring extra scope by reviewing existing code that works fine when doing a review). I understand that being newly introduced to the codebase, you're going to notice a lot of small things that look like they can be improved, and in a lot of places, that's probably correct. However, most of these things can a) be addressed in a separate ticket/commit and b) are not a huge priority while we're driving towards getting a release in time to support 4.0. There's certainly a lot of work to be done in cleaning up the code in a lot of places, but it's important that we don't get sidetracked by issues in existing code that is working properly. In the future, I'll try to do a better job of marking which code in a pull request is not new so that we can focus on the new stuff being added.

lib/mongo/server/connection.rb
behackett: I don't see where saslSupportedMechs: "<dbname>.<username>" is being added to the ismaster command. Mechanism negotiation won't work without that.

https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#mechanism-negotiation-via-handshake

behackett: NM, I found it.

lib/mongo/auth/scram256/conversation.rb
saghm: Oops, I completely misread this; this is clearly not existing code, so you can ignore what I said. Sorry about that!

saghm:
The SCRAM-SHA-1 and SCRAM-SHA-256 conversations are merged in the latest commit

lib/mongo/auth/scram.rb
```diff
@@ -49,12 +63,14 @@ def initialize(user)
...
       #   on.
+      # @param [ String ] mechanism The auth mechanism to use (either 'SCRAM-SHA-1' or
+      #   'SCRAM-SHA-256');
       #
       # @return [ Protocol::Message ] The authentication response.
       #
       # @since 2.0.0
-      def login(connection)
-        conversation = Conversation.new(user)
+      def login(connection, mechanism = user.mechanism || :scram)
```
p-mongo: It is better to default mechanism to nil here and apply the default in method body. That way downstream code can pass in a mechanism of nil and get the default applied.

lib/mongo/auth/scram/conversation.rb
```diff
@@ -60,6 +55,11 @@ class Conversation
         # @since 2.0.0
         ITERATIONS = /i=(\d+)/.freeze
 
+        # The minimum iteration count for SCRAM-SHA-256.
+        #
+        # @since 2.6.0
```
p-mongo: Mark this `@api private` ?

lib/mongo/cluster/app_metadata.rb
```diff
@@ -55,6 +55,11 @@ def initialize(cluster)
         @app_name = cluster.options[:app_name].to_s if cluster.options[:app_name]
         @platform = cluster.options[:platform]
         @compressors = cluster.options[:compressors] || []
+
+        if cluster.options[:user] && !cluster.options[:auth_mech]
+          auth_db = cluster.options[:auth_source] || 'admin'
```
p-mongo: Isn't default auth database supposed to be the same as connection database?

spec/mongo/auth/scram/negotiation_spec.rb
```diff
@@ -0,0 +1,568 @@
...
+        end
+
+
+        context 'when SCRAM-SHA-256 is specified as the auth mechanism' do
+
+          let(:auth_mech) do
+            :scram256
+          end
+
+          it 'authenticates successfully' do
```
p-mongo: This label does not match the test contents.

spec/spec_helper.rb
```diff
@@ -248,6 +257,18 @@ def auth_enabled?
   end
 end
 
+def need_to_skip_on_sharded_auth_40?
+  sharded? && auth_enabled? && scram_sha_256_enabled?
+end
+
+# Can the driver specify a write concern that won't be overridden? (mongos 4.0+ overrides the write)
```
p-mongo: Closing paren seems to be in the wrong place.

lib/mongo/cluster/app_metadata.rb
saghm: The authentication options for a connection will be the same as the cluster that its server belongs to, which is what's being checked here before applying the default.

saghm:
All current feedback addressed either in new commit or in inline comments

lib/mongo/auth/scram.rb
```diff
@@ -22,10 +22,24 @@ module Auth
     # @since 2.0.0
     class SCRAM
 
-      # The authentication mechinism string.
+      # The authentication mechanism string for SCRAM-SHA-1.
```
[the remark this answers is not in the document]
saghm: This comment specifically refers to the constant 'SCRAM-SHA-1'; I assume you mean that I should change the class-level comment, which I'll do.

lib/mongo/auth/scram/conversation.rb
```diff
@@ -60,6 +55,13 @@ class Conversation
         # @since 2.0.0
         ITERATIONS = /i=(\d+)/.freeze
 
+        # The minimum iteration count for SCRAM-SHA-256.
+        #
+        # @api private
+        #
+        # @since 2.6.0
+        MIN_SCRAM_SHA_256_ITER_COUNT = 4096
```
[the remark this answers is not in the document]
saghm: The code currently doesn't enforce the minimum for SCRAM-SHA-1; should I change this?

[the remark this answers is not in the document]
behackett: Yes, enforce it for both.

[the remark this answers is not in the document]
saghm: Will do. Out of curiosity, will this break things with old server versions?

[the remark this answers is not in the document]
behackett: Unlikely. The old default was 10000. It's unlikely anyone would lower it. Regardless, this improves security.

saghm:
@behackett @p-mongo Is it all right if I rebase against master and push again? There are some conflicts, but I don't want to change history if there are any outstanding issues

PR #938, opened 2018-06-01
Disable fail fast when not in CI

p-mongo:
Instead run all tests and show all errors.


p-mongo:
@saghm Okay, removed.

PR #939, opened 2018-06-01
Expose error code and code name for errors and write concern errors

p-mongo:
https://jira.mongodb.org/browse/RUBY-1331

Extracted from work for https://jira.mongodb.org/browse/RUBY-1333

### DIFF
--- a/lib/mongo/error/operation_failure.rb
+++ b/lib/mongo/error/operation_failure.rb
@@ -52,7 +52,7 @@ module Mongo
 
       def_delegators :@result, :operation_time
 
-      # @return [ Fixnum ] code The error code parsed from the document.
+      # @return [ Integer ] code The error code parsed from the document.
       # @since 2.6.0
       attr_reader :code
 
@@ -96,7 +96,7 @@ module Mongo
       # @param [ Operation::Result ] result The result object.
       # @param [ Hash ] options Additional parameters
       #
-      # @option options [ Fixnum ] :code Error code
+      # @option options [ Integer ] :code Error code
       # @option options [ String ] :code_name Error code name
       #
       # @since 2.5.0, options added in 2.6.0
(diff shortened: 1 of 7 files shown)


lib/mongo/error/operation_failure.rb
```diff
@@ -52,6 +52,14 @@ class OperationFailure < Error
 
       def_delegators :@result, :operation_time
 
+      # @return [ Fixnum ] code The error code parsed from the document.
```
saghm: Given that `Fixnum` is deprecated in favor of `Integer` in newer Ruby versions and that `#is_a?` will still return true for integers before this change, I think we should just document this as returning `Integer` (here and elsewhere in the change).

lib/mongo/error/operation_failure.rb
```diff
@@ -81,12 +89,21 @@ def write_retryable?
...
+      # @param [ String ] message The error message.
+      # @param [ Operation::Result ] result The result object.
+      # @param [ Hash ] options Additional parameters
+      #
+      # @option options [ Fixnum ] :code Error code
+      # @option options [ String ] :code_name Error code name
       #
-      # @since 2.5.0
-      def initialize(message = nil, result = nil)
+      # @since 2.5.0, options added in 2.6.0
```
saghm: Out of curiosity, is this standard syntax? I'm wondering if it makes more sense to just version the options were added on line 97.

lib/mongo/error/operation_failure.rb
p-mongo: OK.

lib/mongo/error/operation_failure.rb
saghm: That's for Javadocs though; we use [YARD](https://www.rubydoc.info/gems/yard/file/docs/Tags.md#since). I wasn't sure if there were restrictions on `since` tags in the way that it formatted them, but from the link I gave, it doesn't appear so. I'm on the fence about putting the detail about the options on the line where the options are described versus on the `since` tag, so I'll defer to whatever you think is preferable.

PR #940, opened 2018-06-01
RUBY-1332 Sync SDAM spec tests (not monitoring)

p-mongo:
This is part of https://jira.mongodb.org/browse/RUBY-1332 which deals with sdam subdirectory. The only code change needed was raising an error when any server in a cluster has an unacceptable wire protocol, whereas before only the server that was selected was subject to wire protocol check.

Verbose cluster/server inspectors were added for debugging.

### DIFF
--- a/spec/spec_tests/sdam_spec.rb
+++ b/spec/spec_tests/sdam_spec.rb
@@ -94,8 +94,9 @@ describe 'Server Discovery and Monitoring' do
 
             it 'raises an UnsupportedFeatures error' do
               expect {
-                Mongo::ServerSelector.get(mode: :primary).select_server(@client.cluster)
-                Mongo::ServerSelector.get(mode: :secondary).select_server(@client.cluster)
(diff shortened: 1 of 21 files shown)


spec/spec_tests/sdam_spec.rb
```diff
@@ -93,9 +93,21 @@
             end
 
             it 'raises an UnsupportedFeatures error' do
+              # Need to ensure the servers are polled by monitoring thread
+              # prior to performing the following operations
+              deadline = Time.now + 5
```
saghm: In general, I'm not fond of using manual countdowns/sleeps as a workaround for races in tests, as they tend to be rather fragile. Is there a way to achive this without relying on the event occuring within a defined period of time (e.g. by monkey patching the method that does the polling)?

p-mongo: I am not sure what you mean but I think I can take that out.

PR #942, opened 2018-06-04
Reduce travis log size

p-mongo:
This PR reduces Travis log size from ~10,000 lines to ~1,000 lines.

A big change here is replacement of default documentation formatter with Fuubar which provides 1) immediate failure messages for failing tests, 2) progress bar for overall test runtime and 3) compact overall output.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -50,8 +50,6 @@ require 'support/connection_string'
 require 'support/gridfs'
 
 RSpec.configure do |config|
-  config.color     = true
-  config.formatter = 'documentation'
   config.include(Authorization)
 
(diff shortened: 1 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -50,8 +50,10 @@
 require 'support/gridfs'
 
 RSpec.configure do |config|
-  config.color     = true
```
saghm: Why is the color setting removed? Does the color actually increase the log size?

p-mongo: It is specified in .rspec.

saghm: Ah, okay

PR #943, opened 2018-06-05
Spec tests/retryable writes usability changes

p-mongo:
- Move retryable writes spec tests to their own file, so that they can be easily run by themselves
- If these tests are not runnable due to the configuration of running mongo replica set, define the tests and mark them pending rather than not define them, so that the developer knows that these tests exist
- Document requirements for these tests to run (replica set, fail points enabled)
- Add an option to show debug output from Mongo client used in the test suite, so that test suite hangs can be diagnosed
- Set a timeout on server selection to 1 second to fail the test suite in a reasonable amount of time when the running replica set is not usable by the ruby driver (e.g. due to a hostname/IP mismatch in replica set configuration)
- All spec tests moved to `spec/spec_tests`


spec/spec_helper.rb
```diff
@@ -1,3 +1,44 @@
+# The tests run against a mongo server or a replica set which is
```
saghm: This technically makes it sound like sharded clusters aren't supported; it's probably best to just say "The tests run run against a MongoDB cluster..."

p-mongo: Changed.

p-mongo:
@saghm CI is green, added a commit moving all spec tests to spec/spec_tests.

PR #944, opened 2018-06-05
RUBY-1339 Add support for name_only option in Database#list_collections

saghm:


lib/mongo/database.rb
```diff
@@ -122,11 +122,13 @@ def collection_names(options = {})
...
     #
-    # @return [ Array<Hash> ] Info for each collection in the database.
+    # @param [ Boolean ] name_only Whether to return only the names of the collections.
+    #
+    # @return [ Array<Hash | String> ] Info for each collection in the database.
     #
     # @since 2.0.5
-    def list_collections
-      View.new(self).list_collections
+    def list_collections(name_only = false)
```
p-mongo: I would prefer a `list_collection_names` method which invokes `list_collections` and performs the map to go from hash to string. This will avoid having return value type be parameter dependent.

saghm: Makes sense; I'll make the change

saghm: Actually, now that I look at it again, there already is a `collection_names` method; I think we can just close this ticket without needing an changes.

saghm: Never mind, we do still need to do the change where the `nameOnly` field is sent to the server. I'll revise the change to only do that.

p-mongo: Would it be using nameonly though?

saghm:
Updated the change to just add `nameOnly` option when calling `Database::View#collection_names`

lib/mongo/database/view.rb
```diff
@@ -93,29 +93,29 @@ def initialize(database)
 
       private
 
-      def collections_info(server, session, &block)
-        cursor = Cursor.new(self, send_initial_query(server, session), server, session: session)
+      def collections_info(server, session, name_only = false, &block)
```
p-mongo: For future extensibility should this be an options hash?

saghm: Good idea!

PR #946, opened 2018-06-08
Add debug logging for topology events

p-mongo:
This is to assist in understanding SDAM state transitions, ultimately for https://jira.mongodb.org/browse/RUBY-1332

Eventually the various debug events should be organized, and support being turned on and off, by a subsystem as otherwise they would be too noisy.

### DIFF
--- a/lib/mongo/event.rb
+++ b/lib/mongo/event.rb
@@ -36,15 +36,10 @@ module Mongo
     #
     # @since 2.0.6
     DESCRIPTION_CHANGED = 'description_changed'.freeze
-
-    class Base
-      def inspect
-        "#<{#{self.class}>"
-      end
-    end
   end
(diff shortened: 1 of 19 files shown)


lib/mongo/event.rb
```diff
@@ -46,3 +38,12 @@ module Event
     DESCRIPTION_CHANGED = 'description_changed'.freeze
   end
 end
+
```
saghm: Is there a reason for moving the require statements to the bottom (here and below)?

p-mongo: Yes, in general the inner classes should be defined after the outer classes are and once the inner classes start referencing things in the outer classes, or outer classes need to derive from something, the inverted definition order ceases to work.

PR #947, opened 2018-06-09
Lite specs

p-mongo:
- Created a lite spec helper to be used instead of the full spec helper for tests that do not interact with a real mongod. This is meant to accomplish two goals: 1) remove noise from global client init for tests that also deal with client initialization, like topology ones and 2) speed up the tests.
- Changed two more tests to have explicit skips when preconditions are not met.

### DIFF
--- a/spec/spec_tests/command_monitoring_spec.rb
+++ b/spec/spec_tests/command_monitoring_spec.rb
@@ -41,7 +41,13 @@ describe 'Command Monitoring Events' do
 
         test.expectations.each do |expectation|
 
-          it "generates a #{expectation.event_name} for #{expectation.command_name}", unless: ignore?(test) do
+          before do
+            if ignore?(test)
+              skip 'Preconditions not met'
+            end
+          end
+
+          it "generates a #{expectation.event_name} for #{expectation.command_name}" do
             begin
               test.run(authorized_collection)
               event = subscriber.send(expectation.event_type)[expectation.command_name]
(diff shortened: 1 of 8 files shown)


spec/spec_tests/command_monitoring_spec.rb
```diff
@@ -41,7 +41,13 @@ def ignore?(test)
 
         test.expectations.each do |expectation|
 
-          it "generates a #{expectation.event_name} for #{expectation.command_name}", unless: ignore?(test) do
+          before do
```
saghm: What's the advantage of having the test still generated if we don't want it to run?

spec/spec_tests/crud_spec.rb
```diff
@@ -6,16 +6,16 @@
...
+      end
+    end
+
     context(spec.description) do
 
       spec.tests.each do |test|
 
-        around do |example|
-          if spec.server_version_satisfied?(authorized_client)
-            example.run
```
saghm: Shouldn't this line still exist somewhere?

spec/spec_tests/command_monitoring_spec.rb
p-mongo: This way an engineer running the tests knows there are tests which weren't run.

spec/spec_tests/crud_spec.rb
p-mongo: before/after and around hooks work differently in this regard.

PR #949, opened 2018-06-11
Should probably change description before claiming discovery

p-mongo:
For https://jira.mongodb.org/browse/RUBY-1332


p-mongo:
This breaks a lot of stuff so maybe not.

PR #955, opened 2018-06-12
Retryable writes prerequisites

p-mongo:
This PR contains changes that were made as part of retryable writes work but aren't part of the ticket proper, to avoid making the next PR unreadable.

https://jira.mongodb.org/browse/RUBY-1333

### DIFF
--- a/spec/mongo/cluster/topology_spec.rb
+++ b/spec/mongo/cluster/topology_spec.rb
@@ -1,4 +1,4 @@
-require 'spec_helper'
+require 'lite_spec_helper'
 
 describe Mongo::Cluster::Topology do
 
@@ -103,28 +103,6 @@ describe Mongo::Cluster::Topology do
 
         it 'returns an unknown topology' do
           expect(topology).to be_a(Mongo::Cluster::Topology::Unknown)
-        end
-      end
-
-      context 'when provided a single mongos', if: single_mongos? do
-
-        let(:topology) do
-          described_class.initial(ADDRESSES, monitoring, TEST_OPTIONS)
-        end
-
-        it 'returns a sharded topology' do
-          expect(topology).to be_a(Mongo::Cluster::Topology::Sharded)
-        end
-      end
-
-      context 'when provided a single replica set member', if: single_rs_member? do
-
-        let(:topology) do
-          described_class.initial(ADDRESSES, monitoring, TEST_OPTIONS)
-        end
-
-        it 'returns a single topology' do
-          expect(topology).to be_a(Mongo::Cluster::Topology::Single)
         end
       end
     end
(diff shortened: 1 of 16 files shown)


spec/mongo/cluster/topology_spec.rb
```diff
@@ -105,28 +105,6 @@
           expect(topology).to be_a(Mongo::Cluster::Topology::Unknown)
         end
       end
-
```
saghm: What's the rationale for removing these tests?

spec/mongo/retryable_writes_spec.rb
```diff
@@ -1,754 +1,751 @@
 require 'spec_helper'
 
-describe 'Retryable Writes' do
```
saghm: Wow, Github's calculated diff of this file is _terrible_...

spec/spec_helper.rb
```diff
@@ -25,22 +39,40 @@
 # CLIENT_DEBUG: Show debug messages from the client.
 #   CLIENT_DEBUG=1
 #
-# MONGODB_URI: Connection string to use instead of 127.0.0.1:27017.
-# Specify RS_ENABLED or SHARDED_ENABLED if connecting to a replica set
-# or a sharded cluster, otherwise the test suite will establish a
-# direct connection to the first host in the URI.
-#   MONGODB_URI=127.0.0.1:27001,127.0.0.1:27002
+# MONGODB_URI: Connection string to use. This must be a valid MongoDB URI;
```
saghm: I have to wonder if it's worth reducing the number of these variables at some point in the future; it seems to me that `MONGODB_URI` and `TOPOLOGY` should be sufficient. It's out of scope for this ticket though, so we can revisit another time.

spec/spec_tests/retryable_writes_spec.rb
```diff
@@ -19,7 +19,7 @@
           let(:client) do
             authorized_client_with_retry_writes
           end
-
```
saghm: I've noticed a lot of places with small formatting issues like this. After the 2.6.0 release, we should discuss standardizing on some code formatting tool.

spec/mongo/cluster/topology_spec.rb
p-mongo: As mentioned in the commit these tests are duplicated.

spec/spec_helper.rb
p-mongo: I intend to redo test suite configuration at a future time. There are several other considerations that need to be addressed, for example some tests connect to sockets directly which can't be expressed with either MONGODB_URI or MONGODB_ADDRESSES.

PR #956, opened 2018-06-13
Retryable writes

p-mongo:
Ruby driver implements legacy retryable writes, modern retryable writes, retryable reads and change stream resumes.

This PR:

1. Modifies modern retryable write logic (when the driver is using a session with mongodb 3.6+) to match the current retryable write specification. This includes:
  - consolidating error messages that are retried (to not master & node is recovering)
  - adding retries on specified codes (code names are ignored)
2. Does not modify legacy retryable write logic. Specifically:
  - network errors are not retried in legacy logic, whereas they are in modern logic
  - error messages that are retried are consolidated to not master & node is recovering, as they are for modern retry logic, because examination of server code revealed that other messages the driver used to match on seemingly were never returned by mongodb 2.6+ (maybe they were used prior to 2.6)
  - note that this code path can (still) be taken on mongodb 3.6+ if session or write concern requirements for modern retryable writes are not met
3. Does not modify read retry logic. Read retries are scheduled to be specified and implemented across all drivers soon.
4. Changes change stream resume logic to match the current specification (retry getMore on all errors other than ones in the specified list, never retry non-getMore errors)

https://jira.mongodb.org/browse/RUBY-1333
Also takes care of https://jira.mongodb.org/browse/RUBY-1327


p-mongo:
Need to repair bulk write behavior.

lib/mongo/error/operation_failure.rb
```diff
@@ -77,12 +94,73 @@ def retryable?
...
       #
       # @since 2.4.2
       def write_retryable?
-        WRITE_RETRY_MESSAGES.any? { |m| message.include?(m) }
+        WRITE_RETRY_MESSAGES.any? { |m| message.include?(m) } ||
+        write_retryable_code?
+      end
+
+      def write_retryable_code?
+        if code
```
saghm: I think this could be made more succient by just doing `code && WRITE_RETRY_ERRORS.any? { |e| e[:code] == code }`

lib/mongo/retryable.rb
```diff
@@ -123,7 +123,7 @@ def write_with_retry(session, write_concern, &block)
 
     def retry_write_allowed?(session, write_concern)
       session && session.retry_writes? &&
-          (write_concern.nil? || write_concern.acknowledged?)
+          (write_concern.nil? || write_concern.acknowledged?) or false
```
saghm: Just to make sure, is the point of using `or false` here to ensure that the method returns false instead of nil?

p-mongo: Yep.

lib/mongo/error/operation_failure.rb
p-mongo: That would evaluate to nil if code is nil, which it can be.

saghm:
Some of the changes I made for transactions conflict with the work you did here. Given that I was just doing the bare minimum to get the tests to pass, your changes supersede mine, but we still need the transactions tests to pass. Let me know if you need any help with that after fixing all the conflicts.

PR #957, opened 2018-06-13
Shorten matrix names to make status in pull requests more readable

p-mongo:


.evergreen/.evg.yml
```diff
@@ -533,15 +533,19 @@ buildvariants:
...
   run_on:
     - rhel70-small
   tasks:
      - name: "test"
 
-- matrix_name: "rhel70-latest-all-rubies-all-topologies-all-auth-all-ssl-compression"
+-
+  #matrix_name: "rhel70-latest-all-rubies-all-topologies-all-auth-all-ssl-compression"
+  matrix_name: "3.6-compression"
   matrix_spec: { auth-and-ssl: "*", ruby: "*", mongodb-version: "3.6", topology: "*", compressor: 'zlib' }
```
behackett: Compression should be tested against 3.6+. I think you can specify an array here of 3.6, 4.0, and "latest".

.evergreen/.evg.yml
```diff
@@ -533,16 +533,20 @@ buildvariants:
 #  display_name: "${version} ${topology} ${auth-and-ssl} ${ruby} ${os}"
 #  tasks:
 #     - name: "test"
-- matrix_name: "rhel70-all-mongodbs-all-rubies-all-topologies-all-auth-and-ssl"
+-
+  #matrix_name: "rhel70-all-mongodbs-all-rubies-all-topologies-all-auth-and-ssl"
```
saghm: We can delete this instead of commenting; if we want to go back at some point, we can just check the git history.

.evergreen/.evg.yml
```diff
@@ -533,16 +533,20 @@ buildvariants:
...
   display_name: "${mongodb-version} ${topology} ${auth-and-ssl} ${ruby}"
   run_on:
     - rhel70-small
   tasks:
      - name: "test"
 
-- matrix_name: "rhel70-latest-all-rubies-all-topologies-all-auth-all-ssl-compression"
-  matrix_spec: { auth-and-ssl: "*", ruby: "*", mongodb-version: "3.6", topology: "*", compressor: 'zlib' }
+-
+  #matrix_name: "rhel70-latest-all-rubies-all-topologies-all-auth-all-ssl-compression"
```
saghm: Ditto above about deleting comment.

p-mongo: Amended.

PR #958, opened 2018-06-14
RUBY-1317 Implement transactions spec

saghm:
@ShaneHarvey feel free to take a look at this as well if you have time, but no worries if you're too busy


lib/mongo/client.rb
```diff
@@ -341,6 +343,19 @@ def with(new_options = Options::Redacted.new)
       end
     end
 
+    # Get the read concern for this client.
+    #
+    # @example Get the client read concern.
+    #   client.read_concern
+    #
+    # @return [ Symbol ] The read concern.
```
behackett: In collection.rb read_concern is documented as a Hash. Why is it a Symbol here?

saghm: Ah, this is a mistake left over from when I originally had this method just returning the level rather than the entire read concern. I'll fix it.

lib/mongo/client.rb
```diff
@@ -232,6 +233,7 @@ def hash
     # @option options [ String ] :user The user name.
     # @option options [ Hash ] :write The write concern options. Can be :w =>
     #   Integer|String, :fsync => Boolean, :j => Boolean.
+    # @option options [ Symbol ] :read_concern The read concern option.
```
behackett: Docs need fixing here too.

lib/mongo/error/invalid_transaction_operation.rb
```diff
@@ -0,0 +1,82 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module Mongo
+  class Error
+
+    # Exception raised if the object is not a valid update document.
```
p-mongo: Does this description need to be changed?

lib/mongo/error/invalid_transaction_operation.rb
```diff
@@ -0,0 +1,82 @@
...
+
+      # The error message for when a transaction is started with an unacknowledged write concern.
+      #
+      # @since 2.6.0
+      UNACKNOWLEDGED_WRITE_CONCERN = 'transactions do not support unacknowledged write concern'.freeze
+
+      # Instantiate the new exception.
+      #
+      # @example Instantiate the exception.
+      #   Mongo::Error::InvalidUpdateDocument.new
```
p-mongo: Since constructor parameter is required it should be given here.

lib/mongo/error/invalid_transaction_operation.rb
```diff
@@ -0,0 +1,82 @@
...
+
+      # Create an error message for incorrectly running a transaction operation twice.
+      #
+      # @example Create the error message.
+      #   InvalidTransactionOperation.cannot_call_twice(op)
+      #
+      # @param [ Symbol ] op The operation which was run twice.
+      #
+      # @since 2.6.0
+      def self.cannot_call_twice(op)
```
p-mongo: I would call this `cannot_call_twice_message`

lib/mongo/error/invalid_transaction_operation.rb
```diff
@@ -0,0 +1,82 @@
...
+      # after the previous one.
+      #
+      # @example Create the error message.
+      #   InvalidTransactionOperation.cannot_call_after(last_op, current_op)
+      #
+      # @param [ Symbol ] last_op The operation which was run before.
+      # @param [ Symbol ] current_op The operation which cannot be run.
+      #
+      # @since 2.6.0
+      def self.cannot_call_after(last_op, current_op)
```
p-mongo: And `cannot_call_after_message`

lib/mongo/error/no_server_available.rb
```diff
@@ -29,11 +29,30 @@ class NoServerAvailable < Error
       #   satisfied.
       #
       # @since 2.0.0
-      def initialize(server_selector)
+      def initialize(server_selector, labels = nil)
```
p-mongo: `labels` are not documented.

lib/mongo/error/no_server_available.rb
```diff
@@ -29,11 +29,30 @@ class NoServerAvailable < Error
...
+
+      # Does the error have the given label?
+      #
+      # @example
+      #   error.label?(label)
+      #
+      # @return [ true, false ] Whether the error has the given label.
+      #
+      # @since 2.6.0
+      def label?(label)
```
p-mongo: `has_label?`, analogous to e.g. `has_key?` on Hashes.

lib/mongo/error/operation_failure.rb
```diff
@@ -81,7 +100,28 @@ def retryable?
...
+
+      # Does the error have the given label?
+      #
+      # @example
+      #   error.label?(label)
+      #
+      # @return [ true, false ] Whether the error has the given label.
+      #
+      # @since 2.6.0
+      def label?(label)
```
p-mongo: `has_label?` here also

p-mongo: The parameter is not documented. What type is it?

lib/mongo/operation/list_collections/op_msg.rb
```diff
@@ -26,20 +26,13 @@ class OpMsg
...
-        # Execute the operation.
-        #
-        # @example
-        #   operation.execute(server)
-        #
-        # @param [ Mongo::Server ] server The server to send the operation to.
-        #
-        # @return [ Mongo::Operation::ListCollections::Result ] The operation result.
-        #
-        # @since 2.5.2
```
p-mongo: Why was this docstring deleted?

lib/mongo/session.rb
```diff
@@ -119,6 +176,20 @@ def ended?
...
 
+    # Add the autocommit field to a command document if applicable.
+    #
+    # @example
+    #   session.add_autocommit!(cmd)
+    #
+    # @return [ Hash, BSON::Document ] The command document.
+    #
+    # @since 2.6.0
+    def add_autocommit!(command)
```
p-mongo: `command` is not documented. I will stop commenting on this but the entire diff should be reviewed for undocumented parameters.

p-mongo: Also, I imagine many of the added methods are not meant to be called from outside the driver and should be marked with `@api private` to indicate that.

lib/mongo/error/no_server_available.rb
saghm: rubocop suggests not using `has_` as a prefix on boolean methods like this, specifically citing that [`key?`](https://ruby-doc.org/core-2.5.1/Hash.html#method-i-key-3F) is preferred to `has_key?`. I don't feel strongly on this though, so I can change this if you prefer.

lib/mongo/session.rb
saghm: For the `add_*` methods, I followed the convention of the existing methods (e.g. `add_id`), which didn't put API private (or document their parameters).

p-mongo:
Makes sense, let's leave it at `label?`.

lib/mongo/error/no_server_available.rb
p-mongo: Let's keep it as `label?`.

lib/mongo/session.rb
p-mongo: After reading driver code for the last month I found that it was easy to get a general idea of what the method does but specific parameter requirements are sometimes difficult to figure out. Hence the value of documentation should be in providing these details. For example, it seems like this method would mutate passed command but one can't tell for sure from the docstring.

p-mongo:
There are many instances of duplicated code added in this PR. I think the duplication can be reduced, but in the interests of meeting the 4.0 deadline this can be left for after the release provided the new code is marked private when possible so that we are not committed to maintaining the APIs.

lib/mongo/session.rb
saghm: Makes sense! Let's file a ticket to sweep through the public API and document all the parameters that are missing; we can do it after the release.

saghm:
@p-mongo feel free to file tickets for where you think refactoring would improve the code

PR #959, opened 2018-06-18
RUBY-1362 Implement SASLprep for Ruby versions below 2.2

saghm:
This PR also gets rid of the messy testing conditionals that were added as part of the work for implementing SCRAM-SHA-256, as they are no longer necessary after this change.


lib/mongo/auth/stringprep.rb
```diff
@@ -104,7 +96,12 @@ def mapping(c, mappings)
       end
 
       def normalize!(out)
-        out.unicode_normalize!(:nfkc)
+        if RUBY_VERSION < '2.2.0'
+          require 'mongo/auth/stringprep/unicode_normalize/normalize'
```
p-mongo: You may consider only requiring this if `::UnicodeNormalize` is not an existing constant.

saghm:
This was taken out of the latest stable, 2.5.1

THIRD-PARTY-NOTICES
```diff
@@ -0,0 +1,64 @@
+------------------------------------------------------------------------
```
behackett: Use the format found here:

https://github.com/mongodb/mongo-python-driver/blob/master/THIRD-PARTY-NOTICES#L1-L16

As specified here:

https://jira.mongodb.org/browse/DRIVERS-323

Including specifying what files the license applies to.

lib/mongo/auth/stringprep.rb
behackett: Good idea. That covers alternative implementations that might add it in a different version. For example, does RUBY_VERSION < '2.2.0' also cover JRuby?

saghm:
Updated for feedback

PR #960, opened 2018-06-18
RUBY-1354 Implement new count API

saghm:


spec/spec_tests/crud_spec.rb
```diff
@@ -6,27 +6,25 @@
...
-    end
-
     context(spec.description) do
 
       spec.tests.each do |test|
 
         context(test.description) do
 
           before(:each) do
+            unless spec.server_version_satisfied?(authorized_client)
```
saghm: When testing locally, this block skipped _all_ of the tests even when not all of them entered the `unless` block; moving this block here seemed to fix it.

lib/mongo/collection.rb
```diff
@@ -350,10 +350,54 @@ def watch(pipeline = [], options = {})
...
+    # @option opts :limit [ Integer ] Max number of docs to return.
+    # @option opts :max_time_ms [ Integer ] The maximum amount of time to allow the
+    #   command to run.
+    # @option opts [ Hash ] :read The read preference options.
+    # @option opts [ Hash ] :collation The collation to use.
+    #
+    # @return [ Integer ] The document count.
+    #
+    # @since 2.6.0
+    def count_documents(filter = nil, options = {})
```
behackett: filter is a required parameter. The user has to provide a hash, but an empty hash is a valid value.

lib/mongo/collection.rb
```diff
@@ -350,10 +350,54 @@ def watch(pipeline = [], options = {})
...
+    # Get a count of matching documents in the collection.
+    #
+    # @example Get the number of documents in the collection.
+    #   collection_view.count_documents
+    #
+    # @param [ Hash ] filter A filter for matching documents.
+    # @param [ Hash ] options Options for the count command.
+    #
+    # @option opts :skip [ Integer ] The number of documents to skip.
+    # @option opts :hint [ Hash ] Override default index selection and force
```
behackett: Document that hint requires MongoDB 3.6+.

https://docs.mongodb.com/manual/reference/command/aggregate/

lib/mongo/collection/view/readable.rb
```diff
@@ -150,6 +152,68 @@ def count(opts = {})
...
+        def estimated_document_count(opts = {})
+          cmd = { count: collection.name }
+          cmd[:maxTimeMS] = opts[:max_time_ms] if opts[:max_time_ms]
+          cmd[:readConcern] = collection.read_concern if collection.read_concern
+          read_pref = opts[:read] || read_preference
+          selector = ServerSelector.get(read_pref || server_selector)
+          with_session(opts) do |session|
+            read_with_retry(session) do
+              server = selector.select_server(cluster)
+              apply_collation!(cmd, server, opts)
```
behackett: Collation doesn't add any value here since there is no filter and / or sort.

lib/mongo/collection.rb
saghm: Will do; should I add this to the existing methods as well (which don't seem to document it already)?

lib/mongo/collection.rb
```diff
@@ -350,10 +350,54 @@ def watch(pipeline = [], options = {})
...
+    #
+    # @option opts :max_time_ms [ Integer ] The maximum amount of time to allow the command to
+    #   run.
+    # @option opts [ Hash ] :read The read preference options.
+    #
+    # @return [ Integer ] The document count.
+    #
+    # @since 2.6.0
+    def estimated_document_count(options = {})
+      View.new(self, {}, options).estimated_document_count(options)
```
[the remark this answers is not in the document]
saghm: It will not; `View#estimated_document_count` does not add a filter, but `View` expects that the filter will be a hash (even if it doesn't get used).

lib/mongo/collection.rb
saghm: Ack

lib/mongo/collection/view/readable.rb
```diff
@@ -150,6 +152,67 @@ def count(opts = {})
           end
         end
 
+        # Get a count of matching documents in the collection.
+        #
+        # @example Get the number of documents in the collection.
+        #   collection_view.count
+        #
+        # @param [ Hash ] opts Options for the count command.
```
behackett: Same here. Not the count command.

lib/mongo/collection/view/readable.rb
```diff
@@ -150,6 +152,67 @@ def count(opts = {})
...
+          opts.select! { |k, _| [:hint, :max_time_ms, :read, :collation].include?(k) }
+          aggregate(pipeline, opts).first['n'].to_i
+        end
+
+        # Gets an estimate of the count of documents in a collection using collection metadata.
+        #
+        # @example Get the number of documents in the collection.
+        #   collection_view.estimated_document_count
+        #
+        # @param [ Hash ] opts Options for the count command.
```
behackett: And here. The count command is an implementation detail.

lib/mongo/collection.rb
```diff
@@ -350,10 +350,54 @@ def watch(pipeline = [], options = {})
...
       View.new(self, filter || {}, options).count(options)
     end
 
+    # Get a count of matching documents in the collection.
+    #
+    # @example Get the number of documents in the collection.
+    #   collection_view.count_documents
+    #
+    # @param [ Hash ] filter A filter for matching documents.
+    # @param [ Hash ] options Options for the count command.
```
behackett: Not the count command. Say something like the options for this operation.

lib/mongo/collection.rb
```diff
@@ -350,10 +350,54 @@ def watch(pipeline = [], options = {})
...
+    def count_documents(filter, options = {})
+      View.new(self, filter, options).count_documents(options)
+    end
+
+    # Gets an estimate of the count of documents in a collection using collection metadata.
+    #
+    # @example Get the number of documents in the collection.
+    #   collection_view.estimated_document_count
+    #
+    # @param [ Hash ] options Options for the count command.
```
behackett: Same here. The count command is an implementation detail here. This helper may use the $collStats aggregate stage in the future.

lib/mongo/collection.rb
```diff
@@ -350,10 +350,54 @@ def watch(pipeline = [], options = {})
     # @return [ Integer ] The document count.
     #
     # @since 2.1.0
+    #
+    # @deprecated Use count_documents or estimated_document_count instead.
     def count(filter = nil, options = {})
       View.new(self, filter || {}, options).count(options)
     end
 
+    # Get a count of matching documents in the collection.
```
p-mongo: I think this should explain the difference between count_documents and the deprecated count, to prevent user confusion.

lib/mongo/collection/view/readable.rb
```diff
@@ -150,6 +152,67 @@ def count(opts = {})
           end
         end
 
+        # Get a count of matching documents in the collection.
```
p-mongo: "Get the number of matching documents" or "count the matching documents" I think would be more grammatically correct.

spec/spec_helper.rb
```diff
@@ -312,5 +312,26 @@ def initialize_scanned_client!
...
 end
 
+# Converts a 'camelCase' string or symbol to a :snake_case symbol.
+def camel_to_snake(ident)
+  ident = ident.is_a?(String) ? ident.dup : ident.to_s
+  ident[0] = ident[0].downcase
+  ident.chars.reduce('') { |s, c| s + (/[A-Z]/ =~ c ? "_#{c.downcase}" : c) }.to_sym
+end
+
+# Creates a copy of a hash where all keys and string values are converted to camel-case symbols.
```
p-mongo: Snake case symbols?

spec/spec_helper.rb
```diff
@@ -312,5 +312,26 @@ def initialize_scanned_client!
...
+# For example, `{ 'fooBar' => { 'baz' => 'bingBing', :x => 1 } }` converts to
+# `{ :foo_bar => { :baz => :bing_bing, :x => 1 } }`.
+def symbolize_hash(value)
```
p-mongo: `snakeize_hash`?

spec/support/crud.rb
```diff
@@ -83,7 +83,8 @@ def initialize(file)
       #
       # @since 2.4.0
       def server_version_satisfied?(client)
-        lower_bound_satisfied?(client) && upper_bound_satisfied?(client)
+        x = lower_bound_satisfied?(client) && upper_bound_satisfied?(client)
+        x
```
p-mongo: Undo?

behackett:
@p - estimated_document_count accepts one argument for the count command - maxTimeMS. That said, since the Ruby driver supports explicit read preference everywhere in the API it should be allowed here as well. It's not a count command option, but a directive to the driver or mongos to choose a specific server for the operation.

saghm:
Updated for feedback

lib/mongo/collection.rb
```diff
@@ -350,10 +350,55 @@ def watch(pipeline = [], options = {})
...
+    # @example Get the number of documents in the collection.
+    #   collection_view.count_documents
+    #
+    # @param [ Hash ] filter A filter for matching documents.
+    # @param [ Hash ] options Options for the operation.
+    #
+    # @option opts :skip [ Integer ] The number of documents to skip.
+    # @option opts :hint [ Hash ] Override default index selection and force
+    #   MongoDB to use a specific index for the query. Requires server version 3.6+.
+    # @option opts :limit [ Integer ] Max number of docs to return.
```
behackett: Minor nit, this is a limit of the number of documents to count, not return.

lib/mongo/collection/view/readable.rb
```diff
@@ -150,6 +152,67 @@ def count(opts = {})
...
+        #
+        # @example Get the number of documents in the collection.
+        #   collection_view.count
+        #
+        # @param [ Hash ] opts Options for the operation.
+        #
+        # @option opts :skip [ Integer ] The number of documents to skip.
+        # @option opts :hint [ Hash ] Override default index selection and force
+        #   MongoDB to use a specific index for the query. Requires server version 3.6+.
+        # @option opts :limit [ Integer ] Max number of docs to return.
```
behackett: Same here.

p-mongo:
I meant that, for instance, the full count supports conditions and the estimated count doesn't. Do we have a test mirroring a user invoking the estimated count method with conditions by mistake, such as due to performing a global search and replace?

saghm:
> I meant that, for instance, the full count supports conditions and the estimated count doesn't. Do we have a test mirroring a user invoking the estimated count method with conditions by mistake, such as due to performing a global search and replace?

@p-mongo I'm not really sure how we could test that; `estimated_document_count` only takes a single parameter, namely the options. I don't think it's worth adding a test to check that an error is raised when you pass two arguments to a method with a single parameter, and I'm not sure what a test would even look like to ensure that passing a filter instead of options doesn't run a regular count.

Beyond that, are there any other things that you think need to be addressed before merging?

p-mongo:
OK that's fine. Good to merge.

PR #961, opened 2018-06-18
Command monitoring improvements and tests

p-mongo:
Added an integration test to verify command monitoring functionality per the spec. Spec needed to be fixed - https://github.com/mongodb/specifications/pull/359.

Changes made to monitoring code itself:
- DRY the code - extract Subscribable module with common methods for per-client and global subscribers
- Add unsubscribe method so that the testing can be done
- Document how multiple subscriptions of the same listener are handled
- ~~Expose command arguments in started and failure events since the events have them~~
- Expose failure document in failure events per specification
- `command_name` on command monitoring events is now always a string, as was documented

### DIFF
--- a/lib/mongo/monitoring/event/command_failed.rb
+++ b/lib/mongo/monitoring/event/command_failed.rb
@@ -63,6 +63,7 @@ module Mongo
         # @param [ Float ] duration The duration the command took in seconds.
         #
         # @since 2.1.0
+        # @api private
         def initialize(command_name, database_name, address, request_id, operation_id, message, failure, duration)
           @command_name = command_name.to_s
           @database_name = database_name
@@ -89,6 +90,7 @@ module Mongo
         # @return [ CommandFailed ] The event.
         #
         # @since 2.1.0
+        # @api private
         def self.generate(address, operation_id, payload, message, failure, duration)
           new(
             payload[:command_name],
--- a/lib/mongo/monitoring/event/command_failed.rb
+++ b/lib/mongo/monitoring/event/command_failed.rb
@@ -53,24 +53,24 @@ module Mongo
         #
         # @example Create the event.
         #
-        # @param [ BSON::Document ] failure The error document, if any.
         # @param [ String ] command_name The name of the command.
         # @param [ String ] database_name The database_name name.
         # @param [ Server::Address ] address The server address.
         # @param [ Integer ] request_id The request id.
         # @param [ Integer ] operation_id The operation id.
         # @param [ String ] message The error message.
+        # @param [ BSON::Document ] failure The error document, if any.
         # @param [ Float ] duration The duration the command took in seconds.
         #
         # @since 2.1.0
-        def initialize(failure, command_name, database_name, address, request_id, operation_id, message, duration)
-          @failure = failure
+        def initialize(command_name, database_name, address, request_id, operation_id, message, failure, duration)
           @command_name = command_name.to_s
           @database_name = database_name
           @address = address
           @request_id = request_id
           @operation_id = operation_id
           @message = message
+          @failure = failure
           @duration = duration
         end
 
@@ -79,25 +79,25 @@ module Mongo
         # @example Create the event.
         #   CommandFailed.generate(address, 1, payload, duration)
         #
-        # @param [ BSON::Document ] failure The error document, if any.
         # @param [ Server::Address ] address The server address.
         # @param [ Integer ] operation_id The operation id.
         # @param [ Hash ] payload The message payload.
         # @param [ String ] message The error message.
+        # @param [ BSON::Document ] failure The error document, if any.
         # @param [ Float ] duration The duration of the command in seconds.
         #
         # @return [ CommandFailed ] The event.
         #
         # @since 2.1.0
-        def self.generate(failure, address, operation_id, payload, message, duration)
+        def self.generate(address, operation_id, payload, message, failure, duration)
           new(
-            failure,
             payload[:command_name],
             payload[:database_name],
             address,
             payload[:request_id],
             operation_id,
             message,
+            failure,
             duration
           )
         end
--- a/lib/mongo/monitoring/event/command_failed.rb
+++ b/lib/mongo/monitoring/event/command_failed.rb
@@ -69,7 +69,7 @@ module Mongo
         # @since 2.1.0
         def initialize(failure, command_name, command, database_name, address, request_id, operation_id, message, duration)
           @failure = failure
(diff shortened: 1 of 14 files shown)


lib/mongo/monitoring/event/command_failed.rb
```diff
@@ -27,13 +27,23 @@ class CommandFailed < Mongo::Event::Base
...
         # @return [ String ] database_name The name of the database_name.
         attr_reader :database_name
 
         # @return [ Float ] duration The duration of the command in seconds.
         attr_reader :duration
 
-        # @return [ String ] message The error message.
+        # @return [ BSON::Document ] failure The error document, if present.
+        #   This will only be filled out for errors communicated by a
+        #   MongoDB server. In other situations, for example in case of
```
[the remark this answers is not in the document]
p-mongo: OK, do you want me to do something like that?

lib/mongo/monitoring/event/command_failed.rb
```diff
@@ -27,13 +27,23 @@ class CommandFailed < Mongo::Event::Base
         # @return [ String ] command_name The name of the command.
         attr_reader :command_name
 
+        # @return [ BSON::Document ] command The command arguments.
+        attr_reader :command
```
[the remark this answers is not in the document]
p-mongo: For usability - right now I print all 3 types of events and look at it as a human to figure out what is going on. Imagine a system that wanted to log failures - it would have to process all start events, record them somewhere, then for each failure it would need to figure out the matching start event to figure out the command. All of this when the driver has the command trivially available upon operation completion.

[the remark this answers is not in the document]
behackett: The request_id and, especially for batched write operations, the operation_id fields serve the purpose of correlating start, succeed, and fail events. The API in the spec was standardized for use by APM tools like New Relic and App Dynamics. We can't change the API in one driver without doing it in all other drivers. I'm open to adding more fields, but we can't unilaterally do it here. Open a spec ticket to start that discussion for a later release.

https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst#api

I noticed that the spec says operation_id is optional, but that appears to be a bug in the spec. A different section of the spec requires it, and you can't correlate bulk write operations that require multiple batches with out it. Happily that doesn't seem to be a problem in the Ruby driver. :-)

https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst#bulk-writes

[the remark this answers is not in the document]
p-mongo: OK I will take this out, do you want me to make a PR or ticket for spec to mark operation id required?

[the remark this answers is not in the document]
behackett: Send a pull request when you have some free time.

p:
Addressed feedback, found out that command_name is sometimes a symbol and sometimes a string contrary to both the driver docstrings and the specification. Made it to always be a string. This can be a breaking change for clients that look for specific commands by symbols but existing behavior contradicts our own documentation and is an annoyance (and nothing prevents the driver from switching symbols to string at any point thus breaking code that would be broken by this change anyway).

Clients like newrelic that handle all events shouldn't be affected, as they already must account for command_name being either a string or a symbol.

lib/mongo/monitoring/event/command_failed.rb
```diff
@@ -52,16 +59,18 @@ class CommandFailed < Mongo::Event::Base
         # @param [ Integer ] request_id The request id.
         # @param [ Integer ] operation_id The operation id.
         # @param [ String ] message The error message.
+        # @param [ BSON::Document ] failure The error document, if any.
         # @param [ Float ] duration The duration the command took in seconds.
         #
         # @since 2.1.0
-        def initialize(command_name, database_name, address, request_id, operation_id, message, duration)
-          @command_name = command_name
+        def initialize(command_name, database_name, address, request_id, operation_id, message, failure, duration)
```
saghm: Given that this isn't marked with `@api private`, is it worth putting the new parameter last and adding a default value so that it doesn't break API? An alternative would be to mark this as `@api private`, although that still seems like we'd be making a breaking change.

lib/mongo/monitoring/event/command_failed.rb
```diff
@@ -74,19 +83,21 @@ def initialize(command_name, database_name, address, request_id, operation_id, m
...
         # @param [ Hash ] payload The message payload.
         # @param [ String ] message The error message.
+        # @param [ BSON::Document ] failure The error document, if any.
         # @param [ Float ] duration The duration of the command in seconds.
         #
         # @return [ CommandFailed ] The event.
         #
         # @since 2.1.0
-        def self.generate(address, operation_id, payload, message, duration)
+        def self.generate(address, operation_id, payload, message, failure, duration)
```
saghm: Ditto above about API changes on a non-private method

lib/mongo/monitoring/publishable.rb
```diff
@@ -99,10 +99,10 @@ def command_succeeded(result, address, operation_id, payload, duration)
         )
       end
 
-      def command_failed(address, operation_id, payload, message, duration)
+      def command_failed(failure, address, operation_id, payload, message, duration)
```
saghm: Ditto above API changes on a non-private method

spec/mongo/client_spec.rb
```diff
@@ -1474,7 +1474,7 @@
       end
 
       let(:command) do
-        EventSubscriber.started_events.find { |c| c.command_name == :listDatabases }.command
+        EventSubscriber.started_events.find { |c| c.command_name == 'listDatabases' }.command
```
saghm: I'm wondering if it might be worth adding an `EventSubscriber#command_started?` method (which could convert symbols to strings internally) given that we seem to be doing this operation a lot. Not worth changing for this PR, but something to think about in the future.

lib/mongo/monitoring/event/command_failed.rb
p-mongo: I will mark it private. Can you think of a use case when someone would be creating events outside of the driver?

spec/mongo/client_spec.rb
p-mongo: A filter by command name would make sense, but per Bernie's comments we can't be adding any public methods to events in one driver.

p-mongo: Can certainly do that at some point then.

p-mongo:
@saghm Made the requested changes.

PR #963, opened 2018-06-19
RUBY-1337 Ensure primary hostname check is case-insenstive

saghm:


spec/mongo/server/description_spec.rb
```diff
@@ -431,19 +431,36 @@
...
+          expect(description).to be_primary
+        end
       end
 
-      it 'returns false' do
-        expect(description).to be_primary
+      context 'when the hostname contains capital letters' do
+
+        let(:description) do
+          described_class.new('localhost:27017',
```
p-mongo: Is there a test for this hostname having capital letters and ismaster response being in lowercase?

saghm: The driver always converts addresses to lowercase before attempting to make a connection, so a ServerDescription would never have a non-lowercase address. The issue came up specifically because the ServerDescription was not changing the address in the primary host field of the isMaster response in the same way.

p-mongo: OK then.

PR #965, opened 2018-06-22
RUBY-1267 Atlas connectivity tests for Ruby driver

p-mongo:
https://jira.mongodb.org/browse/RUBY-1267


.evergreen/evg-atlas.yml
```diff
@@ -0,0 +1,432 @@
...
+  - func: "install dependencies"
+
+axes:
+  - id: "atlas"
+    display_name: Atlas
+    values:
+      - id: "atlas"
+        display_name: "atlas"
+        variables:
+           RVM_RUBY: "ruby-2.5.0"
```
behackett: We should test all the versions of Ruby we have in Evergreen. We're looking for TLS issues, and potential workarounds for them, that we can document before users hit them. If the Ruby OpenSSL module is anything like the Python ssl module, it has probably evolved a lot over the years.

.evergreen/run-atlas-tests.sh
```diff
@@ -0,0 +1,85 @@
...
+  export ROOT_USER_PWD="pwd123"
+fi
+if [ "$COMPRESSOR" == "zlib" ]; then
+  export COMPRESSOR="zlib"
+fi
+export CI=true
+export DRIVER_TOOLS_CLIENT_CERT_PEM="${DRIVERS_TOOLS}/.evergreen/x509gen/client-public.pem"
+export DRIVER_TOOLS_CLIENT_KEY_PEM="${DRIVERS_TOOLS}/.evergreen/x509gen/client-private.pem"
+export DRIVER_TOOLS_CA_PEM="${DRIVERS_TOOLS}/.evergreen/x509gen/ca.pem"
+export DRIVER_TOOLS_CLIENT_KEY_ENCRYPTED_PEM="${DRIVERS_TOOLS}/.evergreen/x509gen/password_protected.pem"
```
behackett: You shouldn't use any explicit certs. The Atlas clusters all use a cert signed by DigiCert. One thing we want to check is if the various Ruby versions we support can verify that cert by default.

.evergreen/evg-atlas.yml
p-mongo: Added all the other ones.

.evergreen/run-atlas-tests.sh
p-mongo: That was a copy paste from the main config, took it out.

PR #967, opened 2018-06-22
RUBY-1342 Add change stream helpers for Database and Client

saghm:


lib/mongo/client.rb
```diff
@@ -95,7 +95,10 @@ class Client
     def_delegators :@database, :command, :collections
 
     # Delegate subscription to monitoring.
-    def_delegators :@monitoring, :subscribe, :unsubscribe
+    def_delegators :monitoring, :subscribe, :unsubscribe
+
+    # Delegate monitoring to cluster.
+    def_delegators :cluster, :monitoring
```
saghm: I was running into the same command monitoring issue I ran into in the transactions tests (namely that creating a new client with `with` left the client in the state where its `monitoring` instance was different than `cluster`'s. I realized that since they are always supposed to be the same, the cleanest solution is to have clients just not have their own instance and delegate to the clusters.

p-mongo: :+1:

spec/support/change_streams.rb
```diff
@@ -0,0 +1,266 @@
...
+          @expectations.each_with_index.all? do |e, i|
+            actual[i] && match?(e, actual[i])
+          end
+        end
+
+        private
+
+        def events
+          EventSubscriber.started_events.reduce([]) do |evs, e|
+            next evs if [:saslStart, :saslContinue, :killCursors].include?(e.command_name)
```
p-mongo: command_name is now always a string thus this shouldn't work.

saghm: Interestingly, it does work though; I inspected the output with and without this line, and the sasl and killCursor commands are indeed removed (and the branch is rebased against master).

p-mongo: Okay great.

PR #973, opened 2018-06-25
RUBY-1342 Fix broken push

saghm:


saghm:
@p-mongo This fixes the build failure; I had these changes pushed at the point that you LGTM'd but accidentally pushed over them after rebasing against master locally

p-mongo:
There are merge conflict markers in the diff which may be contributing to evergreen failing every build.

PR #974, opened 2018-06-25
RUBY-1347 Do not use sessions by default in parallel scan.

p-mongo:
A user is able to explicitly provide a session still,
but the driver will execute the parallel scan outside of a session
by default.

https://jira.mongodb.org/browse/RUBY-1347

### DIFF
--- a/lib/mongo/collection/view/readable.rb
+++ b/lib/mongo/collection/view/readable.rb
@@ -549,22 +549,26 @@ module Mongo
         end
 
         def parallel_scan(cursor_count, options = {})
-          session = client.send(:get_session, @options)
+          if options[:session]
+            session = client.send(:get_session, @options)
+          else
+            session = nil
+          end
           server = server_selector.select_server(cluster)
           cmd = Operation::ParallelScan.new({
                   :coll_name => collection.name,
                   :db_name => database.name,
                   :cursor_count => cursor_count,
                   :read_concern => collection.read_concern,
-                  :session => session
+                  :session => session,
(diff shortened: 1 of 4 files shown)


lib/mongo/collection/view/readable.rb
```diff
@@ -574,7 +578,7 @@ def parallel_scan(cursor_count, options = {})
                          :coll_name => collection.name
                      }).execute(server)
             end
-            Cursor.new(self, result, server, session: session)
+            Cursor.new(self, result, server)
```
behackett: If parallelCollectionScan was run with an explicit session the cursors all must use that same explicit session.

p-mongo: Yep thanks.

lib/mongo/collection/view/readable.rb
```diff
@@ -549,22 +549,26 @@ def server_selector
...
                   :cursor_count => cursor_count,
                   :read_concern => collection.read_concern,
-                  :session => session
+                  :session => session,
                 }.merge!(options))
           cmd.execute(server).cursor_ids.map do |cursor_id|
             result = if server.features.find_command_enabled?
                        Operation::GetMore.new({
-                         :selector => {:getMore => cursor_id,
+                          :selector => {:getMore => cursor_id,
```
saghm: It's hard to tell from the Github diff viewer, but is this being changed from 2 to 3 space indentation (relative to the beginning of `Operation::GetMore`)?

PR #976, opened 2018-06-25
RUBY-1340 Provide Transactions example for Docs

saghm:
See the ticket description for the shell code being ported

### DIFF
--- a/lib/mongo/error/socket_timeout_error.rb
+++ b/lib/mongo/error/socket_timeout_error.rb
@@ -20,10 +20,6 @@ module Mongo
     # @since 2.0.0
     class SocketTimeoutError < Error
       include WriteRetryable
-
-      def initialize(msg = nil)
-        super(msg) if msg
-      end
     end
   end
 end
--- a/lib/mongo/error/socket_timeout_error.rb
+++ b/lib/mongo/error/socket_timeout_error.rb
@@ -20,6 +20,10 @@ module Mongo
     # @since 2.0.0
     class SocketTimeoutError < Error
       include WriteRetryable
+
+      def initialize(msg = nil)
+        super(msg) if msg
+      end
     end
   end
 end
--- a/spec/mongo/transactions_examples_spec.rb
+++ b/spec/mongo/transactions_examples_spec.rb
@@ -47,8 +47,8 @@ describe 'transactions examples in Ruby', if: test_transactions? do
         rescue Mongo::Error::OperationFailure, Mongo::Error::SocketError,
                Mongo::Error::NoServerAvailable => e
 
-          if e.label?(Mongo::Session::UNKNOWN_TRANSACTION_COMMIT_LABEL)
-            puts "#{Mongo::Session::UNKNOWN_TRANSACTION_COMMIT_LABEL}, retrying commit operation..."
+          if e.label?(Mongo::Error::UNKNOWN_TRANSACTION_COMMIT_RESULT_LABEL)
+            puts "#{Mongo::Error::UNKNOWN_TRANSACTION_COMMIT_RESULT_LABEL}, retrying commit operation..."
             next
           else
             puts 'Error during commit ...'
@@ -87,9 +87,9 @@ describe 'transactions examples in Ruby', if: test_transactions? do
                  Mongo::Error::NoServerAvailable => e
 
             puts 'Transaction aborted. Caught exception during transaction.'
-            raise unless e.label?(Mongo::Session::TRANSIENT_TRANSACTION_ERROR_LABEL)
+            raise unless e.label?(Mongo::Error::TRANSIENT_TRANSACTION_ERROR_LABEL)
 
-            puts "#{Mongo::Session::TRANSIENT_TRANSACTION_ERROR_LABEL}, retrying transaction ..."
+            puts "#{Mongo::Error::TRANSIENT_TRANSACTION_ERROR_LABEL}, retrying transaction ..."
             next
           end
         end
@@ -122,8 +122,8 @@ describe 'transactions examples in Ruby', if: test_transactions? do
           rescue Mongo::Error::OperationFailure, Mongo::Error::SocketError,
                  Mongo::Error::NoServerAvailable => e
 
-            if e.label?(Mongo::Session::UNKNOWN_TRANSACTION_COMMIT_LABEL)
-              puts "#{Mongo::Session::UNKNOWN_TRANSACTION_COMMIT_LABEL}, retrying commit operation..."
+            if e.label?(Mongo::Error::UNKNOWN_TRANSACTION_COMMIT_RESULT_LABEL)
+              puts "#{Mongo::Error::UNKNOWN_TRANSACTION_COMMIT_RESULT_LABEL}, retrying commit operation..."
               next
             else
               puts 'Error during commit ...'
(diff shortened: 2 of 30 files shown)


spec/mongo/transactions_examples_spec.rb
```diff
@@ -0,0 +1,236 @@
...
+
+      loop do
+        begin
+          session.commit_transaction
+          puts 'Transaction committed.'
+          break
+        rescue Mongo::Error::OperationFailure, Mongo::Error::SocketError,
+               Mongo::Error::NoServerAvailable => e
+
+          if e.label?(Mongo::Session::UNKNOWN_TRANSACTION_COMMIT_LABEL)
```
p-mongo: How did UnknownTransactionCommitResult become UNKNOWN_TRANSACTION_COMMIT_LABEL, and why is it under Session?

spec/mongo/transactions_examples_spec.rb
```diff
@@ -0,0 +1,236 @@
...
+
+      # Start Transactions Retry Example 1
+
+      def run_transaction_with_retry(session)
+        loop do
+          begin
+            yield session # performs transaction
+            break
+          rescue Mongo::Error::OperationFailure, Mongo::Error::SocketError,
+                 Mongo::Error::NoServerAvailable => e
```
p-mongo: Can `Mongo::Error` be rescued here instead?

spec/mongo/transactions_examples_spec.rb
```diff
@@ -0,0 +1,236 @@
...
+
+            puts 'Transaction aborted. Caught exception during transaction.'
+            raise unless e.label?(Mongo::Session::TRANSIENT_TRANSACTION_ERROR_LABEL)
+
+            puts "#{Mongo::Session::TRANSIENT_TRANSACTION_ERROR_LABEL}, retrying transaction ..."
+            next
```
p-mongo: `next` is unneeded here.

spec/mongo/transactions_examples_spec.rb
```diff
@@ -0,0 +1,236 @@
...
+
+    # Start Transactions Intro Example 1
+
+    def update_employee_info(session)
+      employees_coll = session.client.use(:hr)[:employees]
+      events_coll = session.client.use(:reporting)[:events]
+
+      session.start_transaction(read_concern: { level: :snapshot },
+                                write_concern: { w: :majority })
+      employees_coll.update_one({ employee: 3 }, { '$set' => { status: 'Inactive'} },
```
p-mongo: Is `{ '$set' => { status: 'Inactive'} }` equivalent to `{ status: 'Inactive'}` ?

spec/mongo/transactions_examples_spec.rb
saghm: This is the constant the driver uses for this string internally, and it's provided publicly to allow users to use it specifically for this case. I think users should use our constants for labels rather than risk making a typo, so I used the constant here to demonstrate the label.

spec/mongo/transactions_examples_spec.rb
saghm: No; not all error types have labels

spec/mongo/transactions_examples_spec.rb
saghm: Given that the sample code explicitly uses `continue` (as do the already-published examples for other languages), I opted to put it here as well

saghm:
Keep in mind that users won't be looking at this file directly; the docs team has a script that will parse it out and put it on a tab next to the other languages' examples. Check out the DRIVERS ticket for a link to the site with some languages already merged.

spec/mongo/transactions_examples_spec.rb
```diff
@@ -0,0 +1,236 @@
...
+      session = client.start_session
+
+      begin
+        run_transaction_with_retry(session) { |s| update_employee_info(s) }
+      rescue StandardError => e
+        # Do something with error
+        raise
+      end
+
+      # End Transactions Example 3
```
kay-kim: # End Transactions Retry Example 3  instead of End Transactions Example 3

spec/mongo/transactions_examples_spec.rb
p-mongo: How does the user know which exceptions have labels and which don't? How was this set of exceptions determined?

spec/mongo/transactions_examples_spec.rb
p-mongo: It's not necessary there either.

spec/mongo/transactions_examples_spec.rb
p-mongo: The name of the constant does not match the documented name of the label ("result" part has gone away). Additionally, as a user it is not at all clear why this constant exists under Session. This is two cases of driver implementation details being exposed to the users. While specifying a literal string is somewhat error-prone, encouraging users to utilize driver guts like this to me is worse.

spec/mongo/transactions_examples_spec.rb
kay-kim: FYI -- you can see it rendered:

https://docs-mongodbcom-staging.corp.mongodb.com/kay/foo2/core/transactions.html#retry-transaction

(ruby is under Other drop down).

p-mongo:
Well, the code in this PR as written is not even usable without rspec/outside of a test suite, thus it is not clear to me how the docs team is going to make it into something directly usable in an application.

saghm:
This is how we (the drivers team) always does examples for the docs team; I highly encourage you to look at the rendered examples that Kay provided to see what the end result actually looks like. Additionally, you can look at at the Python example (the code review link is on their ticket, which is linked to by the DRIVERS ticket).

If you're still not clear on how this all works, we can discuss it offline; this pull request is not the best place to be litigating the logistics of how we and the docs team coordinate to provide examples for the website.

spec/mongo/transactions_examples_spec.rb
p-mongo: The Java version has constants for labels and exposes them appropriately.

saghm: The only time this label will ever exist is on errors returned by the commit_transaction method, so I don't see how namespacing the error label in this class exposes anything more than that. I don't see any issue with the way that the Python driver deals with this issue, and we're doing it the same way.

spec/mongo/transactions_examples_spec.rb
saghm: Technically, `update_one` is not allowed to be called without a `$set` (or some other `$` key first); `replace_one` is supposed to be used for cases when the entire document is overwritten. I followed what the Python driver did here.

saghm:
I don't find the reasoning behind the suggested changes compelling (especially given that other drivers do things this way and that we already have error messages namespaced under `Mongo::Session`), but in the interests of getting things merged in time for the release, I've made the changes. In the future, let's not bikeshed code that's already been reviewed in unrelated pull requests, especially when we're under a tight deadline for a release.

lib/mongo/error/socket_timeout_error.rb
```diff
@@ -20,6 +20,10 @@ class Error
     # @since 2.0.0
     class SocketTimeoutError < Error
       include WriteRetryable
+
+      def initialize(msg = nil)
+        super(msg) if msg
+      end
```
p-mongo: What is the purpose of this method and why does it call super conditionally?

spec/mongo/transactions_examples_spec.rb
p-mongo: Gotcha.

saghm:
Waiting on CI after a rebase

saghm:
@kay-kim all merged; feel from to parse and post!

kay-kim:
Thanks much!  Will do.

PR #981, opened 2018-06-28
Improve ssl read performance

p-mongo:
https://jira.mongodb.org/browse/RUBY-1364

### DIFF
--- a/lib/mongo/socket.rb
+++ b/lib/mongo/socket.rb
@@ -171,9 +171,10 @@ module Mongo
     def read_from_socket(length)
       data = ('x'*length).force_encoding('BINARY')
       deadline = (Time.now + timeout) if timeout
-      # OpenSSL reads in 16 KB chunks max, there is no benefit to allocating
-      # a larger buffer as extra space won't be used anyway:
-      # https://linux.die.net/man/3/ssl_read
+      # We want to have a fixed and reasonably small size buffer for reads
+      # because, for example, OpenSSL reads in 16 kb chunks max.
+      # Having a 16 mb buffer means there will be 1000 reads each allocating
+      # 16 mb of memory and using 16 kb of it.
       #
       # The binary encoding is important, otherwise ruby performs encoding
       # conversions of some sort during the write into the buffer which
--- a/lib/mongo/socket.rb
+++ b/lib/mongo/socket.rb
@@ -234,6 +234,7 @@ module Mongo
 
     def read_buffer_size
       # Buffer size for non-SSL reads
+      # 64kb
       65536
     end
 
--- a/lib/mongo/socket.rb
+++ b/lib/mongo/socket.rb
@@ -169,17 +169,29 @@ module Mongo
     private
 
     def read_from_socket(length)
-      data = ('x'*length).force_encoding('BINARY')
+      # Just in case
+      if length == 0
+        return ''.force_encoding('BINARY')
+      end
+
       deadline = (Time.now + timeout) if timeout
+
       # We want to have a fixed and reasonably small size buffer for reads
       # because, for example, OpenSSL reads in 16 kb chunks max.
       # Having a 16 mb buffer means there will be 1000 reads each allocating
       # 16 mb of memory and using 16 kb of it.
-      #
+      buf_size = read_buffer_size
+      data = nil
+
+      # If we want to read less than the buffer size, just allocate the
+      # memory that is necessary
+      if length < buf_size
+        buf_size = length
+      end
+
       # The binary encoding is important, otherwise ruby performs encoding
       # conversions of some sort during the write into the buffer which
       # kills performance
-      buf_size = read_buffer_size
       buf = ('x'*buf_size).force_encoding('BINARY')
       retrieved = 0
       begin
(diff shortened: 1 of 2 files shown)


lib/mongo/socket.rb
```diff
@@ -169,11 +169,27 @@ def eof?
     private
 
     def read_from_socket(length)
-      data = String.new
+      data = ('x'*length).force_encoding('BINARY')
```
ShaneHarvey: Does the encoding of the final `data` string matter?

lib/mongo/socket.rb
```diff
@@ -169,11 +169,27 @@ def eof?
...
       deadline = (Time.now + timeout) if timeout
+      # OpenSSL reads in 16 KB chunks max, there is no benefit to allocating
+      # a larger buffer as extra space won't be used anyway:
+      # https://linux.die.net/man/3/ssl_read
```
ShaneHarvey: This comment can move to `ssl.rb`.

lib/mongo/socket.rb
p-mongo: It shouldn't since the contents should be binary data anyway.

lib/mongo/socket.rb
p-mongo: Yep, will reword it to be more relevant to the way the code is now.

lib/mongo/socket.rb
saghm: I assume the character `x` is just a placeholder here so that we can preallocate the string? I think a comment explaining why we're doing this might be useful.

lib/mongo/socket.rb
```diff
@@ -169,11 +169,28 @@ def eof?
...
+      # We want to have a fixed and reasonably small size buffer for reads
+      # because, for example, OpenSSL reads in 16 kb chunks max.
+      # Having a 16 mb buffer means there will be 1000 reads each allocating
+      # 16 mb of memory and using 16 kb of it.
+      #
+      # The binary encoding is important, otherwise ruby performs encoding
+      # conversions of some sort during the write into the buffer which
+      # kills performance
+      buf_size = read_buffer_size
+      buf = ('x'*buf_size).force_encoding('BINARY')
```
saghm: Ditto here. Also, if we're doing this in multiple places, I think it might be useful to have a helper method called something like `binary_string_with_capacity(length)` and just call that everywhere.

lib/mongo/socket.rb
```diff
@@ -186,6 +203,11 @@ def read_from_socket(length)
       data
     end
 
+    def read_buffer_size
+      # Buffer size for non-SSL reads
+      65536
```
saghm: To help any readers who haven't memorized the powers of two, I think a comment here explaining that this is 64kb (mirroring the comment in the SSL socket file) would be useful.

p-mongo:
The last commit restores small read performance and makes large read non-SSL performance better vs master. Current performance deltas on the benchmark in https://github.com/p-mongo/tests/blob/master/transfer-rate/transfer_rate_all.rb:

Reads of small documents: no change
Reads of large documents without SSL: ~20% improvement
Reads of large documents with SSL: ~20x improvement

saghm:
It doesn't look like my first two comments were addressed; can you either respond to them or make the specified changes?

p-mongo:
I think it's pretty clear that 'x'*1000 allocates a string of length 1000.

saghm:
It doesn't seem obvious to me that Ruby won't preallocate a larger size than the initial string, nor that 'x' is merely a placeholder and doesn't have some magic meaning. Moreover, that doesn't change that it's duplicated code and that there's no explanation of why we're doing it.

lib/mongo/socket.rb
```diff
@@ -169,11 +169,57 @@ def eof?
...
+      # If we want to read less than the buffer size, just allocate the
+      # memory that is necessary
+      if length < buf_size
+        buf_size = length
+      end
+
+      # The binary encoding is important, otherwise ruby performs encoding
+      # conversions of some sort during the write into the buffer which
+      # kills performance
+      buf = ('x'*buf_size).force_encoding('BINARY')
```
ShaneHarvey: You can use `String.new(capacity=buf_size, encoding='BINARY')` in Ruby >= 2.4. Should be faster than 'x'*bufsize, see:
https://ruby-doc.org/core-2.4.0/String.html
https://bugs.ruby-lang.org/issues/12024

saghm: Unfortunately we support back to 1.9, hence needing to do this manually

saghm: That being said, it might be worth doing a check here and falling back to manual allocations given that the purpose of these changes are to optimize

saghm: We do a similar thing here due to the lack of string normalization code in older Ruby versions: https://github.com/mongodb/mongo-ruby-driver/blob/d2090d73dddff8ad1eef950bcbee6781377918ef/lib/mongo/auth/stringprep.rb#L98-L105

p-mongo: Thanks for digging that out Shane. It makes a measurable difference for non-SSL/Unix sockets with their 64k buffers, for SSL sockets with 16k buffers the difference is minimal.

PR #986, opened 2018-07-09
stop building every mongodb version on travis

saghm:
Given the long running times of our Travis jobs, it has been requested that we only test the latest version of MongoDB on Travis rather than every version (since we already test all versions on Evergreen). I also don't think it's necessary to test multiple different stable versions of Ruby for similar reasons.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,15 +12,12 @@ before_script:
   - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-${MONGODB_VERSION}.tgz -O /tmp/mongodb.tgz
   - tar -xvf /tmp/mongodb.tgz
   - mkdir /tmp/data
-  - ${PWD}/mongodb-linux-x86_64-${MONGODB_VERSION}/bin/mongod --dbpath /tmp/data --bind_ip 127.0.0.1 --auth &> /dev/null &
+  - ${PWD}/mongodb-linux-x86_64-${MONGODB_VERSION}/bin/mongod --setParameter enableTestCommands=1 --dbpath /tmp/data --bind_ip 127.0.0.1 --auth &> /dev/null &
   - export MONGODB_VERSION=${MONGODB_VERSION}
 
 rvm:
   - 1.9.3
-  - 2.4.2
   - 2.5
-  - ruby-head
-  - jruby-9.1.16.0
 
 branches:
   only:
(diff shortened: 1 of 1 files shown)


p-mongo:
There should definitely be a ticket for this filed by the requester along with what they consider acceptable test time and, perhaps, how much time other drivers consume.

Travis already only tests the single topology, testing single topology on one ruby version on one server version is almost useless for CI purposes.

p-mongo:
Travis tests failed because fail points are not enabled on the cluster.

Now that I made more progress on the toolchain update, it is good that this PR retains ruby-head and jruby targets since they are not tested in Evergreen. Long term it is probably better to test those on evergreen only.

junaruga:
> Travis tests failed because fail points are not enabled on the cluster.

Now the Travis tests are going to be passed by below PR.
https://github.com/mongodb/mongo-ruby-driver/pull/992

>  Long term it is probably better to test those on evergreen only.

As `evergreen` is internal CI in MongoDB, I still need the opened Travis CI to test in the future too as a contributor outside of MongoDB.

saghm:
@junaruga note that when a pull request from an outside contributor is made, we can still start an Evergreen build by merely clicking a button. We do plan on keeping Travis around specifically for outside contributions, but we don't need to test every single configuration on it.

.travis.yml
```diff
@@ -27,29 +24,13 @@ branches:
...
-    - rvm: ruby-head
-  exclude:
-    - rvm: jruby-9.1.16.0
-      env: MONGODB=2.6.12
-    - rvm: jruby-9.1.16.0
-      env: MONGODB=3.0.12
-    - rvm: jruby-9.1.16.0
-      env: MONGODB=3.2.11
-    - rvm: jruby-9.1.16.0
-      env: MONGODB=3.4.1
```
junaruga: It seems that L34 - L41 this `exclude` part does not work.
https://github.com/mongodb/mongo-ruby-driver/blob/master/.travis.yml#L33-L41

Because actually used env is `MONGODB_VERSION`. Not `MONGODB`.

p-mongo:
With #996 merged this PR can remove ruby-head from Travis.

p-mongo:
However, given that tests take a long time to run and downloading a mongodb binary doesn't, I think Travis should test mongodb 4.0 instead of an older version that came on the installed OS.

I suggest the following matrix for right now:

- ruby 2.5, server 4.0 (standalone)
- ruby 1.9, server 2.6 (standalone)
- jruby 9.1, server 4.0 (standalone)

Post toolchain update, consider using jruby 9.2 instead of 9.1.

With parallel runner done, consider testing replica set & sharded clusters on 2.5+4.0 with a concurrency of 4.

saghm:
Bernie had originally suggested that we just run Travis on whatever version of MongoDB travis would install by default, but I'm not opposed to that matrix of Ruby/MongoDB versions. My instinct is that we don't need to do multiple cluster configurations though; I think we want to limit the number of servers we're taking up for the tests at a given time, not just the time that the tests run.

saghm:
@p-mongo this is ready for review again

junaruga:
It looks great! minimal 2 test cases.
Thanks!

saghm:
@junaruga Thanks for your feed during the review process!

PR #987, opened 2018-07-09
fix RUBY-1388

p-mongo:


docs/tutorials/ruby-driver-create-client.txt
```diff
@@ -555,6 +571,49 @@ When ``#close`` is called on a client by any thread, all connections are closed:
...
+   child processes.
+
+It is easiest to not create any Mongo::Client instances prior to the fork.
+If the parent process needs to perform operations on the MongoDB database,
+reset the client in an ``after_fork`` handler which is defined in
+``unicorn.rb``::
+
+.. code-block:: ruby
+
+    after_fork do |server, worker|
```
behackett: This ensures that the monitoring threads are restarted after a fork? Is there a solution for the child deadlocking on a mutex acquired in the parent right before the fork?

p-mongo: `reconnect` restarts the monitoring thread, but I think you are correct, there are 3 mutexes in cluster and more drastic measures are needed to ensure they are not deadlocked.

p-mongo: Reviewing all uses of Mutex I would say that in normal applications they should never deadlock, as the applications shouldn't be forking with any of these mutexes being taken.

lib/mongo/client.rb
```diff
@@ -258,14 +258,34 @@ def hash
     #   See Ruby's Zlib module for valid levels.
     # @option options [ true, false ] :retry_writes Retry writes once when
     #   connected to a replica set or sharded cluster versions 3.6 and up.
+    # @option options [ Proc ] :sdam_proc A Proc to invoke with the client
+    #   as the argument prior to performing service discovery and monitoring.
```
saghm: This should be "server discovery and monitoring"

lib/mongo/client.rb
```diff
@@ -258,14 +258,34 @@ def hash
...
+        addresses = addresses_or_uri
+      end
+      options = options.dup
+      # Special handling for sdam_proc as it is only used during client
+      # construction
+      sdam_proc = options.delete(:sdam_proc)
+      @options = validate_options!(Database::DEFAULT_OPTIONS.merge(options)).freeze
+      @database = Database.new(self, @options[:database], @options)
+      monitoring = Monitoring.new(@options)
+      if sdam_proc
```
saghm: I think this would be cleaner as

```
if sdam_proc
  @cluster = Cluster.new([], monitoring, @options)
  sdam_proc.call(self)
end

@cluster = Cluster.new(addresses, monitoring, @options)
```

p-mongo:
This should be ready to go now.

PR #991, opened 2018-07-20
Add rake tasks to start and stop mongod.

junaruga:
It is hard for me to start mongod manually, when I want to run unit tests.
So, I added rake tasks to start and stop `mongod` daemon.

Start `mongod`.

```
$ bundle exec rake spec:start_mongod
```

Stop `mongod`.

```
$ bundle exec rake spec:stop_mongod
```

Run unit test starting and stopping `mongod`.
When `MONGODB_VERSION` is not set, current installed `mongod` command is used, without downloading MongoDB archive from server.
This is the case I want to use on development.

```
$ bundle exec rake spec:ci
```

When `MONGODB_VERSION` is set, download the MongoDB archive, and run the downloaded `mongod` command. and run unit tests. The behavior is basically same with current CI's one.

```
$ MONGODB_VERSION=3.4.1 bundle exec rake spec:ci
```

Task list.

```
$ bundle exec rake -T
[mongo] Warning: No private key present, creating unsigned gem.
I, [2018-07-20T19:34:49.115691 #23416]  INFO -- : Celluloid 0.17.3 is running in BACKPORTED mode. [ http://git.io/vJf3J ]
rake benchmark:micro:deep:encode    # Benchmarking for deep bson documents
rake benchmark:micro:flat:encode    # Benchmarking for flat bson documents
rake benchmark:micro:full:encode    # Benchmarking for full bson documents
rake benchmark:multi_doc:find_many  # Run the common driver multi-document benchmarking tests
rake benchmark:parallel:import      # Run the common driver paralell ETL benchmarking tests
rake benchmark:single_doc:command   # Run the common driver single-document benchmarking tests
rake build                          # Build mongo-2.6.1.gem into the pkg directory
rake clean                          # Remove any temporary products
rake clobber                        # Remove any generated files
rake docs                           # Generate all documentation
rake docs:yard                      # Generate yard documention
rake install                        # Build and install mongo-2.6.1.gem into system gems
rake install:local                  # Build and install mongo-2.6.1.gem into system gems without net...
rake release[remote]                # Create tag v2.6.1 and build and push mongo-2.6.1.gem to rubyge...
rake spec                           # Run RSpec code examples
rake spec:ci                        # Run RSpec code examples for CI
rake spec:start_mongod              # Start mongod
rake spec:stop_mongod               # Stop mongod
```

I can recommend to use the `rake`'s library `sh`, `mkdir_p` and `rm_rf` instead of `system`. It is good to show the command string as a output.

How do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,6 @@ before_install:
 
 install: ruby -S bundle install --without release development
 
-before_script:
-  - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-${MONGODB_VERSION}.tgz -O /tmp/mongodb.tgz
-  - tar -xvf /tmp/mongodb.tgz
-  - mkdir /tmp/data
-  - ${PWD}/mongodb-linux-x86_64-${MONGODB_VERSION}/bin/mongod --dbpath /tmp/data --bind_ip 127.0.0.1 --auth &> /dev/null &
-  - export MONGODB_VERSION=${MONGODB_VERSION}
-
 rvm:
   - 1.9.3
   - 2.4.2
--- a/Rakefile
+++ b/Rakefile
@@ -22,7 +22,43 @@ end
 task :default => :spec
 
 namespace :spec do
-  task :ci => [:spec]
+  MONGODB_VERSION = ENV['MONGODB_VERSION']
+  TMP_DIR = 'tmp'
+  mongod_cmd = 'mongod'
+  archive_url = nil
+  if MONGODB_VERSION
+    mongod_cmd = File.join(TMP_DIR, "mongodb-linux-x86_64-#{MONGODB_VERSION}", "bin", mongod_cmd)
+    mongod_cmd = File.expand_path(mongod_cmd)
+    archive_url = "http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-#{MONGODB_VERSION}.tgz"
+  end
+  DATA_DIR = File.join(TMP_DIR, 'data')
+  LOG_FILE = File.join(TMP_DIR, 'mongod.log')
+
+  desc "Run RSpec code examples for CI"
+  task :ci => [:start_mongod, :spec, :stop_mongod]
+
+  desc "Start mongod"
+  task :start_mongod do
+    # Initialize temporary directory.
+    rm_rf TMP_DIR
+    mkdir_p TMP_DIR
+
+    # Download and extract MongoDB.
+    if MONGODB_VERSION
+      archive_file = File.join(TMP_DIR, 'mongodb.tgz')
+      sh "wget #{archive_url} -O #{archive_file}"
+      sh "tar -xvf #{archive_file} -C #{TMP_DIR}"
+    end
+
+    # Start mongod
+    mkdir_p DATA_DIR
+    sh "#{mongod_cmd} --dbpath #{DATA_DIR} --bind_ip 127.0.0.1 --auth --fork --logpath #{LOG_FILE}"
+  end
+
+  desc "Stop mongod"
+  task :stop_mongod do
+    sh "#{mongod_cmd} --shutdown --dbpath #{DATA_DIR}"
+  end
 end
 
 task :release => :spec do


.travis.yml
```diff
@@ -8,13 +8,6 @@ before_install:
 
 install: ruby -S bundle install --without release development
 
-before_script:
-  - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-${MONGODB_VERSION}.tgz -O /tmp/mongodb.tgz
-  - tar -xvf /tmp/mongodb.tgz
-  - mkdir /tmp/data
-  - ${PWD}/mongodb-linux-x86_64-${MONGODB_VERSION}/bin/mongod --dbpath /tmp/data --bind_ip 127.0.0.1 --auth &> /dev/null &
-  - export MONGODB_VERSION=${MONGODB_VERSION}
```
junaruga: I did not know why "export MONGODB_VERSION=${MONGODB_VERSION}" is set here again. Because it has already been set at `env - matrix` section at first.

p-mongo:
The thing is that running a standalone mongod (one instance of the server) will run less than half of the tests. The other half requires a replica set, which must be configured a certain way. We also have tests for sharded clusters which require a different setup from a replica set.

If you want to run a single node mongod that's easily accomplishable with:

<pre>
mkdir -p /tmp/mongod; mongod --dbpath /tmp/mongod
</pre>

For downloading and selecting a server version try https://github.com/aheckmann/m/.

junaruga:
@p-mongo sure. Thanks for sharing the situation of the tests.

> mkdir -p /tmp/mongod; mongod --dbpath /tmp/mongod

Well, it might be easy.
But as I am working with laptop pc, every time I have to start and stop mongod manually.

But below tasks looks easier doesn't it?
At least this can be replaced from travis before_script section?

```
rake spec:start_mongod
rake spec:stop_mongod
```

About standalone and replicaset
Creating below tasks looks useful?

```
rake spec:mongod::standalone::start
rake spec:mongod::standalone::stop
rake spec:mongod::replicaset::start
rake spec:mongod::replicaset::stop
```

p-mongo:
Per #986 we are reducing the number of configurations we are running in Travis. We may run one replica set and one sharded cluster setup in Travis, or may not. 

To give you an idea of what is involved for replica sets, take a look at https://github.com/p-mongo/dev/blob/master/script/launch-4.0-replset which is how I launch those and note that that uses mlaunch (from https://github.com/rueckstiess/mtools) which is effectively an additional dependency.

While launching mongods by hand is generally way too annoying to be practical for any length of time, most if not engineers at MongoDB have their own cluster management scripts that all differ in some ways. Having one way of launching mongods in the driver rakefile is not that useful when every person working on the driver does this differently.

p-mongo:
A lighter weight solution is shell functions in your .zshrc or equivalent:

<pre>
launch-mongod-4.0.0() { m 4.0.0 && mkdir -p /tmp/mongod-4.0.0 && mongod --dbpath -tmp/mongod-4.0.0 }
</pre>

Run that in a separate terminal and ctrl-c it when you're done.

p-mongo:
I actually don't know how all of our driver tests are set up still (I haven't needed to touch the sharded cluster configuration yet). We also don't have a single comprehensive CI setup - some configurations are tested in Evergreen only, some in Travis only. The configuration is fairly large and complicated.

I started documenting how to configure the test suite in https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/spec_helper.rb#L1. Unfortunately you are correct, it is not easy for someone who is not a driver engineer to run all of the tests. We have a number of tickets outstanding to improve this situation but it will take time.

junaruga:
> I started documenting how to configure the test suite in https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/spec_helper.rb#L1

I did read the document and pointed #986 now.
In my opinion as a contributor, I am fine to use both Travis and (or) Evergreen.
Though I have to learn how to use Evergreen.

I want to run tests with same condition with CI casually.
That's what I intended for suggested the `rake` tasks.

> We may run one replica set and one sharded cluster setup in Travis, or may not.

For example, I can suggest below flow to run all the cases by CI.

```
rake spec:mongod:standalone:start
rake spec:ci
rake spec:mongo:standalone:stop

rake spec:mongo:replicaset:start
rake spec:ci
rake spec:mongo:replicaset:stop

rake spec:mongo:shard:start
rake spec:ci
rake spec:mongo:shard:stop
```

p-mongo:
The test running guidance covering configuring the server deployment is in https://github.com/mongodb/mongo-ruby-driver/pull/1295.

PR #996, opened 2018-07-24
Build ruby-head on evergreen

p-mongo:


.evergreen/run-tests.sh
```diff
@@ -34,33 +34,44 @@ export DRIVER_TOOLS_CA_PEM="${DRIVERS_TOOLS}/.evergreen/x509gen/ca.pem"
...
-set -o errexit
-
-# Ensure we're using the right ruby
-python - <<EOH
+  # 12.04, 14.04 and 16.04 are good
+  wget -O ruby-head.tar.bz2 http://rubies.travis-ci.org/ubuntu/`lsb_release -rs`/x86_64/ruby-head.tar.bz2
+  tar xf ruby-head.tar.bz2
+  export PATH=`pwd`/ruby-head/bin:`pwd`/ruby-head/lib/ruby/gems/2.6.0/bin:$PATH
+  ruby --version
+  ruby --version |grep dev
```
saghm: Just to make sure I understand, is the purpose of this line just to exit the script if `dev` is not found?

p-mongo: That is correct.

PR #997, opened 2018-07-26
don't use @flag_bits given that it's never initializated

OgAstorga:
Given that @flag_bits is never initializated.

(@flag_bits·&·Msg::FLAGS.index(:checksum_present)) will always equal to false
also
(@flag_bits·&·Msg::FLAGS.index(:checksum_present)) == 1 will always equal to false
and finally 
(@flag_bits·&·Msg::FLAGS.index(:checksum_present)) == 1 ? 32 : 0 will always equal to 0

So it's safe to assume that end_length will always be equal to 0

### DIFF
--- a/lib/mongo/protocol/serializers.rb
+++ b/lib/mongo/protocol/serializers.rb
@@ -187,7 +187,7 @@ module Mongo
         #
         # @since 2.5.0
         def self.deserialize(buffer)
-          end_length = (@flag_bits & Msg::FLAGS.index(:checksum_present)) == 1 ? 32 : 0
+          end_length = 0
           sections = []
           until buffer.length == end_length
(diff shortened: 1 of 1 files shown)


p-mongo:
That is called `@flags` apparently, see 5865ef85e.

p-mongo:
https://jira.mongodb.org/browse/RUBY-1542

p-mongo:
We currently have a partial support for checksums implemented. In particular, when checksum flag bits is written into the message the check sum itself should also be written, but this doesn't happen. The method modified in this PR reads an instance variable but it's a class method, hence an instance variable in it won't ever be set to anything. The best course of action here is to leave the code as is until we fully implement checksum support in the driver.

PR #1000, opened 2018-07-31
 Sdam diagnostics

p-mongo:
This primarily adds the summary method to sdam-related objects. It is marked experimental because it may change as we work on sdam.

As https://jira.mongodb.org/browse/RUBY-1443 involves file and class splits I would like to get this diff into master now.

### DIFF
--- a/lib/mongo/error/no_server_available.rb
+++ b/lib/mongo/error/no_server_available.rb
@@ -27,9 +27,11 @@ module Mongo
       #
       # @param [ Hash ] server_selector The server preference that could not be
       #   satisfied.
+      # @param [ Cluster ] cluster The cluster that server selection was
+      #   performed on. (added in 2.7.0)
       #
       # @since 2.0.0
-      def initialize(server_selector)
+      def initialize(server_selector, cluster=nil)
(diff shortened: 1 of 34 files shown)


p-mongo:
https://gist.github.com/p-mongo/69b1b02ab9a09b752be804490c54b830

lib/mongo/error/no_server_available.rb
```diff
@@ -27,9 +27,11 @@ class NoServerAvailable < Error
       #
       # @param [ Hash ] server_selector The server preference that could not be
       #   satisfied.
+      # @param [ Cluster ] cluster The cluster that server selection was
+      #   performed on. (added in 2.7.0)
       #
       # @since 2.0.0
-      def initialize(server_selector)
+      def initialize(server_selector, cluster=nil)
```
saghm: Ditto question on #1011; why is the cluster parameter here?

p-mongo: Must've gotten killed off.

p-mongo:
Diagnostics when trying to connect with RS name mismatch: https://gist.github.com/p-mongo/bdcb99c45bf2025332d668a4fd9f4370

PR #1001, opened 2018-08-02
 Parallel tests prep

p-mongo:


p-mongo:
This PR makes most tests self-setting up. Previously many tests relied on other tests for either setting up the database (for example, a test would use or drop a collection without ensuring it existed) or cleaning up the database (for example, a test would fail if another test created an index and left that index in the database).

lib/mongo/options/mapper.rb
```diff
@@ -24,20 +24,34 @@ module Mapper
...
       # @param [ Hash ] options The options to transform
       # @param [ Hash ] mappings The key mappings.
+      # @param [ true|false ] drop_extra Whether to ignore options not present
+      #   in mappings
       #
       # @return [ Hash ] The transformed options.
       #
       # @since 2.0.0
-      def transform(options, mappings)
+      def transform(options, mappings, drop_extra=true)
```
saghm: Would it be cleaner to implement this as `keep_extra = false` so that passing nil will have the same behavior as not passing an argument?

spec/mongo/collection_spec.rb
```diff
@@ -2,8 +2,15 @@
 
 describe Mongo::Collection do
 
-  after do
-    authorized_collection.delete_many
+  before do
+    begin
+      authorized_collection.delete_many
+    rescue Mongo::Error::OperationFailure
```
saghm: Just to make sure I understand, is the reason for catching errors here (and below) that we can't delete things from the collection while another thread is using it?

p-mongo: Some collections/indexes cannot be removed when they don't exist, I have not bothered to investigate why yet.

lib/mongo/options/mapper.rb
p-mongo: It's possible but the default going forward (3.0+) is to not drop extra keys.

spec/mongo/collection_spec.rb
p-mongo: I changed this to drop collection and a number of tests failed.

p-mongo:
https://jira.mongodb.org/browse/RUBY-1458 for options mapper change

PR #1003, opened 2018-08-06
Don't remove unknown secondaries servers from the topology

saghm:


saghm:
I've added comments describing the SDAM spec behavior that those methods are following.

lib/mongo/cluster/topology/replica_set.rb
```diff
@@ -213,6 +213,12 @@ def remove_hosts?(description)
         end
 
         # Whether a specific server in the cluster can be removed, given a description.
+				# As described in the SDAM spec, a server should be removed if the server's
```
behackett: What happened with the whitespace here?

saghm: Oops, looks like I need to configure tabs vs. spaces properly on my personal laptop

PR #1004, opened 2018-08-06
RUBY-1378 Fix use documentation to match reality

p-mongo:


p-mongo:
@saghm Made the changes.

PR #1005, opened 2018-08-06
RUBY-1414 Transaction example in Ruby driver tutorial

p-mongo:


docs/tutorials/ruby-driver-transactions.txt
```diff
@@ -0,0 +1,73 @@
...
+
+.. code-block:: ruby
+
+  session = client.start_session
+  session.start_transaction(
+    read_concern: {level: :majority},
+    write_concern: {w: 3},
+    read: {mode: :primary})
+
+A transaction must be committed or aborted. The corresponding methods
```
saghm: > A transaction must be committed or aborted

This isn't technically true; when a session ends (either because the user explicitly ends it or the session is garbage collected), any transaction in progress will be implicitly aborted. I think it's better to phrase this as "may be committed or aborted" and then to move the sentence below about implicit aborts here.

docs/tutorials/ruby-driver-transactions.txt
```diff
@@ -0,0 +1,73 @@
...
+  session.abort_transaction
+
+If a session ends with an open transaction,
+`the transaction is aborted <https://docs.mongodb.com/manual/core/transactions/#transactions-and-sessions>`_.
+
+
+Retrying Commits
+----------------
+
+Transaction commit `can be retried
```
saghm: "The transaction commit"

p-mongo:
@saghm Made the changes.

PR #1007, opened 2018-08-14
RUBY-1375 Per-test timeout

p-mongo:


p-mongo:
Set timeout to 30 seconds as 10 seconds causes failures. This still deals with runaway tests that never complete.

spec/README.md
```diff
@@ -0,0 +1,85 @@
...
+configured and started externally to the test suite. This allows
+running the entire test suite against, for example, a standalone
+mongod as well as a replica set. The flip side to this is the
+test suite will not work without a running mongo cluster, and
+tests which are not applicable to or cannot be performed on the
+running mongo cluster are skipped.
+
+Not only does the test suite require an externally launched cluster,
+the test suite must also be told how the cluster is configured
+via MONGODB_URI, TOPOLOGY, MONGODB_ADDRESSES, RS_ENABLED, RS_NAME and
```
saghm: I think it's worth documenting that the user needs to set _either_ `MONGODB_URI` or a valid set of `MONGODB_ADDRESSES`, `RS_ENABLED`, `RS_NAME`, and `SHARDED_ENABLED`.

spec/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+Additionally some of the tests assume that the seed list (given in
+MONGODB_URI or MONGODB_ADDRESSES) encompasses all servers in the cluster,
+and will fail when MONGODB_URI includes only one host of a replica set.
+It is best to include all hosts of the cluster in MONGODB_URI and
+MONGODB_ADDRESSES.
+
+The test suite seems to have issues connecting to a replica set
+via IP addresses if the replica set hosts are defined with hostnames
+(i.e., 127.0.0.1 vs localhost). Try to exactly match the contents of
```
saghm: I think this issue is dependent on correctly setting `localhost` in the `/etc/hosts` file; have you had issues using `localhost` even after setting that up?

spec/README.md
p-mongo: Not true - they don't need to set any of them.

spec/README.md
p-mongo: Every linux machine has localhost in /etc/hosts, so yes.

saghm: I've never had any issues with `localhost` not working before; is this something you've experienced only locally? I'm a little hesitant to add this to the documentation when it seems like it might just be a configuration issue (which we can document how to fix if we can diagnose it)

spec/README.md
saghm: Apologies, I worded that poorly; my point is that if a user sets `MONGODB_URI`, they don't need to set the others (and vice-versa), and using "and" here is a bit misleading.

spec/README.md
p-mongo: Well, this comment does document this issue, which other documentation do you have in mind?

The driver does not know that 127.0.0.1 and localhost are the same thing and can remove either one from its list of addresses depending on what the cluster returns in ismaster.

saghm: It's my understanding that the driver does DNS lookups (through `getaddrinfo`), which would mean it can resolve localhost to 127.0.0.1 assuming the hosts file is correctly configured. I also feel that the documentation in the current state is not accurate; I've never seen an issue due to defining a replica set name with hostnames (except in the absence of a localhost entry in /etc/hosts), which is why I'm trying to determine in what circumstances you've seen these issues.

spec/README.md
p-mongo: Also not true, and this discussion is outside the scope of this PR anyway.

spec/README.md
p-mongo: DNS resolution and server removal due to me mismatch are two different things, and again, this discussion is outside the scope of this PR.

spec/README.md
p-mongo: Have it your way.

PR #1011, opened 2018-08-21
Sync sdam spec tests

p-mongo:
https://jira.mongodb.org/browse/RUBY-1332


p-mongo:
https://github.com/mongodb/specifications/pull/388 for the spec change

lib/mongo/error/no_server_available.rb
```diff
@@ -27,9 +27,11 @@ class NoServerAvailable < Error
       #
       # @param [ Hash ] server_selector The server preference that could not be
       #   satisfied.
+      # @param [ Cluster ] cluster The cluster that server selection was
+      #   performed on. (added in 2.7.0)
       #
       # @since 2.0.0
-      def initialize(server_selector)
+      def initialize(server_selector, cluster=nil)
```
saghm: It doesn't look like cluster is used at all here; what's it for?

p-mongo:
I took out the test changes, with this PR the driver passes spec tests as they are written.

saghm:
Nice work!

PR #1012, opened 2018-08-22
RUBY-1406 Set up junit xml-ish test reports in evergreen

p-mongo:


p-mongo:
This adds test results in evergreen ui in the task view as well as junit xml in artifacts. The evergreen ui addition halts the ui to a crawl in firefox as the frontend is trying to render all the tests we have which is a lot of tests. I might take this out if it proves too annoying or if anyone else feels it's annoying.

PR #1017, opened 2018-08-28
Remove addresses from cluster and takem off servers

p-mongo:


lib/mongo/server.rb
```diff
@@ -173,12 +173,27 @@ def initialize(address, cluster, monitoring, event_listeners, options = {})
       @address = address
       @cluster = cluster
       @monitoring = monitoring
+      options = options.dup
+      monitor = options.delete(:monitor)
       @options = options.freeze
+      @monitor = Monitor.new(address, event_listeners, options.merge(
+        app_metadata: cluster.app_metadata))
+      unless monitor == false
```
saghm: Similar to the Mongoid pull request from the other day, can we document that `monitor: nil` in the options is different than `monitor: false`?

p-mongo: This option is not documented because it is not meant for public consumption.

p-mongo:
Which typo are you referring to?

saghm:
Apologies, I was looking at the pull request title, not the commit message

PR #1018, opened 2018-08-28
Reduce driver matrix size

p-mongo:


p-mongo:
Matrix size was reduced by 2/3, how did you arrive at your conclusion?

PR #1020, opened 2018-09-06
RUBY-1443 Separate ReplicaSetNoPrimary & ReplicaSetWithPrimary

p-mongo:


lib/mongo/cluster/topology.rb
```diff
@@ -35,14 +29,26 @@ class Cluster
     # @since 2.0.0
     module Topology
       extend self
+    end
+  end
+end
 
+require 'mongo/cluster/topology/replica_set_no_primary'
```
saghm: What's the purpose of moving the require statements to the middle of the file?

lib/mongo/cluster/topology/replica_set_with_primary.rb
```diff
@@ -0,0 +1,27 @@
...
+
+module Mongo
+  class Cluster
+    module Topology
+
+      # Defines behavior when a cluster is in replica set topology,
+      # and is a no primary which has been discovered by the driver.
+      #
+      # @since 2.7.0
+      class ReplicaSetWithPrimary < ReplicaSetNoPrimary
```
saghm: Just to make sure I understand, are we planning on implementing any behavioral differences between ReplicaSetWithPrimary or ReplicaSetNoPrimary, or is the purpose of the separate classes just to make monitoring easier?

spec/spec_tests/data/sdam/rs/secondary_wrong_set_name_with_primary_second.yml
```diff
@@ -0,0 +1,73 @@
+description: "Secondary wrong setName with primary when secondary responds first"
```
saghm: Has this been merged into the specifications repo yet?

lib/mongo/cluster/topology.rb
p-mongo: I'd move them to the bottom but that does not work.

lib/mongo/cluster/topology/replica_set_with_primary.rb
p-mongo: The purpose right now is spec compliance which mandates the difference between the two. Depending on how much work/refactoring is done in the other tickets there may be behavior differences. For example, there shouldn't be a need to query servers for primary to find out if a topology has a primary.

spec/spec_tests/data/sdam/rs/secondary_wrong_set_name_with_primary_second.yml
p-mongo: This change is not wanted in specs, see my PR there.

lib/mongo/cluster/topology.rb
saghm: Sounds good! Thanks for explaining

lib/mongo/cluster/topology/replica_set_with_primary.rb
saghm: Makes sense

spec/spec_tests/data/sdam/rs/secondary_wrong_set_name_with_primary_second.yml
saghm: Ah, I see Jesse's comment now. Thanks for pointing me to it!

PR #1022, opened 2018-09-07
Fix RUBY-1357 Client constructor does I/O contrary to spec requirements

p-mongo:


p-mongo:
depends: ruby-1353

p-mongo:
https://jira.mongodb.org/browse/RUBY-1458 for options mapper change

p-mongo:
Create a 3.0 ticket to change the default.

lib/mongo/cluster.rb
```diff
@@ -118,6 +125,34 @@ def initialize(seeds, monitoring, options = Options::Redacted.new)
 
       @connecting = false
       @connected = true
+
+      if options[:scan] != false
```
saghm: I'm a little confused about this diff; it looks like this is only adding things in the case that the option is true; wouldn't that mean that the option being false would lead to the same behavior as before (which is still doing I/O)?

saghm: Oh, never mind, I missed the deletion of `monitor.scan!` below

spec/mongo/client_construction_spec.rb
```diff
@@ -55,6 +58,37 @@
...
+
+    context 'with default scan: true' do
+      # TODO this test requires there being no outstanding background
+      # monitoring threads running, as otherwise the scan! expectation
+      # can be executed on a thread that belongs to one of the global
+      # clients for instance
+      it 'performs one round of sdam' do
+        # Does not work due to
+        # https://github.com/rspec/rspec-mocks/issues/1242.
+        #expect_any_instance_of(Mongo::Server::Monitor).to receive(:scan!).
```
saghm: It looks like the issue is closed and they don't have any intention of fixing it. Given that it doesn't seem like there's any expectation of being able to use this, can we delete it?

p-mongo: If I keep running into this I will look into fixing it myself.

saghm: Sounds good

PR #1024, opened 2018-09-11
RUBY-1454 Global registry of clients in test suite

p-mongo:


p-mongo:
This adds the registry and local client cleanup after each example. Dealing with global clients is somewhat of a rabbit hole and will come separately.

spec/integration/retryable_writes_spec.rb
```diff
@@ -3,7 +3,7 @@
 describe 'Retryable writes integration tests' do
   include PrimarySocket
 
-  after do
+  before do
```
saghm: I think the idea behind having it be `after` was that we would leave the database in a pristine state after the tests ended. I'm guessing we don't do that consistently across our tests right now, but do you have any objections to using `around` here instead?

p-mongo: Each test is responsible for setting the environment up for itself correctly. If this is not done, a test may fail because another test ran or did not run (which is exactly what happens in a parallel run - the set of tests run on a particular database instance is no longer fixed).

Performing cleanup in after hooks results in setup being run twice per test, which increases test runtime for no benefit.

saghm: Fair enough

PR #1027, opened 2018-09-13
RUBY-1499 Create a topology base class

p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-1499

mbroadst:
@p-mongo the benefit of the SDAM topology/topology_description/server_description design is that you can model connections to replica sets, sharded deployments, and single servers with the same underlaying machinery. Might it be a better idea to refactor towards unifying all the disparate topology types here into a single `Topology` type, which simply provides a `TopologyType` rather than expressing that through inheritance?

p-mongo:
This PR does unify the behavior of different topology types, specifically with respect to addresses (which some topologies stored and some did not). As such it is a step toward having a single topology class eventually.

PR #1029, opened 2018-09-14
Change stream docs improvements

p-mongo:
https://jira.mongodb.org/browse/RUBY-1422
https://jira.mongodb.org/browse/RUBY-1457


docs/tutorials/ruby-driver-collection-tasks.txt
```diff
@@ -101,53 +101,3 @@ documents must contain an ``age`` field which is a number.
                    { '$type' => "number" }
                  }
              })
-
-
-Change Streams
```
saghm: What do you think of putting a link here to the change stream documentation (instead of removing it) so that existing links people have to this section (using a hashref) will still give them some info instead just going to the top of the page and not having any info about change streams?

p-mongo: Okay will do that.

p-mongo:
@saghm Added the change stream link.

PR #1032, opened 2018-09-17
Spec config expansion

p-mongo:
This moves some of the global constants into spec config, for use in the task that will create test users (https://jira.mongodb.org/browse/RUBY-1461).


p-mongo:
@saghm Ready now.

PR #1033, opened 2018-09-17
RUBY-1461 Separate test user creation from test suite

p-mongo:


spec/support/spec_setup.rb
```diff
@@ -0,0 +1,71 @@
+require_relative './spec_config'
+
+class SpecSetup
+  def run
+    begin
+      create_user(admin_unauthorized_client, SpecConfig.instance.root_user)
+    rescue Mongo::Error::OperationFailure => e
```
saghm: This is awesome! Throwing away all the errors like we were doing before when creating users was definitely not a good solution; checking them to make sure we don't throw away meaningful errors is super helpful

PR #1036, opened 2018-09-20
Client registry #with handling

p-mongo:
Last part of https://jira.mongodb.org/browse/RUBY-1454.

### DIFF
--- a/spec/support/client_registry.rb
+++ b/spec/support/client_registry.rb
@@ -1,5 +1,40 @@
 require 'singleton'
 
+module Mongo
+  class Client
+    alias :with_without_registry :with
+    def with(*args)
+      with_without_registry(*args).tap do |client|
+        ClientRegistry.instance.register_local_client(client)
+      end
+    end
+  end
+end
+
+# Test suite uses a number of global clients (with lifetimes spanning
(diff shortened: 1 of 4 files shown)


spec/support/client_registry.rb
```diff
@@ -1,5 +1,40 @@
 require 'singleton'
 
+module Mongo
+  class Client
+    alias :with_without_registry :with
```
saghm: I'm having a little trouble understanding what this does; it looks like the call to `with_without_registry` is recursive (due to the alias above), but that would make the stack overflow, so I assume that's not what it's doing. Any chance you could add a comment to explain what's going on here?

p-mongo: This is standard Ruby functionality. https://www.thoughtco.com/aliasing-in-ruby-2908190 has a decent explanation. `alias` effectively copies the method so that original method can be redefined.

p-mongo: Also see https://apidock.com/rails/Module/alias_method_chain which is a more user-friendly version of the same pattern that Rails provides.

saghm: Ah, I got the directionality of `alias` backwards

PR #1037, opened 2018-09-23
Jruby SNI support

p-mongo:
https://jira.mongodb.org/browse/RUBY-1470


p-mongo:
It is, the fix is 1 line of code.

PR #1040, opened 2018-09-24
RUBY-1319 Handshake changes for SDAM and Auth

p-mongo:
Part 1 - omit scram bits from ismaster on monitoring sockets


p-mongo:
This removes scram bits from ismaster request on monitoring sockets.

PR #1041, opened 2018-09-24
Lock arrays in event subscriber to make jruby happy, which accesses t…

p-mongo:
…hem concurrently in different threads


p-mongo:
It is hard to tell whether the new jruby failures are a result of this PR or simply exposed by this PR fixing something, but in any event I will deal with those later.

PR #1044, opened 2018-09-25
RUBY-1319 Handshake changes for SDAM and Auth

p-mongo:


p-mongo:
This is part 2 which disconnects all non-monitoring sockets to a server upon an auth failure during a handshake. Sockets which are checked out of the connection pool are disconnected when they are returned to the pool. Generations are implemented to facilitate this following their implementation in Java driver.

p-mongo:
This is part 2 which disconnects all non-monitoring sockets to a server upon an auth failure during a handshake. Sockets which are checked out of the connection pool are disconnected when they are returned to the pool. Generations are implemented to facilitate this following their implementation in Java driver.

PR #1048, opened 2018-09-28
Always check for a null session.

jonhyman:
While upgrading a server from MongoDB 3.4 to MongoDB 3.6, we got a few thousand application errors for `undefined method `in_transaction?' for nil:NilClass` which persisted until application restart.

The labeling needs to be more defensive that a session could be null.

image

### DIFF
--- a/lib/mongo/operation/aggregate/op_msg.rb
+++ b/lib/mongo/operation/aggregate/op_msg.rb
@@ -42,7 +42,7 @@ module Mongo
           process_result(result, server)
           result.validate!
         rescue Mongo::Error::SocketError => e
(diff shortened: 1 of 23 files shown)


saghm:
Hi @jonhyman! I'm actually having a little trouble reproducing this error locally. Can you describe the topology that the driver was connected to when this happened and which node was replaced?

saghm:
Never mind, I just reproduced it. Sorry for the confusion!

p-mongo:
Superseded byhttps://github.com/mongodb/mongo-ruby-driver/pull/1222.

PR #1049, opened 2018-09-28
Test on auth & nossl and noauth & ssl to diagnose the auth/ssl failures

p-mongo:


p-mongo:
Looks like #1044 is succeeding on auth/nossl and failing on noauth/ssl, making this PR valuable already.

saghm:
Not testing on those configurations isn't an oversight; it's done on purpose specifically do cut down on the size of our evergreen runs. The Python driver does the same thing as well; this was actually something originally suggested by Bernie.

I'm willing to add these variants, but only if we set them not to run by default on Github pull request jobs (and we make an effort not to run them unless necessary on manual jobs). I don't think that running twice as many evergreen tasks on ever single run that we do is worth it to be able to differentiate between auth and ssl errors on the first run; if we run into an issue where we're not sure which is the cause, we can always run a manual job that just does those tasks.

p-mongo:
This PR adds 6 configurations to the 80 that we otherwise have, down from 200+ some time ago.

p-mongo:
https://jira.mongodb.org/browse/RUBY-1480

PR #1051, opened 2018-10-01
RUBY-1311 Test kerberos in evergreen

saghm:


p-mongo:
Can we pick one of "kerberos", "gssapi" and "enterprise auth" and use that everywhere?

saghm:
I've eliminated usage of "GSSAPI" in favor of kerberos.

The reason for the distinction between "enterprise auth" and "kerberos" is that we'll likely add LDAP tests at some point, and we'll also want to run those in the enterprise auth variant/task. I've renamed a couple of things in the evergreen config and the spec repo to try to clear up the distinction between them.

spec/enterprise_auth/kerberos_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  let(:user) do
+   "#{ENV['SASL_USER']}%40#{ENV['SASL_HOST'].upcase}"
+  end
+
+  let(:host) do
+    "#{ENV['SASL_HOST']}"
+  end
+
+  let(:gssapi_db) do
+    "#{ENV['GSSAPI_DB']}"
```
p-mongo: What about this env var?

saghm:
Currently, the names of the environmental variables are just copied from the names of the Evergreen project variables, and I think it makes sense to keep them the same as each other. I guess we can change those names in the Evergreen project if you'd prefer that.

saghm:
Sounds good; I'll change our evergreen config variables to use "kerberos" instead of "gssapi" and then update the code accordingly

saghm:
Everything good to merge now?

p-mongo:
:+1:

saghm:
Awesome, thanks!

PR #1054, opened 2018-10-02
RUBY-1269 Change handling of network errors or timeouts during connection handshake

p-mongo:


p-mongo:
This brings handling of exceptions during authentication into compliance with the spec, handshake exceptions will be done in a follow-up PR.

PR #1058, opened 2018-10-04
RUBY-1222 Use IP address for creating sockets and stop caching hostna…

saghm:
…me resolution

This pull request fixes two bugs in the hostname resolution/socket creation functionality of the driver. The first is that sockets are incorrectly being created with the hostname rather than the IP address that was obtained with `getaddrinfo`. This breaks the ability to fallback to IPv4 in the case that connecting to the IPv6 address fails. The second bug is the caching of the resolver in the address; doing a DNS lookup each time a socket is created ensures that the driver will detect any changes to where the hostname resolves to without requiring the Client to be recreated.

### DIFF
--- a/lib/mongo/address.rb
+++ b/lib/mongo/address.rb
@@ -136,6 +136,12 @@ module Mongo
 
     # Get a socket for the provided address, given the options.
     #
+    # The address the socket connects to is determined by the algorithm described in the
+    # #intialize_resolver! documentation. Each time this method is called, #initialize_resolver!
+    # will be called, meaning that a new hostname lookup will occur. This is done so that any
+    # changes to which addresses the hostname resolves to will be picked up even if a socket has
+    # been connected to it before.
+    #
     # @example Get a socket.
     #   address.socket(5, :ssl => true)
     #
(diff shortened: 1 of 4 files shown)


jarthod:
Looks good to me :+1: wouldn't this deserve some tests to avoid future another regression?

lib/mongo/address.rb
```diff
@@ -146,7 +152,7 @@ def inspect
     #
     # @since 2.0.0
     def socket(socket_timeout, ssl_options = {})
-      @resolver ||= initialize_resolver!(ssl_options)
+      @resolver = initialize_resolver!(ssl_options)
       @resolver.socket(socket_timeout, ssl_options)
```
p-mongo: Since the resolver is now local, it should no longer be an instance variable (remove `@`).

saghm: Hmm, looks like some tests fail when I do that. I'll look into what it takes to fix them.

saghm:
Hi @jarthod! Sorry for the delay in the response about testing. It looks like the mock test that existed to test this fallback only tested that the socket was able to make a connection in the case that both IPv6 and IPv4 were returned by `getaddrinfo`; I've changed this to assert that the socket actually resolves to an IPv4 address (see [this commit](https://github.com/mongodb/mongo-ruby-driver/pull/1058/commits/c951d40be49a679ade3246c6526e9f548a50d80b)). The test fails on master but passes on the branch for this PR, so I think this should be sufficient to detect any failures in the future.

saghm:
@p-mongo Sorry for all the work on this after the initial LGTM! I think this has everything it should need now, and it's passing with the modified tests, so only one more look-over needed

PR #1063, opened 2018-10-05
RUBY-1265 Test that killCursors really succeeds

p-mongo:


spec/integration/cursor_reaping_spec.rb
```diff
@@ -32,22 +26,71 @@
...
+      expect(started_event).not_to be_nil
+
+      succeeded_event = EventSubscriber.succeeded_events.detect do |event|
+        event.command_name == 'killCursors' && event.request_id == started_event.request_id
+      end
+
+      expect(succeeded_event).not_to be_nil
+
+      # ok is integer on older versions, float on newer versions
+      expect(succeeded_event.reply['ok'].to_i).to be 1
```
saghm: Not a huge deal, but I think `expect(succeeded_event.reply['ok']).to eq(1)` might be a bit cleaner than casting the type

p-mongo: Made the change, thanks.

PR #1064, opened 2018-10-05
minor: test kerberos with jruby

saghm:


saghm:
Given that this passes the Enterprise Auth JRuby variant, I think it should be okay to merge despite the other failures, since they're the same ones that sporadically pop up on other PRs.

PR #1065, opened 2018-10-08
RUBY-1269 Change handling of network errors or timeouts during connection handshake

p-mongo:


p-mongo:
Part 2 - marking connection unknown when handshake fails.

spec/mongo/server/connection_spec.rb
```diff
@@ -74,16 +100,16 @@ class ConnectionSpecTestException < Exception; end
...
 
         it 'clears socket' do
-          result
+          error
           expect(connection.send(:socket)).to be nil
         end
 
         it 'attempts to reconnect after failure when asked' do
-          # for some reason referencing result here instead of
+          # for some reason referencing error here instead of
```
saghm: I think it might be due to laziness? I think the `let` block is only evaluated once and then uses the previous result for subsequent evaluations. Since `authenticate!` was already called in the tests above, it isn't called this time.

p-mongo: I think I only used it once but maybe not, I'll keep this in mind for the future.

PR #1068, opened 2018-10-09
RUBY-1493 Implement topology closed sdam event

p-mongo:


lib/mongo/cluster.rb
```diff
@@ -431,7 +447,10 @@ def disconnect!
     # @return [ true ] Always true.
     #
     # @since 2.1.0
+    # @deprecated Use Client#reconnect to reconnect to the cluster instead of
+    #   calling this method. This method does not send SDAM events.
     def reconnect!
+      @connecting = true
```
saghm: Should this to set to false at the end of the method? It seems a little inconsistent with the way `#initialize` sets it to false to leave it true after this method; otherwise, it's possible for it to be both "connecting" and "connected", which seems incorrect.

p-mongo: Sure, fixed.

PR #1069, opened 2018-10-09
RUBY-1316 Add test for canonicalize_host_name with kerberos auth

saghm:


spec/enterprise_auth/kerberos_spec.rb
```diff
@@ -39,4 +39,19 @@
     expect(doc['kerberos']).to eq(true)
     expect(doc['authenticated']).to eq('yeah')
   end
+
+  context 'when canonicalizeHostName is true' do
```
p-mongo: How come the setting is in camel case here and in all caps in the URI string?

saghm: Good point; I wrote the test before I looked up the syntax for providing the option in the driver, but I should change it to match.

PR #1081, opened 2018-10-11
Memoize a variable that could change by a separate thread.

jonhyman:
I assume that what happened here was that `@cluster.logical_session_timeout` changed in the middle of the execution, since it comes from an isMaster command. I believe we had a server failure at or around this time.

image


p-mongo:
Thanks!

https://jira.mongodb.org/browse/RUBY-1507?filter=26037&jql

PR #1085, opened 2018-10-12
Fix shortening of fixture file paths in sdam spec

p-mongo:


p-mongo:
Needed since fixtures were moved under data subdir some time ago.

PR #1095, opened 2018-10-15
Drop inspectors

p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-1535

PR #1102, opened 2018-10-25
Test enterprise auth on older distros

p-mongo:


p-mongo:
Segfaults are tracked in https://jira.mongodb.org/browse/RUBY-1534.

PR #1104, opened 2018-10-27
RUBY-1539 Extract sdam flow into own class

p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-1517

The next PR will replace extracted flow with a spec compliant rewrite.

PR #1105, opened 2018-10-29
RUBY-1517 Rewrite topology type state transitions to be spec compliant

p-mongo:


p-mongo:
I thought I had a run which didn't have any of the tests get stuck, but maybe not.

https://github.com/mongodb/mongo-ruby-driver/pull/1107 is meant to improve reliability of SDAM spec tests specifically but I suspect the "no IO" provision could also be used by some of the other tests.

p-mongo:
#1111 fixed the slowdown, #1109 and #1110 also contributed to retaining performance here.

PR #1106, opened 2018-10-29
RUBY-1538 Make server description query methods spec compliant

p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-1538

PR #1107, opened 2018-10-29
RUBY-1543 Optionally completely remove monitoring I/O from servers & clusters

p-mongo:


lib/mongo/cluster.rb
```diff
@@ -80,6 +80,11 @@ class Cluster
     #   this option to nil enables scanning seeds in constructor in driver
     #   version 2.x. Driver version 3.x will recognize this option but
     #   will ignore it and will never scan seeds in the constructor.
+    # @option options [ true, false ] :monitoring_io For internal driver
```
saghm: Is it worth monkey-patching this into Client in the tests rather than exposing it as part of a public API? Even with the documentation, I'm not convinced that we want to expose this to users. I don't feel strongly enough about this to block it from being merged, though

p-mongo: I read the API docs when I need to know something. This documentation is not intended for users, it's just the best place I could find for it.

p-mongo: Regarding patching this functionality in, I don't see how this can be done in a reasonable way.

saghm: > I read the API docs when I need to know something. This documentation is not intended for users, it's just the best place I could find for it.

Fair enough; it's definitely worth having documentation for us and future developers as well

PR #1109, opened 2018-10-29
Speed up tests by omitting monitoring io where possible

p-mongo:


spec/mongo/client_construction_spec.rb
```diff
@@ -12,7 +12,7 @@
           :primary, :primary_preferred, :secondary, :secondary_preferred,  :nearest
         ].each do |sym|
           it "accepts #{sym} as symbol" do
-            client = new_local_client(['127.0.0.1:27017'],
+            client = new_local_client_nmio(['127.0.0.1:27017'],
```
saghm: I'm not super fond of the name `new_local_client_nmio`, since the meaning of the name isn't clear if you don't already know what it does, but I don't feel strongly enough to block merging on changing it

PR #1112, opened 2018-11-01
RUBY-1548 Spec tests for server addition/removal events and sdam event sequencing

p-mongo:


saghm:
In general, I'm not super fond of adding spec tests before they're merged into the upstream repo, since they're subject to change during review. Can we do the pull request to the specs repo first, or is this heavily blocking other work that you need to do?

p-mongo:
I made https://github.com/mongodb/specifications/pull/410, in the mean time these tests are important for ongoing sdam work therefore I would like to merge them. If https://github.com/mongodb/specifications/pull/410 produces any changes I will update the tests here.

PR #1115, opened 2018-11-02
RUBY-1537 Rename replica_set -> replica_set_name in topology ctor options

p-mongo:


lib/mongo/cluster/topology.rb
```diff
@@ -66,13 +66,25 @@ module Topology
...
+          Unknown
         end
+        # Options here are client/cluster/server options.
+        # In particular the replica set name key is different for
+        # topology.
+        # If replica_set_name is given (as might be internally by driver),
+        # use that key.
+        # Otherwise (e.g. options passed down from client),
+        # move replica_set to replica_set_name.
+        if [ReplicaSetNoPrimary, ReplicaSetWithPrimary].include?(cls) && !options[:replica_set_name]
```
saghm: Minor nit, but maybe the topology type classes could have an `is_replica_set?` class method?

p-mongo:
Replica set and replica set name are different things. This PR fixes the naming for topologies where this change is not part of public API. Eventually everything up to clients should refer to replica set names as `replica_set_name`.

PR #1118, opened 2018-11-03
Fix RUBY-1536 Server & topology change events are dispatched out of order

p-mongo:


p-mongo:
This PR goes with the tests in #1112.

lib/mongo/cluster/sdam_flow.rb
```diff
@@ -284,20 +296,31 @@ def update_rs_without_primary
...
+        server.start_monitoring
+      end
     end
 
     # Adds all servers referenced in the given description (which is
     # supposed to have come from a good primary) which are not
     # already in the topology, to the topology.
+    #
+    # @note Servers are added unmonitored. Monitoring must be started later
+    # separately.
```
saghm: Can we document the return value here?

lib/mongo/cluster/sdam_flow.rb
```diff
@@ -356,7 +368,7 @@ def commit_changes
       # (which is important for spec tests because they have real i/o
       # happening against bogus addresses which yield unknown responses
       # before the responses are mocked).
-      unless updated_desc.unknown? && previous_desc.unknown?
+      unless updated_desc.unknown? && previous_desc.unknown? || updated_desc.object_id == previous_desc.object_id
```
saghm: Using `unless` along with both an `&&` and an `||` makes this a bit complex to think through (especially if the reader doesn't remember which of `&&` and `||` has higher precedence; can we refactor this to be a bit easier to read? I think something like this would be easier for people to read:

```ruby
if updated_desc.unknown? && previous_desc.unknown? 
  return
end

if updated_desc.object_id == previous_desc.object_id
  return
end

publish_sdam_event(...) # etc.
```

lib/mongo/cluster/sdam_flow.rb
```diff
@@ -366,20 +378,39 @@ def commit_changes
...
         topology_changed_event_published = true
+        @need_topology_changed_event = false
       end
 
       # If a server description changed, topology description change event
       # must be published with the previous and next topologies being of
       # the same type, unless we already published topology change event
       unless topology_changed_event_published
-        unless updated_desc.unknown? && previous_desc.unknown?
+        unless updated_desc.unknown? && previous_desc.unknown? || updated_desc.object_id == previous_desc.object_id
```
saghm: Ditto above

PR #1121, opened 2018-11-05
Fix RUBY-1442 Topology(Description) has no server knowledge

p-mongo:


lib/mongo/cluster/topology/server_description_list.rb
```diff
@@ -0,0 +1,56 @@
...
+module Mongo
+  class Cluster
+    module Topology
+      # A list of server descriptions. Stored in a topology.
+      #
+      # This class adds methods for retrieving server descriptions
+      # by their address.
+      #
+      # @since 2.7.0
+      class ServerDescriptionList < Array
```
saghm: Is there a reason we don't use use a hash of addresses to server descriptions? Most other SDAM implementations I've seen use this pattern

p-mongo: This implementation follows the spec tests which use a list of server descriptions. Which drivers use hashes?

lib/mongo/cluster.rb
```diff
@@ -120,10 +120,14 @@ def initialize(seeds, monitoring, options = Options::Redacted.new)
         add(seed, monitor: false)
       end
 
-      publish_sdam_event(
-        Monitoring::TOPOLOGY_CHANGED,
-        Monitoring::Event::TopologyChanged.new(opening_topology, @topology)
-      ) if seeds.size >= 1
+      if seeds.size >= 1
```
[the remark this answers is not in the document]
saghm: Understood

lib/mongo/cluster/topology/server_description_list.rb
saghm: Python, for one. See [here](https://github.com/mongodb/mongo-python-driver/blob/341b2b5964b246fe7419a812e31b84e023c7c030/pymongo/topology_description.py#L44-L53) where they initialize the list of server descriptions in the topology description to a dictionary and [here](https://github.com/mongodb/mongo-python-driver/blob/341b2b5964b246fe7419a812e31b84e023c7c030/pymongo/topology.py#L110) where they initialize the list of servers in the topology to a dictionary.

More generally, it just seems like a more intuitive way to store things to me. Given that we need to be able to look up servers by the address and never care about the order of them in the container, a hash seems like the most logical choice. 

The spec does use an array [here](https://github.com/mongodb/specifications/blob/97fd1f26c3fa7506353f7e310063822b958a53b0/source/server-discovery-and-monitoring/tests/rs/new_primary.yml#L8) for the isMaster responses, but that's specifically because they need to be applied to the cluster in a certain order. (I'm honestly not sure why they make each element of that array an array as well; I think an object with the address as a key would make more sense). However, when describing the expected state of the topology after applying the isMaster responses, the spec does use an object, not an array, as shown [here](https://github.com/mongodb/specifications/blob/97fd1f26c3fa7506353f7e310063822b958a53b0/source/server-discovery-and-monitoring/tests/rs/new_primary.yml#L21-L36).

p-mongo: Okay I will change it to be a hash.

p-mongo:
Made the requested changes.

PR #1142, opened 2018-11-07
1547

p-mongo:


spec/mongo/operation/update/command_spec.rb
```diff
@@ -105,7 +106,7 @@
         }
       end
 
-      it 'creates the correct Command message', unless: op_msg_enabled? do
+      it 'creates the correct Command message' do
```
saghm: Should this have a `max_server_version` requirement, or is the test not actually dependent on being below 3.6?

p-mongo: There is a max server version constraint in an outer context higher up in the diff.

saghm: Ah, I missed that. Thanks!

PR #1144, opened 2018-11-07
Fix CI on ruby 1.9

p-mongo:


p-mongo:
The move is for readability only.

PR #1150, opened 2018-11-09
Fix sdam spec runner to require that all actual servers are accounted for in the spec

p-mongo:


p-mongo:
Following comments in https://github.com/mongodb/specifications/pull/410.

PR #1151, opened 2018-11-09
RUBY-1581 Notify waiting server selector when a server is available

p-mongo:


lib/mongo/client.rb
```diff
@@ -325,16 +325,25 @@ def initialize(addresses_or_uri, options = nil)
...
+      yield(self) if block_given?
+    end
+
+    # @api private
+    def cluster_options
       # We share clusters when a new client with different CRUD_OPTIONS
       # is requested; therefore, cluster should not be getting any of these
       # options upon instantiation
-      cluster_options = @options.reject do |key, value|
+      options.reject do |key, value|
```
saghm: Would caching this value in an instance variable cause incorrect behavior? I don't think the options should change

saghm: We do it all over the place in the driver already. What's different about here?

p-mongo: It may be a premature optimization elsewhere as well.

saghm: I don't feel strongly enough about this to block merging, so it's fine to leave it as-is

PR #1155, opened 2018-11-11
RUBY-1580 Deduplicate code across OP_MSG operation classes

p-mongo:


lib/mongo/operation/shared/polymorphic_result.rb
```diff
@@ -0,0 +1,35 @@
...
+    # Shared behavior of instantiating a result class matching the
+    # operation class.
+    #
+    # @api private
+    module PolymorphicResult
+      private
+
+      def result_class
+        bits = self.class.name.sub(/::[^:]*$/, "::Result").split('::')
+        cls = Object
```
saghm: Not necessary to change if you prefer it this way, but this looks like it's equivalent to `bits.reduce(Object) { |cls, name| cls.const_get(name) }`

p-mongo: Thanks, I've been looking for a shorter way to express that.

PR #1163, opened 2018-11-16
RUBY-1587 Properly handle error labels sent by server

saghm:


lib/mongo/error/operation_failure.rb
```diff
@@ -182,6 +182,16 @@ def initialize(message = nil, result = nil, options = {})
...
         @code_name = options[:code_name]
         super(message)
+
+        if result
+          err_doc = result.send(:first_document)
+
+          if err_doc && err_doc['errorLabels']
+            err_doc['errorLabels'].each do |label|
+              add_label(label)
+            end
```
p-mongo: There is a Mongo::Error::Parser which parses all the other fields out of the result, how come this code is added to the exception class and not to the parser?

spec/support/transactions/operation.rb
```diff
@@ -114,12 +114,12 @@ def execute(collection, session0, session1)
...
           'errorCodeName' => err_doc['codeName'] || err_doc['writeConcernError']['codeName'],
           'errorContains' => e.message,
-          'errorLabels' => (e.instance_variable_get(:@labels) || []) + (err_doc['errorLabels'] || [])
+          'errorLabels' => e.instance_variable_get(:@labels)
         }
       rescue Mongo::Error => e
         {
           'errorContains' => e.message,
-          'errorLabels' => e.instance_variable_get(:@labels) || []
+          'errorLabels' => e.instance_variable_get(:@labels)
```
p-mongo: I think Mongo::Error should have a `labels` method added following this part of https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#error-reporting-changes:

> Drivers MAY expose the list of all error labels for an exception object.

lib/mongo/error/operation_failure.rb
saghm: Good idea! I'll change it to do that

spec/support/transactions/operation.rb
saghm: Will do

saghm:
All suggested changes have been made

saghm:
Updated again per offline discussion

lib/mongo/operation/get_more/result.rb
```diff
@@ -52,7 +52,7 @@ def cursor_document
         end
 
         def first_document
-          @first_document ||= reply.documents[0]
+          @first_document ||= reply && reply.documents[0]
```
p-mongo: Can this be now returned to unconditionally reading `reply`?

saghm: Done

saghm:
Sorry, I must have misunderstood how you wanted the tests changed from our offline discussion. I've added the test back and changed it to use `Result#raise_operation_failure` to ensure the correct code path is covered

PR #1170, opened 2018-11-27
Implement uri option spec tests

saghm:
Spec test pull request: https://github.com/mongodb/specifications/pull/414


lib/mongo/uri.rb
```diff
@@ -615,6 +626,94 @@ def auth_mech_props(value)
...
     end
 
+    # Parses the zlib compression level.
+    #
+    # @param value [ String ] The zlib compression level string.
+    #
+    # @return [ Integer | nil ] The compression level value if it is between -1 and 9 (inclusive),
+    #   otherwise nil (and a warning will be raised).
+    def zlib_compression_level(value)
+      if /^-?\d+$/ =~ value
```
p-mongo: Ruby uses \A and \z for beginning and end of input. ^ and $ mean beginning and end of line.

lib/mongo/uri.rb
```diff
@@ -615,6 +626,94 @@ def auth_mech_props(value)
...
+      nil
+    end
+
+
+    # Parses a boolean value.
+    #
+    # @param value [ String ] The URI option value.
+    #
+    # @return [ true | false | nil ] The boolean value parsed out, otherwise nil (and a warning
+    #   will be raised).
```
p-mongo: warning logged

lib/mongo/uri.rb
```diff
@@ -615,6 +626,94 @@ def auth_mech_props(value)
...
+    def bool(value)
+      case value
+      when "true"
+        true
+      when "false"
+        false
+      else
+        log_warn("invalid boolean: #{value}")
```
p-mongo: This log message should also specify what the option was that the value goes with, to assist in troubleshooting.

lib/mongo/uri.rb
```diff
@@ -615,6 +626,94 @@ def auth_mech_props(value)
...
+
+    # Parses a boolean value and returns its inverse. This is used for options where the spec
+    # defines the boolean value semantics of enabling/disabling something in the reverse way that
+    # the driver does. For instance, the client option `ssl_verify` will disable certificate
+    # checking when the value is false, but the spec defines the option
+    # `tlsAllowInvalidCertificates`, which disables certificate checking when the value is true.
+    #
+    # @param value [ String ] The URI option.
+    #
+    # @return [ true | false | nil ] The inverse of boolean value parsed out, otherwise nil (and a
```
p-mongo: Logical not never returns nil. Only true and false are viable return values of this method as implemented.

p-mongo: Which means also options like tlsallowinvalidcertificates will always be parsed into true/false and not nil when not provided.

lib/mongo/uri.rb
```diff
@@ -615,6 +626,94 @@ def auth_mech_props(value)
...
+    end
+
+    # Parses the max staleness value, which must be either "0" or an integer greater or equal to 90.
+    #
+    # @param value [ String ] The max staleness string.
+    #
+    # @return [ Integer | nil ] The max staleness integer parsed out if it is valid, otherwise nil
+    #   (and a warning will be raised).
+    def max_staleness(value)
+      if /^\d+$/ =~ value
```
p-mongo: \A, \z

lib/mongo/uri.rb
```diff
@@ -615,6 +626,94 @@ def auth_mech_props(value)
...
+        int = value.to_i
+
+        if int >= 0 && int < 90
+          log_warn("max staleness must be either 0 or greater than 90: #{value}")
```
p-mongo: But the code then returns the `int` even though the value is unacceptable?

lib/mongo/uri.rb
```diff
@@ -615,6 +626,94 @@ def auth_mech_props(value)
...
+    end
+
+    # Parses an unsigned integer value.
+    #
+    # @param value [ String ] The URI option value.
+    #
+    # @return [ Integer | nil ] The integer parsed out, otherwise nil (and a warning will be
+    #   raised).
+    def unsigned_int(value)
+     unless /^?\d+$/ =~ value
```
p-mongo: This is not the correct regexp to use here. For example, this regexp matches  `abc90`.

lib/mongo/uri.rb
```diff
@@ -625,6 +724,11 @@ def auth_mech_props(value)
     #
     # @since 2.0.0
     def ms_convert(value)
+      unless /^-?\d+(\.\d+)?$/ =~ value
```
p-mongo: \A, \z

lib/mongo/uri.rb
```diff
@@ -625,6 +724,11 @@ def auth_mech_props(value)
...
+        log_warn("Invalid ms value: #{value}")
```
p-mongo: Specify the option that the value goes with.

lib/mongo/uri.rb
```diff
@@ -636,6 +740,11 @@ def ms_convert(value)
     def hash_extractor(value)
       value.split(',').reduce({}) do |set, tag|
         k, v = tag.split(':')
+        if v.nil?
+          log_warn("#{value} is not a valid URI hash")
```
p-mongo: Specify the option that the value goes with.

spec/spec_tests/uri_options_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+              info = ::Socket.getaddrinfo(host, nil, family, ::Socket::SOCK_STREAM)
+              FAMILY_MAP[info.first[4]].new(info[3], port, host)
+            end
+          end
+
+          class Server
+
+            # The constructor keeps the same API, but does not instantiate a
+            # monitor and run it.
+            alias :original_initialize :initialize
```
p-mongo: This code was removed from master, please update this spec to match what is currently in the connection string spec.

spec/spec_tests/uri_options_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+            alias :disconnect! :original_disconnect!
+            remove_method(:original_disconnect!)
+          end
+        end
+      end
+
+      spec.tests.each do |test|
+
+        # Skip optional tests for options the driver doesn't support.
+        if test.description.start_with?('tlsAllowInvalidHostnames') || test.description.start_with?('tlsInsecure')
```
p-mongo: Why does Ruby not implement these options?

spec/spec_tests/uri_options_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+          end
+
+          context 'when the uri is valid', if: test.valid? && !test.warn? do
+
+            before do
+              expect(Mongo::Logger.logger).not_to receive(:warn)
+            end
+
+            it 'does not raise an exception or warning' do
+              expect(test.uri).to be_a(Mongo::URI)
```
p-mongo: Why is this test not performed when `test.valid? && test.warn?` ?

lib/mongo/uri.rb
saghm: There are existing tests in the driver that assert this behavior (e.g. [here](https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/mongo/uri_spec.rb#L634)). I'm fine with changing those tests and the behavior here if you think that's better.

spec/spec_tests/uri_options_spec.rb
saghm: According to the `Client` documentation, the `ssl_verify` option only toggles whether or not invalid certificates are accepted. Given that the driver currently does not implement the functionality to allow invalid hostnames, this option was left off (all of the other options added to this file already existed as `Client` options, just not in the URI).

The spec only requires that either `tlsInsecure` or one of `tlsAllowInvalidHostnames`/`tlsAllowInvalidCertificates` is implemented. Since we implement `tlsAllowInvalidCertificates`, we don't need to implement `tlsInsecure` (which would also currently be equivalent to `tlsAllowInvalidCertificates` since as mentioned above, the driver doesn't have the functionality to allow invalid hostnames.

lib/mongo/uri.rb
saghm: I'm not super worried about users embedding newlines into their URI options, but ok.

spec/spec_tests/uri_options_spec.rb
p-mongo: Ruby implemented hostname verification in https://github.com/ruby/openssl/pull/60. As far as I can tell this is included in 2.4.0. I cannot make out what the default is or when hostname verification gets turned on exactly, therefore, we should also have tests that the hostname verification is performed.

saghm: I'm a little confused about what you're asking here. Are you asking that I go and implement hostname verification (which the driver currently doesn't do at all) as part of this PR? That feels a bit out of scope, given that it's not necessary work for the spec

lib/mongo/uri.rb
```diff
@@ -615,6 +626,94 @@ def auth_mech_props(value)
       properties
     end
 
+    # Parses the zlib compression level.
+    #
+    # @param value [ String ] The zlib compression level string.
+    #
+    # @return [ Integer | nil ] The compression level value if it is between -1 and 9 (inclusive),
+    #   otherwise nil (and a warning will be raised).
```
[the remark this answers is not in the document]
saghm: Fixed

lib/mongo/uri.rb
```diff
@@ -571,7 +582,7 @@ def auth_source(value)
     #
     # @return [Symbol] The transformed authentication mechanism.
     def auth_mech(value)
-      AUTH_MECH_MAP[value.upcase]
+      AUTH_MECH_MAP[value.upcase] || log_warn("#{value} is not a valid auth mechanism")
```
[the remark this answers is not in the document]
saghm: Fixed

lib/mongo/uri.rb
saghm: Done

lib/mongo/uri.rb
saghm: Fixed

lib/mongo/uri.rb
saghm: Fixed

saghm: Fixed

lib/mongo/uri.rb
saghm: Done

lib/mongo/uri.rb
saghm: Fixed

lib/mongo/uri.rb
saghm: Done

lib/mongo/uri.rb
saghm: Fixed

spec/spec_tests/uri_options_spec.rb
saghm: Done

spec/spec_tests/uri_options_spec.rb
saghm: Fixed

p-mongo:
This PR should also have a ticket reference, API documentation for added options and tutorial documentation for added options.

saghm:
There is no ticket to reference because the drivers tickets won't be made until the spec is merged; this is a PoC, so it's supposed to land before then.

spec/spec_tests/uri_options_spec.rb
p-mongo: Since hostname verification is mentioned in the spec, that functionality should be implemented and documented in Ruby driver or there should be tickets created for this work.

spec/spec_tests/uri_options_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+            alias :create_resolver :original_create_resolver
+            remove_method(:original_create_resolver)
+          end
+        end
+      end
+
+      spec.tests.each do |test|
+
+        # Skip optional tests for options the driver doesn't support.
+        next if test.description.start_with?('tlsAllowInvalidHostnames', 'tlsInsecure')
```
p-mongo: tlsInsecure should be implemented by Ruby driver and when set, this should set tlsAllowInvalidHostnames and tlsAllowInvalidCertificates to true.

saghm:
We currently don't have any API documentation for URI options; we have it for client options, but no new client options have been added, only additional mappings to them from URI options. If you want me to, I can open a separate ticket for us to add API documentation for URI options.

spec/spec_tests/uri_options_spec.rb
saghm: https://jira.mongodb.org/browse/RUBY-1608

spec/spec_tests/uri_options_spec.rb
saghm: `tlsInsecure` has been added

docs/tutorials/ruby-driver-create-client.txt
```diff
@@ -97,8 +97,20 @@ URI Options Conversions
...
 
-   * - replicaSet=String
-     - ``:replica_set => String``
+   * - appName=String
+     - ``:app_name => String``
+
+   * - authMechanism=String
+     - ``:auth_mech => Symbol``
+
+   * - authMechanismProperties=Strings
```
p-mongo: How would a user determine what to put in `Strings` for this option?

saghm: This already existed in the old code; I just alphabetized the list for easier reading. If you just look at the diff for https://github.com/mongodb/mongo-ruby-driver/pull/1170/commits/e396935447b59a9cf05b2622cb579227623ebdcb, you'll see the options that are new.

spec/spec_tests/uri_options_spec.rb
p-mongo: If tlsInsecure is now supported by the driver, why is its test skipped?

lib/mongo/uri.rb
```diff
@@ -426,29 +426,38 @@ def self.uri_option(uri_key, name, extra = {})
...
 
     # Security Options
     uri_option 'ssl', :ssl
+    uri_option 'tls', :ssl
+    uri_option 'tlsallowinvalidcertificates', :ssl_verify, :type => :ssl_verify
+    uri_option 'tlscafile', :ssl_ca_cert
+    uri_option 'tlsclientcertfile', :ssl_cert
+    uri_option 'tlsclientkeyfile', :ssl_key
+    uri_option 'tlsclientkeypassword', :ssl_key_pass_phrase
+    uri_option 'tlsinsecure', :ssl_verify, :type => :ssl_verify
```
p-mongo: The documentation states that ssl_verify means:

> Whether to perform peer certificate validation.

Either the documentation needs to be updated to reflect the new behavior, which to me seems like the better option, or `tlsInsecure` cannot be an inverse of `ssl_verify`.

spec/spec_tests/uri_options_spec.rb
saghm: Fixed

lib/mongo/uri.rb
saghm: Fixed in documentation

docs/tutorials/ruby-driver-create-client.txt
saghm: I've added extra documentation to this option and the others below where the type being forwarded is different from the type specified in the URI. Let me know if you think any of the other options need extra documentation as well.

saghm:
I believe I've addressed all of the outstanding comments on this, so it's ready for review again

docs/tutorials/ruby-driver-create-client.txt
```diff
@@ -97,8 +97,23 @@ URI Options Conversions
...
+     - ``{ :auth_mech_properties => { :service_realm => String, :canonicalize_host_name => true|false, :service_name => String } }``
+
+       Specified as comma-separated key:value pairs, e.g. ``"SERVICE_REALM:foo,CANONICALIZE_HOSTNAME:TRUE"``
+
+   * - authSource=String
+     - ``:auth_source => String``
+
+   * - compressors=Strings
+     - ``:compressors => Array<String>``
+       Specified as a comma-separated list, e.g. "zlib,snappy"
```
p-mongo: This is a weird default to specify in the Ruby driver because snappy isn't supported by Ruby. Will this warn?

Perhaps a better language is something along the lines of:

> Specified as a comma-separated list. Note that the Ruby driver only supports zlib compression, however other drivers may support snappy. For maximum compatibility with drivers specify "zlib,snappy", if compatibility with other drivers is not a concern specify "zlib".

saghm: I didn't intend this to be interpreted as a default, but rather just to demonstrate how specifying more than one would work. I'm happy to make the suggested change, though.

p-mongo: Okay that makes sense, let's put this language in given user confusion over compression.

saghm: Just pushed

PR #1171, opened 2018-11-28
RUBY-1601 Change server description server type method implementations to match SDAM spec


p-mongo:


lib/mongo/cluster.rb
```diff
@@ -743,24 +744,32 @@ def add(host, add_options=nil)
...
       removed_servers = @servers.select { |s| s.address == address }
       @update_lock.synchronize { @servers = @servers - removed_servers }
-      removed_servers.each do |server|
-        if server.connected?
-          server.disconnect!
-          publish_sdam_event(
-            Monitoring::SERVER_CLOSED,
-            Monitoring::Event::ServerClosed.new(address, topology)
-          )
+      if disconnect != false
```
egiurleo: Is there a reason to write `if disconnect != false` rather than `if disconnect == true` or just `if disconnect` here?

lib/mongo/cluster/sdam_flow.rb
```diff
@@ -73,6 +74,40 @@ def update_server_descriptions
...
+        # we need to add and remove hosts in that description even if
+        # it also has a me mismatch. In the event of a me mismatch, the
+        # server for which we are processing the response will be removed
+        # from topology, which may cause the current thread to terminate
+        # prior to running the entire sdam flow.
+
+        servers = add_servers_from_desc(updated_desc)
+        # Spec tests require us to remove servers based on data in descrptions
+        # with me mismatches. The driver will be more resilient if it only
+        # removed servenrs from descriptions with matching mes.
```
egiurleo: Small typo in `servenrs`

lib/mongo/cluster.rb
p-mongo: This evaluates to `true` when disconnect is `nil`. `if disconnect == true` evaluates to false for truthy values of disconnect like `0`. What this statement intends to convey is the default is false, and anything other than false is true.

p-mongo: I changed the docstring to clarify this.

PR #1176, opened 2018-12-17
RUBY-1618 DRY operation execution

p-mongo:


lib/mongo/operation/indexes.rb
```diff
@@ -27,26 +27,7 @@ module Operation
...
-      def execute(server)
-        if server.features.op_msg_enabled?
-          OpMsg.new(spec).execute(server)
-        elsif server.features.list_indexes_enabled?
-          Command.new(spec).execute(server)
-        else
-          Legacy.new(spec).execute(server)
-        end
-      end
+      include OpMsgOrListIndexesCommand
```
saghm: This appears not to be used anywhere else; is the reason for it to just be consistent about not having any implementations of `#execute` in operation classes?

p-mongo: Yes, for consistency with the other ones.

saghm: Sounds good

PR #1178, opened 2018-12-20
Specify pool size in exception message when pool is at max connection…

p-mongo:
… count


saghm:
This is going to be superseded by the changes due to the CMAP spec that I'm working on; among other things, the CMAP spec specifies a specific error type which has to be thrown here. Is this something that will be helpful in the interim, or is this just intended for better user-facing messaging? If the latter, I can just make sure to include that info in the error messages defined in the CMAP spec.

p-mongo:
OK.

PR #1179, opened 2018-12-20
RUBY-1623 Create a write concern base class

p-mongo:


spec/mongo/bulk_write_spec.rb
```diff
@@ -1914,6 +1914,8 @@
 
             let(:subscriber) { EventSubscriber.new }
 
+            let(:subscriber) { EventSubscriber.new }
```
saghm: I assume this being duplicated is a mistake

p-mongo: Yes indeed.

PR #1182, opened 2018-12-26
RUBY-1450 remove test_sessions?

p-mongo:


spec/mongo/bulk_write_spec.rb
```diff
@@ -1908,7 +1908,9 @@
             it_behaves_like 'an operation using a session'
           end
 
-          context 'when retryable writes are supported', if: test_sessions? do
+          context 'when retryable writes are supported' do
+            min_server_version '3.6'
```
saghm: Given the number of places this is done, maybe we should make a `require_sessions` method and just call it in each of these places?

p-mongo: This is not such a clear cut decision because some of the session-related tests do also test pre-session servers (and do something else on those).

p-mongo: Or they test other topologies.

saghm: Makes sense; thanks for the clarification!

PR #1200, opened 2019-01-04
RUBY-1560 Implement CMAP spec

saghm:


spec/support/connection_monitoring_and_pooling.rb
```diff
@@ -0,0 +1,370 @@
+RSpec::Matchers.define :match_events do |actual|
```
p-mongo: These matchers do not point out what the difference between actual and expected results are, only that the two are different. I implemented these matchers as verifiers in https://github.com/mongodb/mongo-ruby-driver/pull/1203, you can implement a similar verifier for cmap or some other solution that highlights the differences but please no more matchers like this in the driver.

spec/lite_spec_helper.rb
```diff
@@ -67,6 +68,7 @@
 require 'support/sdam_monitoring'
 require 'support/crud'
 require 'support/command_monitoring'
+require 'support/connection_monitoring_and_pooling'
```
p-mongo: Please rename this file to `support/cmap` for consistency.

lib/mongo/server/connection_pool.rb
```diff
@@ -21,31 +21,45 @@ class Server
...
       #   seconds, for a free connection.
       #
       # @since 2.0.0
-      def initialize(options = {}, &block)
-        @options = options.freeze
-        @queue = Queue.new(options, &block)
+      # @api private
+      def initialize(server, &block)
+        @id = ConnectionPool.next_id
+        @address = server.address.to_s
```
p-mongo: Keep the address as a `Mongo::Address` object and expose it as such in all events, for consistency with the rest of the driver.

lib/mongo/server/connection_pool/queue.rb
```diff
@@ -149,10 +166,19 @@ def dequeue
...
         def disconnect!
-          check_count_invariants
+          check_count_invariants(false)
           mutex.synchronize do
-            queue.each{ |connection| connection.disconnect! }
-            @pool_size -= queue.length
+            @pool_size -= queue.size
+            queue.each do |connection|
+              connection.disconnect!
+              publish_cmap_event(
```
p-mongo: Event publishing invokes application code which may raise exceptions and thus abort the process. For each connection in the queue, that connection must be removed from the queue, pool size reduced by 1, then the event published. Please audit the rest of the diff as well considering effects of exceptions raised during event publishing.

lib/mongo/error/wait_queue_timeout.rb
```diff
@@ -0,0 +1,41 @@
...
+      # Instantiate the new exception.
+      #
+      # @example Instantiate the exception.
+      #   Mongo::Error::WaitQueueTimeout.new(address)
+      #
+      # @since 2.7.0
+      def initialize(address)
+        @address = address
+
+        super("timeout when attempting to check out a connection from pool with address #{address}")
```
p-mongo: Per https://github.com/mongodb/mongo-ruby-driver/pull/1178, this message should include also the size of the pool and the timeout that was exceeded, and should begin with a capital letter.

spec/support/connection_monitoring_and_pooling.rb
saghm: I'm not actually convinced that this is any better than what we already have; `verify_operation_result` just prints out the difference without context, e.g. `expected 2; got 3` when a single field is different in a list of documents from a cursor, and I find that harder to debug than what I implemented here. That said, you seem to feel much more strongly than I do on this particular issue, so I'll implement the changes as requested.

lib/mongo/server/connection_pool/queue.rb
```diff
@@ -79,6 +83,10 @@ def initialize(options = {}, &block)
           @queue = Array.new(min_size) { create_connection }
           @mutex = Mutex.new
           @resource = ConditionVariable.new
+
+          @wait_queue = []
+          @wait_queue_mutex = Mutex.new
```
p-mongo: Why are there now two parallel lock implementations?

lib/mongo/server/connection_pool/queue.rb
```diff
@@ -313,19 +329,86 @@ def close_stale_sockets!
...
+              return connection if connection
+
+              wait = deadline - Time.now
+              resource.wait(mutex, wait)
+              raise Error::WaitQueueTimeout.new(@address) if deadline <= Time.now
+            end
+          end
+        end
+
+        def check_wait_queue
```
p-mongo: What does this method do?

lib/mongo/server/connection_pool/queue.rb
```diff
@@ -190,31 +216,21 @@ def disconnect!
...
-                # This should never happen
-                log_warn("ConnectionPool::Queue: unexpected enqueue")
-                0
-              end
-
-              while @pool_size < min_size
-                @pool_size += 1
-                queue.unshift(@block.call(@generation))
+              @wait_queue_mutex.synchronize do
+                @wait_queue.first.broadcast unless @wait_queue.empty?
```
p-mongo: Why are there two broadcast calls in this method?

lib/mongo/server/connection_pool/queue.rb
p-mongo: Please fix any other occurrences of incorrect pool size management as well.

spec/lite_spec_helper.rb
p-mongo: `support/server_discovery_and_monitoring.rb` should be renamed to `support/sdam.rb` for the same reason.

lib/mongo/server/connection_pool/queue.rb
saghm: One is for properly handling access to the list of available connections (i.e. `@queue`) and one is for properly handling access to the list of waiting threads (i.e. `@wait_queue)

lib/mongo/server/connection_pool/queue.rb
saghm: It adds a thread to the wait queue, and, if necessary, returns a semaphore to signal the thread when to wake up and acquire a connection.

lib/mongo/server/connection_pool/queue.rb
saghm: Because there are two separate semaphores being signaled.

lib/mongo/server/connection_pool/queue.rb
```diff
@@ -313,39 +332,101 @@ def close_stale_sockets!
...
-            raise Timeout::Error.new("Timed out attempting to dequeue connection after #{wait_timeout} sec.")
-          end
-          resource.wait(mutex, wait)
-        end
-
-        def check_count_invariants
+        # We only create new connections when we're below the minPoolSize on creation, when we're
+        # disconnecting and starting a new generation, and when we're checking out connections (per
+        # the CMAP spec), so `check_min` should be false in all other cases.
+        def check_count_invariants(check_min = true)
```
p-mongo: Since connections are created on the foreground thread, in which circumstances is the min invariant violated by this PR as currently implemented?

p-mongo:
Queue has this comment:

<pre>
        # Only connections created by this queue should be enqueued
        # back into it, however the queue does not verify whether it
        # originally created the connection being enqueued.
</pre>

This should be changed to reflect current behavior which is to verify ownership.

lib/mongo/server/connection_pool/queue.rb
saghm: In all other situations than described by the comment to this method (i.e. wherever this method is called with `false` as the parameter). One example is when a pool with minPoolSize > 1 is cleared and then a connection is checked out; according to the spec, the connections beyond the first should be created in a non-blocking manner, so we can't create the rest of the connections needed to reach the min pool size until we implement the background thread, which you requested that we hold off for a separate review.

lib/mongo/server/connection_pool/queue.rb
p-mongo: Please add a docstring containing this information, as well as when the semaphore is returned and when it is not returned.

saghm:
This still reflects the current behavior of the queue. Checking the generation is not the same as checking if the connection was created by the same queue.

lib/mongo/server/connection_pool/queue.rb
saghm: We don't currently document any of existing private methods in the Queue; I don't see why this one is any different, especially given that it's much shorter, so it's already obvious what's being returned.

p-mongo:
Please add a prose description to the diff of how the wait queue works. The spec gives two examples of possible implementations and looking at the code it is not clear to me how the Ruby driver implements the wait queue, in part because the implementation is not encapsulated in its own class but spread across queue methods with the "queue" apparently performing the jobs of two queues.

saghm:
One of the examples given of a wait queue in the spec is a queue of asynchronous callbacks, which I think is pretty obviously not what is implemented here. 

The wait queue is a mutex and an array (separate from the mutex and array for the available connections). When a thread attempts to check out a connection, it creates a semaphore, acquires the lock for the wait queue, and adds the semaphore to the wait queue, then releases the wait queue lock. If it's not the first thread in the wait queue, then it sleeps until either a timeout or the semaphore it added to the wait queue is signaled (which happens whenever a connection is checked in). Then, it does the connection checkout logic and pops itself from the wait queue.

p-mongo:
What happens if the thread is not the first thread in the wait queue? How does the thread pop itself off the queue without locking the queue? If another thread (or multiple threads) in the meantime has locked the queue, how long would the first thread end up spending trying to pop itself off the queue after it has already exceeded the wait timeout specified by the user?

saghm:
> What happens if the thread is not the first thread in the wait queue? 

Then it isn't woken up until it's the first in the wait queue.

>How does the thread pop itself off the queue without locking the queue? 

I'm not sure I understand this question; assuming you mean "wait queue" here both times you use the word "queue", then it obviously doesn't pop itself without acquiring the mutex. It acquires the mutex pops itself, and then releases the mutex. I don't see why this would be an issue, since popping itself off the array is a very fast operation, and then it releases the lock immediately after.

>If another thread (or multiple threads) in the meantime has locked the queue, how long would the first thread end up spending trying to pop itself off the queue after it has already exceeded the wait timeout specified by the user?

Assuming you mean "wait queue" here, it would spend as much time as it takes to run `Array#shift` popping itself off, and then the next thread would have to wait that long. I'm not sure what you're suggesting would be an alternative implementation; we obviously need a mutex to be able to use the wait queue in multiple threads, and we can't mutate the array outside the mutex. I don't think that users having the wait the extra amount of time of another thread removing itself from the array is going to actually cause issues for anyone.

lib/mongo/error/wait_queue_timeout.rb
saghm: Done

lib/mongo/server/connection_pool/queue.rb
saghm: I believe this has been resolved by our offline discussions and the recent updates

lib/mongo/server/connection_pool/queue.rb
saghm: I believe this has been resolved by our offline discussions and the recent updates

lib/mongo/server/connection_pool/queue.rb
p-mongo: The existing implementation (https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/server/connection_pool/queue.rb#L151) created replacement connection objects upon disconnecting existing connections, thus maintaining the invariants. What is preventing the new implementation from maintaining this behavior?

lib/mongo/server/connection_pool/wait_queue.rb
```diff
@@ -0,0 +1,58 @@
...
+        extend Forwardable
+
+        def initialize(address)
+          @address = address
+          @wait_queue = []
+          @mutex = Mutex.new
+        end
+
+        def clear!
+          until @wait_queue.empty?
```
p-mongo: All accesses of synchronized data structures including emptiness checks like this must be done under the lock.

lib/mongo/server/connection_pool/wait_queue.rb
```diff
@@ -0,0 +1,58 @@
...
+      class WaitQueue
+        extend Forwardable
+
+        def initialize(address)
+          @address = address
+          @wait_queue = []
+          @mutex = Mutex.new
+        end
+
+        def clear!
```
p-mongo: Why is there a bang on this method?

lib/mongo/server/connection_pool/available_stack.rb
```diff
@@ -119,121 +120,120 @@ def initialize(options = {}, &block)
...
-          check_count_invariants
+          check_count_invariants(false)
         end
 
-        # Disconnect all connections in the queue.
+        # Updates the generation number. The connections will be disconnected and removed lazily
+        # when the stack attempts to pop them.
+        #
+        # @since 2.7.0
+        def clear!
```
p-mongo: Why is there a bang on this method?

lib/mongo/server/connection_pool/available_stack.rb
```diff
@@ -300,31 +288,83 @@ def close_stale_sockets!
...
         ensure
-          check_count_invariants
+          check_count_invariants(false)
+        end
+
+        def connection_removed
+          @pool_size -= 1
+          if @pool_size < 0
+            # This should never happen
+            log_warn("ConnectionPool::AvailableStack: unexpected push")
```
p-mongo: This message needs to be updated because this method can be called as part of a sequence of operations not involving a push.

lib/mongo/server/connection_pool/available_stack.rb
```diff
@@ -119,121 +120,120 @@ def initialize(options = {}, &block)
...
+                  Monitoring::Event::ConnectionClosed.new(
+                      Monitoring::Event::ConnectionClosed::POOL_CLOSED,
+                      @address,
+                      connection.id,
+                  ),
+              )
             end
-            queue.clear
+
+            connections.clear
```
p-mongo: Connections must be removed from the list one at a time as they are being closed.

lib/mongo/server/connection_pool/available_stack.rb
```diff
@@ -119,121 +120,120 @@ def initialize(options = {}, &block)
...
             @generation += 1
```
p-mongo: This generation bump should probably also happen prior to closing any of the connections.

lib/mongo/server/connection_pool/available_stack.rb
```diff
@@ -300,31 +288,83 @@ def close_stale_sockets!
...
+            get_connection(deadline)
+          end
+        end
+
+        def get_connection(deadline)
           loop do
-            return queue.shift unless queue.empty?
+            until connections.empty?
+              connection = connections.shift
+              return connection unless is_stale?(connection) || is_idle?(connection)
```
p-mongo: If the connection is stale or idle and is not returned here, shouldn't it be closed?

lib/mongo/server/connection_pool/available_stack.rb
```diff
@@ -119,121 +120,120 @@ def initialize(options = {}, &block)
...
         # @since 2.0.0
-        def enqueue(connection)
-          check_count_invariants
+        def push(connection)
+          check_count_invariants(false)
           mutex.synchronize do
             if connection.generation == @generation
-              queue.unshift(connection.record_checkin!)
+              connections.unshift(connection.record_checkin!)
               resource.broadcast
```
p-mongo: This broadcast only happens if the connection is of the same generation. If the connection is stale, there is no broadcast (for the thread waiting in get_connection) but the stack size may have reduced below the max # of allowed connections, therefore the waiting thread should check that and potentially create a new connection immediately.

lib/mongo/server/connection_pool/available_stack.rb
```diff
@@ -300,31 +288,83 @@ def close_stale_sockets!
...
+          publish_cmap_event(
+            Monitoring::Event::ConnectionClosed.new(
+              reason,
+              @address,
+              connection.id,
+            )
+          )
+        end
+
+        def is_stale?(connection)
```
p-mongo: `is_stale?` is a query method name and does not match the behavior of this method. Something like `close_if_stale` is more appropriate.

lib/mongo/server/connection_pool/available_stack.rb
```diff
@@ -300,31 +288,83 @@ def close_stale_sockets!
...
+          if connection.generation != @generation
+            close_connection!(connection, Monitoring::Event::ConnectionClosed::STALE)
+            true
+          end
+        end
+
+        def is_idle?(connection)
```
p-mongo: Also here.

lib/mongo/server/connection_pool/available_stack.rb
```diff
@@ -300,31 +288,83 @@ def close_stale_sockets!
...
+            log_warn("ConnectionPool::AvailableStack: unexpected push")
+            @pool_size = 0
+          end
         end
 
         private
 
-        def dequeue_connection
-          deadline = Time.now + wait_timeout
+        def close_connection!(connection, reason)
```
p-mongo: A banged method name signifies a behavior difference from the non-banged method, or that something unusual is going on in the banged method. Simply closing objects is an insufficient reason to add bangs to method names.

lib/mongo/server/connection_pool/available_stack.rb
```diff
@@ -283,15 +271,15 @@ def max_idle_time
         #   if the option is set.
         #
         # @example Close the stale sockets
-        #   queue.close_stale_sockets!
+        #   stack.close_stale_sockets!
         #
         # @since 2.5.0
         def close_stale_sockets!
```
p-mongo: Was this method audited for correct operation given the other changes in the diff? I see that it calls `connection.disconnect!` instead of `close_connection` for example.

lib/mongo/server/connection_pool.rb
```diff
@@ -73,19 +119,79 @@ def checkin(connection)
...
+        raise_if_closed!
+
+        publish_cmap_event(
+          Monitoring::Event::ConnectionCheckoutStarted.new(address)
+        )
+
+        deadline = Time.now + wait_timeout
+        @wait_queue.enter_wait_queue(wait_timeout, deadline) do
+          connections.pop(deadline).tap do |c|
+            publish_cmap_event(
```
p-mongo: It seems to me that this event publication should not be done while the thread is considered to be in the wait queue.

lib/mongo/server/connection_pool.rb
```diff
@@ -73,19 +119,79 @@ def checkin(connection)
...
+      end
+
+      # Updates the generation number. The connections will be disconnected and removed lazily
+      # when the queue attempts to dequeue them.
+      #
+      # @return [ true ] true.
+      #
+      # @since 2.7.0
+      def clear!
+        unless closed?
```
p-mongo: I think `clear` should raise if the pool is closed.

lib/mongo/server/connection_pool.rb
```diff
@@ -73,19 +119,79 @@ def checkin(connection)
...
+        true
+      end
+
+      # Disconnects the pool and prevents any more connections from being checked out afterwards.
+      # If #checkout is called after #close!, a Mongo::Error::PoolClosed error will be raised.
+      #
+      # @return [ true ] true.
+      #
+      # @since 2.7.0
+      def close!
```
p-mongo: I believe the first thing this method needs to do is to set the close flag, otherwise concurrent operations in other threads may put something in the available stack for example while the pool is being closed.

lib/mongo/server/connection_pool.rb
saghm: Fair enough; I'll move that to outside the block

lib/mongo/server/connection_pool.rb
saghm: Okay, I'll change it

lib/mongo/server/connection_pool.rb
saghm: Sounds good; I'll update it. The reason I had done it this way was that I was using `@connections == nil` as the way to check if the pool was closed, and `@connections` is needed for the call to `disconnect!`.

lib/mongo/server/connection_pool/wait_queue.rb
saghm: I thought it was similar in nature to the `disconnect!` method on `ConnectionPool` and `AvailableStack`, so I put it there. I can take it off if you prefer it without.

lib/mongo/server/connection_pool/wait_queue.rb
saghm: Good catch! I'll fix that.

lib/mongo/server/connection_pool/wait_queue.rb
```diff
@@ -0,0 +1,58 @@
...
+          until @mutex.synchronize { @wait_queue.empty? }
+            semaphore = @mutex.synchronize { @wait_queue.shift }
```
p-mongo: This needs to take the lock and under the lock check emptiness and shift. The result will be spelled in a slightly awkward way given there's also a loop involved but that is ok.

saghm: Will do

spec/lite_spec_helper.rb
saghm: Done

p-mongo:
Please rename the verifier's directory to cmap as well.

spec/support/cmap.rb
```diff
@@ -0,0 +1,346 @@
...
+            Mongo::Event::Listeners.new,
+            pool_options)
+
+        state = {
+          'pool' => Mongo::Server::ConnectionPool.get(server) do
+            Mongo::Server::Connection.new(server)
+          end
+        }
+
+        preprocess
```
p-mongo: Running a test should not mutate its expectations. If any expectations need to be altered this should happen during construction.

spec/support/cmap.rb
```diff
@@ -0,0 +1,346 @@
...
+
+          opts
+        end
+      end
+
+      # Places operations run by the non-main thread in the `thread_ops` field of the corresponding
+      # `start` operation.
+      def preprocess
+        until spec_ops.empty?
+          processed_ops << spec_ops.shift
```
p-mongo: I expect `processed_ops` to contain the result of execution, not be a copy of `spec_ops`. What does this line do?

spec/support/cmap.rb
```diff
@@ -0,0 +1,346 @@
...
+
+        state = {
+          'pool' => Mongo::Server::ConnectionPool.get(server) do
+            Mongo::Server::Connection.new(server)
+          end
+        }
+
+        preprocess
+
+        { 'error' => nil }.tap do |result|
```
p-mongo: `'error' => nil` should be assigned once the test runs and the result is known.

lib/mongo/monitoring/event/connection_checked_in.rb
```diff
@@ -0,0 +1,60 @@
...
+        #
+        # @return [ String ] String summary of the event.
+        #
+        # @note This method is experimental and subject to change.
+        #
+        # @since 2.7.0
+        # @api experimental
+        def summary
+          "#<#{self.class.name.sub(/^Mongo::Monitoring::Event::/, '')} " +
+              "connection_id=#{connection_id} address=#{address}>"
```
p-mongo: I think the address should be before connection id in this string.

spec/support/cmap.rb
```diff
@@ -0,0 +1,346 @@
...
+                        'type' => 'ConnectionCheckedIn',
+                        'connectionId' => event.connection_id,
+                      }
+                    when Mongo::Monitoring::Event::PoolCleared
+                      {
+                        'type' => 'ConnectionPoolCleared',
+                        'address' => event.address,
+                      }
+                    else
+                      next
```
p-mongo: I think it would be helpful to have a base class for all connection pool-related events, so that the user code can also verify that it handled all pool events. Then this case would raise on an event which is (a subclass of) this base class and can ignore non-pool events.

spec/support/cmap.rb
```diff
@@ -0,0 +1,346 @@
...
+          opts
+        end
+      end
+
+      # Converts the options given by the spec to the Ruby driver format.
+      def process_options(options)
+        (options || {}).reduce({}) do |opts, kv|
+          case kv.first
+          when 'maxIdleTimeMS'
+            opts[:max_idle_time] = kv.last / 1000.0
```
p-mongo: Are these URI options, and if so, isn't there any code in the driver which would convert the URI options to driver options that can be reused and potentially be made available to users as well?

spec/support/cmap.rb
```diff
@@ -0,0 +1,346 @@
...
+        state[target].join
+      end
+
+      def run_checkout_op(state)
+        conn = state['pool'].checkout
+        state[label] = conn if label
+      end
+
+      def run_checkin_op(state)
+        until state[connection]; end
```
p-mongo: This loop should have a sleep in it, otherwise it will starve the other threads which this thread is in fact waiting for.

spec/support/connection_monitoring_and_pooling/verifier.rb
```diff
@@ -0,0 +1,72 @@
...
+            verify_event(event, expected_events)
+          end
+        end
+      end
+
+      private
+
+      def verify_event(event, events)
+        expect(
+          events.any? { |other| verify_hashes(event, other) }
```
p-mongo: Was this copied from another verifier? This should be a simple loop and verify_hashes should use assertions like verify_events does.

lib/mongo/server/connection_pool/queue.rb
saghm: I believe this is all set now

lib/mongo/server/connection_pool/queue.rb
saghm: This method is gone now, so I'm going to resolve this

lib/mongo/server/connection_pool/queue.rb
saghm: Based on offline discussions, I think we've determined that the new connections should be created by the background thread. Is this good to resolve?

lib/mongo/server/connection_pool/available_stack.rb
saghm: Removed

lib/mongo/server/connection_pool/available_stack.rb
saghm: Done.

lib/mongo/server/connection_pool/available_stack.rb
saghm: Fixed

lib/mongo/server/connection_pool/available_stack.rb
saghm: Done

lib/mongo/server/connection_pool/available_stack.rb
saghm: Done

lib/mongo/server/connection_pool/available_stack.rb
saghm: Changed

lib/mongo/server/connection_pool.rb
saghm: Done.

lib/mongo/server/connection_pool/wait_queue.rb
saghm: Done

spec/support/cmap.rb
```diff
@@ -0,0 +1,346 @@
...
+      #
+      # @since 2.7.0
+      def initialize(file)
+        @test = YAML.load(ERB.new(File::read(file)).result)
+
+        @description = @test['description']
+        @pool_options = snakeize_hash(process_options(@test['poolOptions'])).tap do |opts|
+          # The CMAP spec defines the default minPoolSize as 0, but the existing driver default is
+          # 1. Until we can change it, we need to manually specify their default when the YAML
+          # doesn't specify any.
```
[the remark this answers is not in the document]
saghm: Done

spec/support/cmap.rb
```diff
@@ -0,0 +1,346 @@
...
+
+        monitoring = Mongo::Monitoring.new(monitoring: false)
+        monitoring.subscribe(Mongo::Monitoring::CONNECTION_POOL, subscriber)
+
+        server = Mongo::Server.new(
+            Address.new(SpecConfig.instance.addresses.first),
+            cluster,
+            monitoring,
+            Mongo::Event::Listeners.new,
+            pool_options)
```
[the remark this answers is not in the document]
saghm: Done.

spec/support/connection_monitoring_and_pooling/verifier.rb
saghm: I've revised the implementation of this part of the verifier; let me know if you want any additional changes

lib/mongo/server/connection_pool/wait_queue.rb
saghm: Changed

lib/mongo/server/connection_pool/available_stack.rb
saghm: I've updated it; let me know if you prefer a different error message than the one I picked

lib/mongo/server/connection_pool/available_stack.rb
saghm: Done

spec/support/cmap.rb
```diff
@@ -0,0 +1,346 @@
...
+        @test = YAML.load(ERB.new(File::read(file)).result)
+
+        @description = @test['description']
+        @pool_options = snakeize_hash(process_options(@test['poolOptions'])).tap do |opts|
+          # The CMAP spec defines the default minPoolSize as 0, but the existing driver default is
+          # 1. Until we can change it, we need to manually specify their default when the YAML
+          # doesn't specify any.
+          opts[:min_pool_size] ||= 0
+        end
+        @num_pools = @test['numberOfPools'] || 1
```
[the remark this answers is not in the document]
saghm: It was leftover from a previous version of the spec tests; I've removed it

spec/lite_spec_helper.rb
```diff
@@ -106,6 +108,12 @@
   end
 
   if SpecConfig.instance.ci?
+    # To avoid a buildup of connections, we periodically close and reconnect each client if the
```
saghm: Previously, this was defined in a separate `around` hook in `spec_helper.rb` and without reconnecting the client. This worked because `close_stale_sockets!` previously eagerly reconnected sockets (as discussed offline). When `close_stale_sockets!` was updated to stop doing this, any time `close_local_clients` was called in the hook, tests would fail due to the pool being closed. I've changed `close_local_clients` to take an optional argument to reconnect the clients rather than just closing them. Unfortunately, this can take a while in some cases, and it sometimes pushes tests over the timeout boundary defined here. In order not to count the time it takes to reconnect the clients towards the timeout, I needed to move the implementation into the same `around` hook. To avoid reconnecting clients when only `lite_spec_helper` is included, I guarded it with the definition of a boolean constant in `spec_helper` defined right above the inclusion of `lite_spec_helper`. The client reconnections are done before the running of the test instead of after because [this assertion](https://github.com/mongodb/mongo-ruby-driver/blob/ad144e863879355372b324c3d48bf756372e5cd8/spec/mongo/client_construction_spec.rb#L69) failed due to the monitors starting when clients were reconnected after it.

lib/mongo/server/connection_pool/available_stack.rb
```diff
@@ -283,48 +263,92 @@ def max_idle_time
...
+            end
+
             connection = create_connection
             return connection if connection
-            wait_for_next!(deadline)
+
+            wait = deadline - Time.now
+            connection_available_condvar.wait(mutex, wait)
+
+            # TODO: Raise different error
```
saghm: The consensus in the pull request for the spec change to add a separate error seemed to be against it. I think we can still fix this on our end though; if we make a subclass of `WaitQueueTimeout`, we can raise that here instead, which will let us put the info we want in the error message while still fulfilling the spec requirement that `WaitQueueTimeout` is raised, and users can still just catch `WaitQueueTimeout` if they don't care about which type is raised.

spec/mongo/collection_spec.rb
```diff
@@ -3386,6 +3386,7 @@ def generate
       around(:each) do |spec|
         authorized_client[:validating,
                           :validator => { :a => { '$exists' => true } }].tap do |c|
+          c.drop
```
saghm: After the changes made to reconnect the clients, this below line was raising an error due to the collection already being created

spec/mongo/cursor_spec.rb
```diff
@@ -33,7 +33,7 @@
       end
 
       it 'does not schedule the finalizer' do
-        expect(ObjectSpace).not_to receive(:define_finalizer)
+        expect(ObjectSpace).not_to receive(:define_finalizer).with(Mongo::Cursor, anything)
```
saghm: After making the change to have the clients reconnect instead of just being closed after random tests, this was failing occasionally due to `define_finalizer` being called with `Monitor`, so I changed this to only assert that it wasn't called with `Cursor`

spec/spec_helper.rb
```diff
@@ -1,3 +1,9 @@
+# We use an RSpec `around` hook to close and reconnect the clients around every 100 tests. However,
```
saghm: See the comment in `lite_spec_helper` for why this is needed

spec/support/change_streams.rb
```diff
@@ -156,9 +156,9 @@ def run
         end
 
         def teardown_test
-          @global_client && @global_client.close
           @db1 && @db1.drop
           @db2 && @db2.drop
+          @global_client && @global_client.close
```
saghm: Dropping the databases raised an error after `close_stale_sockets!` was changed not to reconnect the sockets, so this needed to be updated. Given that `@db1` and `@db2` were created from `@global_client`, I think this is more correct anyways.

spec/support/cmap.rb
saghm: I think we should file a ticket for this rather than doing it here; this spec implementation has already grown a lot in scope, and we need to start winding down on it rather than adding more to it

lib/mongo/server/connection_pool/available_stack.rb
p-mongo: The base exception class should be called something like `ConnectionCheckoutError`, with the `WaitQueueTimeoutError` being its subclass.

p-mongo: And let's drop "Error" from `WaitQueueTimeoutError`, the only two classes with Error at the end is SocketError and SocketTimeoutError and the latter of them can also be spelled without the Error.

lib/mongo/monitoring.rb
```diff
@@ -25,6 +25,11 @@ class Monitoring
     # @since 2.1.0
     COMMAND = 'Command'.freeze
 
+    # The connection pool topic.
+    #
+    # @since 2.7.0
+    CONNECTION_POOL = 'ConnectionPool'.freeze
```
p-mongo: This needs to be added to the tutorial under monitoring section with examples for how to subscribe to cmap events.

lib/mongo/monitoring/event/cmap.rb
```diff
@@ -0,0 +1,38 @@
...
+# limitations under the License.
+
+module Mongo
+  class Monitoring
+    module Event
+
+      # Base class for CMAP events.
+      #
+      # @since 2.8.0
+      class Cmap < Mongo::Event::Base
```
p-mongo: Please make Cmap a module and Cmap::Base the base class.

lib/mongo/monitoring/event/cmap/connection_checked_in.rb
```diff
@@ -0,0 +1,62 @@
...
+
+        # Event fired when a connection is returned to a connection pool.
+        #
+        # @since 2.7.0
+        class ConnectionCheckedIn < Mongo::Monitoring::Event::Cmap
+
+          # @return [ String ] address The address of the server the connection was connected to.
+          #
+          # @since 2.7.0
+          attr_reader :address
```
p-mongo: These addresses should also be `Address` objects.

lib/mongo/monitoring/event/cmap/connection_checked_in.rb
```diff
@@ -0,0 +1,62 @@
...
+
+          # @return [ Integer ] connection_id The ID of the connection.
+          #
+          # @since 2.7.0
```
p-mongo: All of the sinces need to be updated to say 2.8.0.

lib/mongo/monitoring/event/cmap/connection_checked_in.rb
```diff
@@ -0,0 +1,62 @@
...
+
+          # Returns a concise yet useful summary of the event.
+          #
+          # @return [ String ] String summary of the event.
+          #
+          # @note This method is experimental and subject to change.
+          #
+          # @since 2.7.0
+          # @api experimental
+          def summary
```
p-mongo: The summary methods of all added events need to be unit tested as the other events are, unless they are tested in another way.

lib/mongo/monitoring/event/cmap/connection_checked_out.rb
```diff
@@ -0,0 +1,63 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module Mongo
+  class Monitoring
+    module Event
+      class Cmap
+
+        # Event fired when a connection is successfully checked out out of a pool.
```
p-mongo: Can we standardize on "event published" since that is what the methods are called?

lib/mongo/monitoring/event/cmap/connection_closed.rb
```diff
@@ -0,0 +1,102 @@
...
+
+          # @return [ Integer ] connection_id The ID of the connection.
+          #
+          # @since 2.7.0
+          attr_reader :connection_id
+
+          # @return [ String ] reason The reason why the connection was closed.
+          #
+          # @since 2.7.0
+          attr_reader :reason
```
p-mongo: It will be more idiomatic Ruby for reason to be a symbol, and `'poolClosed'` to be returned as `:pool_closed`.

lib/mongo/monitoring/event/cmap/connection_checkout_failed.rb
```diff
@@ -0,0 +1,75 @@
...
+          # @return [ Mongo::Address ] address The address of the server the connection would have
+          #   connected to.
+          #
+          # @since 2.7.0
+          attr_reader :address
+
+          # @return [ String ] reason The reason a connection was unable to be acquired.
+          #
+          # @since 2.7.0
+          attr_reader :reason
```
p-mongo: Also symbol reason here.

lib/mongo/protocol/message.rb
```diff
@@ -42,6 +42,7 @@ module Protocol
     class Message
       include Id
       include Serializers
+      include Id
```
p-mongo: This is including Id the second time.

lib/mongo/server.rb
```diff
@@ -61,6 +61,9 @@ def initialize(address, cluster, monitoring, event_listeners, options = {})
       monitor = options.delete(:monitor)
       @options = options.freeze
       @event_listeners = event_listeners
+      @connection_id_gen = Class.new do
+        include Id
+      end
```
p-mongo: Let's not create class definitions when unnecessary and instead define a class in the conventional way, and instantiate it here.

lib/mongo/server/connection.rb
```diff
@@ -100,6 +101,16 @@ def initialize(server, options = {})
       # @since 2.5.0
       attr_reader :last_checkin
 
+      # The ID for the connection. This will be unique across all connections for the same server.
+      #
+      # @since 2.7.0
+      attr_reader :id
```
p-mongo: This attribute and the address below are missing their types.

lib/mongo/monitoring/event/cmap.rb
saghm: Okay

lib/mongo/monitoring.rb
saghm: Done

lib/mongo/monitoring/event/cmap.rb
saghm: Done

lib/mongo/monitoring/event/cmap/connection_checked_in.rb
saghm: Done

lib/mongo/monitoring/event/cmap/connection_closed.rb
```diff
@@ -0,0 +1,102 @@
...
+          #   an error.
+          #
+          # @since 2.8.0
+          ERROR = :error
+
+          # @return [ Symbol ] POOL_CLOSED Indicates that the connection was closed due to the pool
+          #   already being closed.
+          #
+          # @since 2.8.0
+          POOL_CLOSED = :poolClosed
```
p-mongo: `:pool_closed` please

lib/mongo/monitoring/event/cmap/connection_closed.rb
```diff
@@ -0,0 +1,102 @@
...
+
+          # @return [ Symbol ] STALE Indicates that the connection was closed due to it being stale.
+          #
+          # @since 2.8.0
+          STALE = :stale
+
+          # @return [ Symbol ] IDLE Indicates that the connection was closed due to it being idle.
+          #
+          # @since 2.8.0
+          IDLE = :idle
```
p-mongo: Also, we do not need any of these constants, the symbols can be returned directly.

lib/mongo/server/connection.rb
p-mongo: OK, please add the types.

lib/mongo/monitoring/event/cmap/connection_closed.rb
p-mongo: This is not idiomatic Ruby. If you want to keep the constants, that is fine, but their values still need to be appropriate for someone to use directly.

docs/tutorials/ruby-driver-monitoring.txt
```diff
@@ -17,6 +17,7 @@ These events are organized into the following categories:
 - Topology lifecycle;
 - Server lifecycle;
 - Server heartbeats.
+- Connection pooling
```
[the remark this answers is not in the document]
p-mongo: Okay that is fine, as long as the result reads in a sensible manner.

lib/mongo/server.rb
p-mongo: OK.

lib/mongo/monitoring/event/cmap/connection_closed.rb
saghm: I'm not sure what you're suggesting I change here; I've updated the `POOL_CLOSED` constant to `:pool_closed`, and I don't see any casing issues with the others. Can you be more specific about what here needs to be updated?

saghm:
It appears that the issue with clients being sometimes closed after each test randomly was with the changes to `disconnect!` rather than to `close_stale_sockets!`. Since during our offline discussion you indicated didn't want the clients to be reconnected instead of closed, in each of the places where a client was being used after the close hook (namely in `collection_spec.rb`, I conditionally had the test reconnect the client, do the work it needed to do, and then close the client again.

lib/mongo/monitoring/event/cmap/connection_closed.rb
p-mongo: No action is needed, I am just pointing out the fact that most applications will not be using these constants.

p-mongo:
Superseded by other PRs listed in https://jira.mongodb.org/browse/RUBY-1560.

PR #1202, opened 2019-01-07
RUBY-1643 Abstract ID generation for classes

saghm:


lib/mongo/id.rb
```diff
@@ -0,0 +1,44 @@
...
+  # @example Define and use the Id module.
+  #   class Foo
+  #     include Id
+  #   end
+  #
+  #   f = Foo.new
+  #   foo.next_id # => 1
+  #   foo.next_id # => 2
+  #
+  # @since 2.7.0
```
p-mongo: This should also be marked `@api private` I think.

lib/mongo/id.rb
```diff
@@ -0,0 +1,44 @@
...
+  module Id
+    def self.included(klass)
+      klass.class_variable_set(:@@id, 1)
```
p-mongo: This assigns 1 to `id`, but then the next call to `next_id` also returns 1 which is confusing. Either `id` should be renamed to `@@next_id` to match the method, or the method should return the incremented number.

saghm: The idea behind this is that `@@id` should not be used at all outside of this file, and classes should only use `next_id` to obtain the next ID which has not been used yet. The fact that this returns `1` first is a feature rather than a bug. I can add documentation that explains this.

p-mongo: While the instance variables are indeed private, they should be used in a manner that is intuitive to understand. Prior to this PR ids started at 0 and the post-incremented value was returned. This behavior is one of two that I referred to in my previous comment.

saghm: This isn't an instance variable, it's a class variable. If you feel that strongly about what the value of the class variable is even though it shouldn't ever be directly accessed, I'll change it.

lib/mongo/protocol/message.rb
```diff
@@ -190,9 +191,7 @@ def hash
       #   server. The server will put this id in the response_to field of
       #   a reply.
       def set_request_id
-        @@id_lock.synchronize do
-          @request_id = @@request_id += 1
-        end
+        @request_id = Message.next_id
```
p-mongo: `Message.next_id` should be spelled `self.class.next_id` to use the correct class.

lib/mongo/id.rb
```diff
@@ -0,0 +1,63 @@
...
+  # should call `#next_id` in `#initialize` and save the result in the instance being created.
+  #
+  # @example Save the ID in the instance of the including class.
+  #   class Bar
+  #     include Id
+  #
+  #     attr_reader :id
+  #
+  #     def initialize
+  #       @id = next_id
```
p-mongo: Shouldn't `next_id` here also be `self.class.next_id` ?

lib/mongo/id.rb
```diff
@@ -0,0 +1,63 @@
...
+  #   f = Foo.new
+  #   foo.next_id # => 1
+  #   foo.next_id # => 2
+  #
+  # Classes which include Id should _not_ access `@@id` or `@@id_lock` directly; instead, they
+  # should call `#next_id` in `#initialize` and save the result in the instance being created.
+  #
+  # @example Save the ID in the instance of the including class.
+  #   class Bar
+  #     include Id
```
p-mongo: Since `Id` is scoped under `Mongo` and `Bar` is not, this include should probably read `include Mongo::Id`.

saghm: The module is marked `@api private`, so I think it's clear that this should only be used internally, but ok

p-mongo:
It is occurring to me that I do not recall seeing unit tests for the id generating code. If so, the `Id` module should have three tests added for it:

1. That the initial id value returned is 1.
2. That the id value is increasing, i.e. the second returned value is 2.
3. That if two classes include `Id`, their id sequences are independent.

PR #1206, opened 2019-01-11
RUBY-1608 Add tlsAllowInvalidHostnames URI option

saghm:
From looking into what the driver actually does, it turns out that it does actually verify hostnames (by calling `#verify_certificate_identity`), just after the connection was established rather than during the handshake. The functionality that was added in Ruby 2.4 was to verify the hostname during the actual handshake. In order to be consistent about the type of error thrown regardless of the version of Ruby installed, I did not change the code to use the newer functionality, instead just adding a `tlsAllowInvalidHostnames` knob; both `tlsAllowInvalidHostnames` and `tlsAllowInvalidCertificates` override `tlsInsecure`/`ssl_verify`, which otherwise turns off both types of verification when true.


lib/mongo/uri.rb
```diff
@@ -674,6 +690,26 @@ def ssl_verify(value)
...
+    def tls_allow_invalid_certificates(value)
+      bool('tls_allow_invalid_invalid_certificates', value)
+    end
+
+    # Parses the tlsAllowInvalidHostnames value.
+    #
+    # @param value [ String ] The tlsAllowInvalidHostnames value.
+    #
+    # @return [ true | false | nil ] The tlsAllowInvalidHostnames value parsed out, otherwise nil
+    #   (and a warning will be logged).
```
p-mongo: These should be marked `@api private` I think.

lib/mongo/uri.rb
```diff
@@ -658,14 +674,14 @@ def journal(value)
...
+    # tlsInsecure (if present).
     #
-    # @param value [ String ] The tlsAllowInvalidCertificates value.
+    # @param value [ String ] The tlsInsecure value.
     #
     # @return [ true | false | nil ] The ssl_verify value parsed out, otherwise nil (and a warning
     #   will be logged).
     def ssl_verify(value)
-      b = bool('tlsAllowInvalidCertificates', value)
+      b = bool('tlsInsecure', value)
```
p-mongo: Am I reading this correctly that setting `ssl_verify: 'bogus'` would produce an error message referencing `tlsInsecure`?

lib/mongo/uri.rb
```diff
@@ -256,6 +256,19 @@ def initialize(string, options = {})
       parsed_scheme, _, remaining = string.partition(SCHEME_DELIM)
       raise_invalid_error!(INVALID_SCHEME) unless parsed_scheme == scheme
       parse!(remaining)
+
+      # Warn if conflicting insecure TLS options are provided.
+      if @uri_options[:ssl_verify] == false
+        if @uri_options[:tls_allow_invalid_certificates] == false
+          log_warn("tlsInsecure/ssl_verify is set to disable verification, but " +
+                     "tls_allow_invalid_certificates is set to enable it")
```
p-mongo: Can this message also include which option takes precedence?

lib/mongo/uri.rb
saghm: `ssl_verify` cannot be used from the URI at all, so it would just give a warning for an unknown URI option

lib/mongo/uri.rb
saghm: None of the other option-parsing methods here have `@api private` so far. I can add it to all of them if you want.

spec/mongo/socket/ssl_spec.rb
```diff
@@ -279,6 +279,42 @@
...
+          expect {
+            described_class.new(
+              address.host,
+              address.port,
+              host_name,
+              30,
+              ::Socket::PF_INET,
+              options.merge(tls_allow_invalid_hostnames: false)
+            ).connect!
+          }.to raise_error(Mongo::Error::SocketError)
```
p-mongo: This assertion should include something about the hostname not passing verification. As written it will be satisfied by failing to DNS resolve `incorrect_hostname`.

spec/mongo/socket/ssl_spec.rb
```diff
@@ -279,6 +279,42 @@
...
+      context 'when the hostname is not verified' do
+        it 'raises an error' do
+          expect {
+            described_class.new(
+              address.host,
+              address.port,
+              host_name,
+              30,
+              ::Socket::PF_INET,
+              options.merge(tls_allow_invalid_hostnames: true, ssl_verify: false)
```
p-mongo: Where does https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst say what should happen when these two options are both provided?

lib/mongo/uri.rb
p-mongo: If the precedence is not currently defined, I would think it is a good idea to define it.

spec/mongo/socket/ssl_spec.rb
saghm: I don't believe this is true; if you look at the implementation of `Mongo::Socket::SSL`, you'll see that the socket is connected with the first two arguments (which are `address.host` and `address.port`), not with the hostname. Furthermore, when changing line 298 to assert `not_to raise_error`, the error raised is specifically the one the driver raises in `Mongo::Socket::SSL#verify_certificate!` (not related to DNS), meaning that the socket connected to the server properly.

p-mongo:
Per the offline discussion:

- Please change the driver options to `ssl_verify_hostname` & `ssl_verify_certificate` (singular and positive)
- Add tutorial docs for the added options
- Update tutorial doc for `ssl_verify` which currently is stated to perform certificate verification
- Document precedence of `ssl_verify` & the two new options (when opposite values are given at the same time)

saghm:
Ready for review again

lib/mongo/socket/ssl.rb
```diff
@@ -114,16 +114,51 @@ def connectable?
...
 
+      def verify_certificate?
+        @verify_certificate ||=
+          # If ssl_verify_certificate is not present, disable only if ssl_verify is
+          # explicitly set to false.
+          if options[:ssl_verify_certificate].nil?
+            options[:ssl_verify] != false
+          # If ssl_verify_certificate is present, enable or disable based on its value.
+          else
+            options[:ssl_verify_certificate]
```
p-mongo: Can this be spelled `!!options[:ssl_verify_certificate]` to convert it to a boolean and possibly a unit test added to verify precedence?

lib/mongo/socket/ssl.rb
```diff
@@ -114,16 +114,51 @@ def connectable?
...
+
+      def verify_hostname?
+        @verify_hostname ||=
+         # If ssl_verify_hostname is not present, disable only if ssl_verify is
+          # explicitly set to false.
+          if options[:ssl_verify_hostname].nil?
+            options[:ssl_verify] != false
+          # If ssl_verify_hostname is present, enable or disable based on its value.
+          else
+            options[:ssl_verify_hostname]
```
p-mongo: Same here.

lib/mongo/uri.rb
```diff
@@ -256,6 +256,19 @@ def initialize(string, options = {})
       parsed_scheme, _, remaining = string.partition(SCHEME_DELIM)
       raise_invalid_error!(INVALID_SCHEME) unless parsed_scheme == scheme
       parse!(remaining)
+
+      # Warn if conflicting insecure TLS options are provided.
+      if @uri_options[:ssl_verify] == false
+        if @uri_options[:ssl_verify_certificate]
+          log_warn("tlsInsecure is set to disable verification, but tlsAllowInvalidCertificates " +
+                     "is set to enable it")
```
p-mongo: Can this say something like "tlsInsecure is set to disable verification, but tlsAllowInvalidCertificates is set to enable it; tlsAllowInvalidCertificates takes precedence" or "Enabling certificate verification due to tlsAllowInvalidCertificates=false setting despite conflicting tlsInsecure=true setting"?

lib/mongo/uri.rb
```diff
@@ -256,6 +256,19 @@ def initialize(string, options = {})
...
+        end
+
+       if @uri_options[:ssl_verify_hostname]
+          log_warn("tlsInsecure is set to disable verification, but tlsAllowInvalidHostnames is " +
+                     "set to enable it")
```
p-mongo: And here.

spec/mongo/socket/ssl_spec.rb
p-mongo: Can we add the message assertion here?

lib/mongo/socket/ssl.rb
saghm: Sure

saghm:
I've made the requested changes

PR #1212, opened 2019-01-14
RUBY-1666 Test on 3.6 server with 3.4 FCV in CI

p-mongo:


saghm:
I'd like to hold off on this until we discuss FCV once more in our weekly meeting (which we didn't do last week since the leads hadn't discussed it yet). I'm still not convinced that there's any benefit to testing servers in lower FCVs given that the driver doesn't actually have any way to detect it.

p-mongo:
This CI configuration reproduces 1578, which https://github.com/mongodb/mongo-ruby-driver/pull/1222 subsequently fixes. I would like to have that fix in 2.7.0.

saghm:
I'd prefer that we instead remove all testing configurations that set servers to lower FCVs. Let's discuss it in our team meeting on Thursday.

PR #1213, opened 2019-01-15
RUBY-1559 Update URI option spec tests and fix comparisons

saghm:
The URI options have been merged, so this PR includes the changes that have been made since the initial commit for them.

Additionally, it appears that the test runner wasn't properly comparing the expected URI options to their parsed options of different names, so I've fixed that as well.

### DIFF
--- a/lib/mongo/uri.rb
+++ b/lib/mongo/uri.rb
@@ -445,8 +445,8 @@ module Mongo
     uri_option 'maxstalenessseconds', :max_staleness, :group => :read, :type => :max_staleness
 
     # Pool options
-    uri_option 'minpoolsize', :min_pool_size
-    uri_option 'maxpoolsize', :max_pool_size
+    uri_option 'minpoolsize', :min_pool_size, :type => :min_pool_size
+    uri_option 'maxpoolsize', :max_pool_size, :type => :max_pool_size
     uri_option 'waitqueuetimeoutms', :wait_queue_timeout, :type => :wait_queue_timeout
 
     # Security Options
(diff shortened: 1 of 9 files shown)


lib/mongo/uri.rb
```diff
@@ -455,8 +455,7 @@ def self.uri_option(uri_key, name, extra = {})
     uri_option 'tlsallowinvalidcertificates', :ssl_verify, :type => :ssl_verify
     uri_option 'tlscafile', :ssl_ca_cert
     uri_option 'tlscertificatekeyfile', :ssl_cert
-    uri_option 'tlsclientkeyfile', :ssl_key
-    uri_option 'tlscertificatekeypassword', :ssl_key_pass_phrase
+    uri_option 'tlscertificatekeyfilepassword', :ssl_key_pass_phrase
```
p-mongo: Per https://jira.mongodb.org/browse/RUBY-1676 I'm thinking this change should also have an accompanying documentation change, no?

saghm: Done

lib/mongo/uri.rb
```diff
@@ -647,6 +646,37 @@ def zlib_compression_level(value)
       nil
     end
 
+    # Parses the max pool size.
+    #
+    # @param value [ String ] The max pool size string.
+    #
+    # @return [ Integer | nil ] The min pool size if it is valid, otherwise nil (and a warning will)
```
p-mongo: The closing paren here is a bit too early in the sentence.

saghm: Oops! Good catch

PR #1217, opened 2019-01-22
Reorder spec config methods

p-mongo:


p-mongo:
Yes, I thought about copying this file to 2.6 branch but for now decided against.

PR #1220, opened 2019-01-22
RUBY-1676 Add missing client option documentation

saghm:
The diff is a little weird since I reordered everything to be in alphabetical order. It might be easier to just look at the final state of the files rather than the diff.


p-mongo:
Is this for https://jira.mongodb.org/browse/RUBY-1676 then?

PR #1227, opened 2019-01-29
RUBY-1522 Add test that reads ignore db/collection readConcern in transaction

p-mongo:


saghm:
I'm still a little surprised that my fix to RUBY-1665 doesn't fix this, as it definitely does cause `#count_documents` to start a transaction. From looking at [this](https://github.com/mongodb/mongo-ruby-driver/blob/67835eb7255bf65897beef2e5dc6ef6646b82fda/spec/support/transactions/operation.rb#L22-L43) and [this](https://github.com/mongodb/mongo-ruby-driver/blob/67835eb7255bf65897beef2e5dc6ef6646b82fda/spec/support/transactions/operation.rb#L111-L132) of the transactions spec, it looks like a little work needs to be done to support `#count_documents` in the first place in order to pass the session argument along. Can you push the version of this that applied the RUBY-1665 change along with the tests included here (either to this PR or to a separate one) so I can take a look at it?

p-mongo:
I removed the dependency of this PR/1522 on 1665, therefore you can approve this PR and then rebase your 1665 fix on master, remove the skipReasons added in this diff for 1665, and fix the remaining test failures.

saghm:
I'm confused because you commented on the RUBY-1665 ticket with test output, but I don't see what code was actually run for that test. Given that the issue in RUBY-1665 is fixed with my pull request, I suspect that the issue here is not actually related to 1665, so I don't see the point of merging this if there's an issue. If I can see the code that you used to run the test that failed, I can see if the issue is in fact in the transactions test code (which is unrelated to RUBY-1665) rather than with the implementation of `#count_documents` itself.

p-mongo:
Feel free to apply your diff on top of this PR, remove the 1665 skipReasons and convince yourself that the result is the same as without your diff applied which is a bunch of test failures.

#1226  does not have any tests of its own, therefore you may either use this PR to test #1226 or create your own tests for #1226 demonstrating the problem that #1226 repaired.

saghm:
> #1226 does not have any tests of its own

Look at the diff again; I pushed a test yesterday afternoon

> If you apply your diff on top of this PR and remove the 1665 skipReasons you will see that the result is the same as without your diff applied which is a bunch of test failures.

As I explained in my previous comment, the transactions tests harness doesn't currently propagate options for `#count_documents`. If you look at the code sections I linked to, you'll see this. Adding code to a test harness for transactions to make operations of a certain type work is completely orthogonal to #1226, so there's no reason to include it there.

p-mongo:
OK, once https://github.com/mongodb/mongo-ruby-driver/pull/1209 is merged I will take another look at what is needed to make those tests pass. In the meantime this PR is a prerequisite for #1209.

PR #1239, opened 2019-02-05
RUBY-1420 Document session options

p-mongo:


spec/support/transactions.rb
```diff
@@ -257,9 +257,7 @@ def setup_test
         coll = support_client[@spec.collection_name]
         coll.database.drop
         coll.with(write: { w: :majority }).drop
-        support_client.command(
-          { create: @spec.collection_name },
-          { write_concern: { w: :majority } })
+        support_client.command(create: @spec.collection_name)
```
saghm: The spec test runner description says that this should be run with write concern majority; is there a reason this was changed?

p-mongo: According to my reading of the code write concern ended up getting passed to sessions where it was ignored. Database#command (which can be seen in this diff) does not document write concern as an accepted option. Therefore as far as I can tell the write concern here doesn't do anything.

saghm: Interesting; it looks like the read/write concern spec says that a write concern needs to be [directly added to the command document](https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#id6). Can we do this here?

p-mongo: Since repairing transaction spec runner does not have much to do with documenting session options, how about doing that in a follow-up PR?

saghm: If that's the case, why was it changed in this PR in the first place?

p-mongo: This PR documents session options. As part of that work I reviewed the callers to see which options they passed. As a result of that audit this change was made where existing code passed an option that had no effect.

saghm: It feels a little strange not to not make the one line fix here when we're already changing the same method call, but okay

PR #1254, opened 2019-02-18
RUBY-1650 Rewrite connection handshake flow to not recurse

p-mongo:


p-mongo:
This PR also removes client key caching, with https://jira.mongodb.org/browse/RUBY-1213 reopened to reinstate it.

lib/mongo/server/connection.rb
```diff
@@ -84,11 +89,10 @@ class Connection
       #
       # @since 2.0.0
       def initialize(server, options = {})
-        @address = server.address
         @monitoring = server.monitoring
         @options = options.freeze
         @server = server
-        @ssl_options = options.reject { |k, v| !k.to_s.start_with?(SSL) }
+        @ssl_options = options.reject { |k, v| !k.to_s.start_with?(SSL) }.freeze
```
saghm: Minor nit: instead of using a double negative of `reject!` and `!`, you can just do `select!`

lib/mongo/address.rb
```diff
@@ -202,9 +197,11 @@ def create_resolver(ssl_options)
       error = nil
       ::Socket.getaddrinfo(host, nil, family, ::Socket::SOCK_STREAM).each do |info|
         begin
-          res = FAMILY_MAP[info[4]].new(info[3], port, host)
-          res.socket(connect_timeout, ssl_options).connect!(connect_timeout).close
-          return res
+          specific_address = FAMILY_MAP[info[4]].new(info[3], port, host)
+          socket = specific_address.socket(
+            connect_timeout, ssl_options, connect_timeout: connect_timeout)
```
saghm: Why doesn't this need to call `connect!` anymore?

p-mongo: Sockets are now completely initialized by their constructors.

lib/mongo/server/connection.rb
p-mongo: Changed.

lib/mongo/address.rb
saghm: From looking at the diff, I don't see any actual changes in functionality to the constructors. Could you point out to me where the changes were made?

p-mongo: https://github.com/mongodb/mongo-ruby-driver/pull/1254/files#diff-ed625ecdabf4aeff5d1a5fc04b958a3aR86 for SSL socket, https://github.com/mongodb/mongo-ruby-driver/pull/1254/files#diff-cc511f0e87683f58982606a59e0f3657R70 for TCP socket, Unix sockets were already completely initialized by their constructor and had a no-op connect! method.

saghm: Hmm, not sure why I missed them. Thanks!

PR #1255, opened 2019-02-18
RUBY-1450 Change test suite to always define all examples

p-mongo:
https://jira.mongodb.com/browse/RUBY-1450

### DIFF
--- a/spec/mongo/auth/cr_spec.rb
+++ b/spec/mongo/auth/cr_spec.rb
@@ -64,7 +64,9 @@ describe Mongo::Auth::CR do
         }.to raise_error(Mongo::Auth::Unauthorized)
       end
 
-      context 'when compression is used', if: testing_compression? do
+      context 'when compression is used' do
+        require_compression
+        min_server_fcv '3.6'
 
         it 'does not compress the message' do
           expect(Mongo::Protocol::Compressed).not_to receive(:new)
--- a/spec/mongo/auth/cr_spec.rb
+++ b/spec/mongo/auth/cr_spec.rb
@@ -79,6 +79,7 @@ describe Mongo::Auth::CR do
   end
 
   context 'when the user is authorized for the database' do
+    max_server_fcv '2.6'
 
     let(:cr) do
       described_class.new(root_user)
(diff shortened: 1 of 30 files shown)


spec/mongo/auth/cr_spec.rb
```diff
@@ -64,7 +64,9 @@
         }.to raise_error(Mongo::Auth::Unauthorized)
       end
 
-      context 'when compression is used', if: testing_compression? do
+      context 'when compression is used' do
+        require_compression
+        min_server_fcv '3.6'
```
saghm: Just to make sure I understand, is the idea that going forward we will only check FCV and not server version?

spec/mongo/database_spec.rb
```diff
@@ -119,7 +119,7 @@
 
     context 'when specifying a batch size' do
 
-      it 'returns the stripped names of the collections', unless: need_to_skip_on_sharded_auth_40? do
+      it 'returns the stripped names of the collections' do
```
saghm: Do we no longer need to skip this test (and the one below) with a sharded 4.0 cluster and auth?

spec/support/constraints.rb
```diff
@@ -97,11 +91,77 @@ def require_ssl
...
+
+  def min_ruby_version(version)
+    before do
+      if RUBY_VERSION < version
+        skip "Ruby version #{version} or higher required"
+      end
+    end
+  end
+
+  def ruby_version_lt(version)
```
saghm: Can we change the name of either this or `min_ruby_version`? They're both doing the inverse of each other, so I think it would make sense to either rename this something like `max_ruby_version_exclusive` or to rename `min_ruby_version` `ruby_version_gle`.

spec/mongo/auth/cr_spec.rb
p-mongo: There are several failures in the suite if the tests for, say, 3.6 functionality are run on a 3.6 server with 3.4 FCV because the produced errors are different than what a 3.4 server would produce. Requiring FCV instead of server version bypasses these issues. In other words, even though there is support in the test suite to test against clusters where server version != FCV, this is mostly not done currently and thus the min requirement is expressed in FCV and the max requirement in server version.

spec/mongo/database_spec.rb
p-mongo: These tests do pass currently and I was unable to determine why this skip is needed.

spec/support/constraints.rb
p-mongo: Changed to ruby_version_gte.

PR #1257, opened 2019-02-19
RUBY-1727 Unscope connection pools from clusters

p-mongo:


lib/mongo/server.rb
```diff
@@ -274,7 +275,16 @@ def summary
     #
     # @since 2.0.0
     def pool
-      @pool ||= cluster.pool(self)
+      @pool_lock.synchronize do
```
saghm: It doesn't look like the pool will ever be finalized unless the server itself is; why not just define a finalizer on server itself (or on all pools in the pool implementation) instead of defining one on the pool here?

p-mongo: Instance variables of an object being finalized are not available. What exactly are you proposing to do?

saghm: I understand that. Let me rephrase my question; is there any reason why we only want to define finalizers to clean up connection pools that are instance variables of servers instead of just defining one in either the connection pool constructor or `ConnectionPool#get`?

p-mongo: Having the finalizer be set by the connection pool constructor is a better idea indeed.

PR #1258, opened 2019-02-19
Fix RUBY-1647 Reset connection pool on non-timeout network errors

p-mongo:


spec/integration/sdam_error_handling_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+  before(:all) do
+    ClientRegistry.instance.close_all_clients
+  end
+
+  describe 'when there is an error during an operation' do
+    let(:client) { authorized_client }
+
+    before do
+      wait_for_all_servers(client.cluster)
+      # we also need a connection to the primary so that our error
```
saghm: It's not clear to me at all what this comment has to do with the code after it; what does needing a connection to the primary have to do with pinging the server and stopping the monitors?

spec/support/client_registry.rb
```diff
@@ -95,13 +95,19 @@ def new_global_client(name)
       )
     # Provides an authorized mongo client that retries writes.
     when 'authorized_with_retry_writes'
-      global_client('authorized').with(retry_writes: true).tap do |client|
+      global_client('authorized').with(
+        retry_writes: true,
+        server_selection_timeout: 4.97,
```
saghm: How were these server selection timeouts (4.97 and 4.99) chosen? Given how specific they are, I think it makes sense to comment why they are what they are.

lib/mongo/error/operation_failure.rb
```diff
@@ -28,7 +29,7 @@ class OperationFailure < Error
       # @api private
       WRITE_RETRY_ERRORS = [
         {:code_name => 'InterruptedAtShutdown', :code => 11600},
-        {:code_name => 'InterruptedDueToReplStateChange', :code => 11602},
+        {:code_name => 'InterruptedDueToStepDown', :code => 11602},
```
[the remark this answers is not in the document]
p-mongo: See https://github.com/mongodb/specifications/pull/496.

spec/support/client_registry.rb
p-mongo: They are numbers which can be traced to where in the code they are defined, as opposed to a number like `5` which is hard to grep for. The fractions chosen are arbitrary.

lib/mongo/error/operation_failure.rb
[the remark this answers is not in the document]
saghm: Ah, makes sense

spec/integration/sdam_error_handling_spec.rb
p-mongo: Ping establishes a connection to the primary, the comment goes with the ping. Stopping the monitors is separate. Ultimately this is all done to ensure assertions are performed on the right connections (non-monitoring vs monitoring).

p-mongo:
@saghm Anything else for this PR?

PR #1260, opened 2019-02-19
RUBY-1704 Create collections with write concern majority in transacti…

saghm:
…ons spec tests

### DIFF
--- a/spec/support/transactions.rb
+++ b/spec/support/transactions.rb
@@ -257,7 +257,7 @@ module Mongo
         coll = support_client[@spec.collection_name]
         coll.database.drop
         coll.with(write: { w: :majority }).drop
-        support_client.command(create: @spec.collection_name)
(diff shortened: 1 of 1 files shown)


saghm:
I ran the transactions tests locally to ensure that nothing failed; the evergreen run is ongoing

p-mongo:
I meant that this is the correct way of passing the write concern, since tests were passing without write concern being passed as well.

saghm:
The driver doesn't do any handling of the write concern here; it's an option for the [`create` command](https://docs.mongodb.com/manual/reference/command/create/), so it gets passed verbatim to the server. If you're asking if the driver correctly passes options passed in command documents to the server, then yes, it does.

p-mongo:
OK thanks.

p-mongo:
@saghm Merge?

saghm:
Will do

PR #1274, opened 2019-02-21
RUBY-1728 Indicate which auth mechanism was used when auth fails

p-mongo:


lib/mongo/auth/cr/conversation.rb
```diff
@@ -130,7 +130,9 @@ def initialize(user)
         private
 
         def validate!(reply)
-          raise Unauthorized.new(user) if reply.documents[0][Operation::Result::OK] != 1
+          if reply.documents[0][Operation::Result::OK] != 1
```
saghm: When discussing this offline, we were talking about how it might be useful to be able to see the entire reply document when this fails. Is there a reason we don't want to log it here (e.g. leaking sensitive data)?

p-mongo: I do not recall this specific conversation but https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst#security requires redacting conversation requests/responses in command monitoring, therefore logging complete responses would go against spec requirements.

PR #1285, opened 2019-02-25
RUBY-1563 Abstract SRV record query behavior for use in mongos polling

saghm:
This was tested with a manual evergreen patch due to the queue apparently being backed up


lib/mongo/srv/records.rb
```diff
@@ -0,0 +1,72 @@
...
+# distributed under the License is distributed on an 'AS IS' BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'resolv'
+
+module Mongo
+
+  module SRV
```
p-mongo: Since `Mongo::SRV` is not an existing namespace, the new namespace should be called `Mongo::Srv`.

lib/mongo/srv/records.rb
```diff
@@ -0,0 +1,72 @@
...
+    # This SRV::Records class is used to keep track of the SRV records discovered for a given hostname. It also keeps
+    # track of the minimum TTL of the records added so far.
+    #
+    # @api private
+    class Records
+
+      MISMATCHED_DOMAINNAME = "Parent domain name in SRV record result (%s) does not match " +
+                                 "that of the hostname (%s)".freeze
+
+      attr_reader :hostname
```
p-mongo: Even though this entire class is private, please add docstrings with types for each attribute and a description of what they store as well as docstrings for the methods with descriptions of what they do and types for their parameters.

lib/mongo/srv/resolver.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    # The SRV::Resolver class encapsulates the necessary behavior for querying SRV records.
+    #
+    # @api private
+    class Resolver
+      include Loggable
+
+      RECORD_PREFIX = '_mongodb._tcp.'.freeze
+
+      def initialize(raise_on_invalid = false)
```
p-mongo: Why is this an option, given that clients can rescue exceptions?

lib/mongo/srv/resolver.rb
```diff
@@ -0,0 +1,65 @@
...
+        @raise_on_invalid = raise_on_invalid
+        @resolver = Resolv::DNS.new
+      end
+
+      def get_records(hostname)
```
p-mongo: Is a trailing dot in hostname allowed, not allowed, required?

lib/mongo/srv/records.rb
saghm: I think it's confusing to have "SRV" in different case in two different namespaces in the driver, but ok

lib/mongo/srv/resolver.rb
saghm: The polling SRV records spec indicates that warnings should only be logged rather than errors raised. If we only catch the error that's raised externally, then an error with one record will prevent the driver from checking the rest of the records, which might be fine.

lib/mongo/srv/resolver.rb
saghm: I don't think the spec says anything about this, and I don't understand what that question has to do with the section of the code you commented on

p-mongo: The question has to do with how DNS works. DNS records always have trailing dots, e.g. `www.mongodb.com.`. Virtually in all other usage trailing dots are omitted, i.e. we talk about `www.mongodb.com`. Since the code in question deals with DNS records and performs hostname matching, which splits the hostnames on dots, it is important that trailing dots be correctly handled. It is also important for calling code to know whether `hostname` should or should not have the trailing dot.

saghm: All of the code related to parsing the hostname and doing the DNS queries is copied from the existing implementation in `mongo/uri/srv_protocol.rb`; all this PR does is factor it out into separate classes. I think it might be useful for you to read the existing implementation so you understand what's actually being changed here.

lib/mongo/srv/resolver.rb
p-mongo: Then why is there an option to raise exceptions? It also has no test coverage.

saghm: The initial seedlist discovery spec requires that an error is raised if any invalid records are found, whereas the polling SRV records for mongos spec requires that errors are not raised. We need to be able to do different behavior for each of them, but most of the code is the same, which is why this is being factored out in the first place. The only reason I separated out this code is that you asked me to make a separate PR, but it seems like it's creating more work than it's saving.

p-mongo:
Superseded by https://github.com/mongodb/mongo-ruby-driver/pull/1426.

PR #1287, opened 2019-02-26
Update URI options spec tests with changes for conflicting TLS options

saghm:


spec/spec_tests/uri_options_spec.rb
```diff
@@ -40,7 +40,7 @@
             end
           end
 
-          context 'when the uri should not warn', if: !test.warn? do
+          context 'when the uri should not warn', if: !test.warn? && test.valid? do
```
p-mongo: Why is the `warn?` condition insufficient here?

saghm: Since the spec changes add tests that are expected to raise errors (but not warn), the test will still be run and raise an error (on line 50) if we don't also check that the test is valid.

PR #1288, opened 2019-02-26
RUBY-1749 Reject two-part SRV hostnames with a trailing dot

saghm:


p-mongo:
Please add a docstring for `validate_hostname!` stating what it validates, as well as its policy on trailing dots. It should either always accept or always reject trailing dots, and this should be tested with inputs like `foo.bar.one.` and `foo.bar.one`, unless this somehow breaks the rest of the driver.

Please add unit tests for `validate_hostname!` with each of the cases that it checks.

p-mongo:
Other inputs currently accepted (i.e. considered valid) by this method:

- `.foo.bar`
- `...`
- `a...`
- `...a`
- `aaa...aaaa`

While a full hostname parser is not needed, we can easily verify that:

- each component is non-empty, and reject a hostname with empty components aka sequential dots
- leading dots are rejected (assuming this is what should happen)
- trailing dots are rejected (assuming this is what should happen)

lib/mongo/uri/srv_protocol.rb
```diff
@@ -103,12 +103,23 @@ def parse_creds_hosts!(string)
...
 
+      # Validates that a hostname fits the expected format.
+      #
+      # @param [ String ] hostname The hostname to validate.
+      #
+      # @raise [ Mongo::Error::InvalidURI ] If any of the following conditions are true:
+      #   * the hostname is empty
+      #   * the hostname contains either ',' or ':'
+      #   * the hostname begins with '.'
+      #   * the hostname contains consecutive '.' characters
```
p-mongo: Please also state that the trailing dots are not allowed.

spec/mongo/uri/srv_protocol_spec.rb
```diff
@@ -940,5 +951,99 @@
...
+        Mongo::URI::SRVProtocol.new("#{scheme}#{valid_hostname}/")
+      end
+
+      let(:validate) do
+        dummy_uri.send(:validate_hostname!, hostname)
+      end
+
+      context 'when the hostname is valid' do
+        let(:hostname) do
+          valid_hostname
```
p-mongo: Please also test that `foo` is a valid hostname.

lib/mongo/uri/srv_protocol.rb
saghm: That's incorrect; trailing dots are not forbidden by the spec, which is why there is a test that verifies it works

saghm:
I realized that I accidentally removed the condition that the hostname must have at least three parts, which required me to rewrite most of the unit tests I had added.

p-mongo:
Superseded by https://github.com/mongodb/mongo-ruby-driver/pull/1296.

PR #1290, opened 2019-02-27
RUBY-1706 Support Retryable Writes on by Default

p-mongo:


spec/mongo/client_construction_spec.rb
```diff
@@ -155,18 +155,42 @@
...
+            expect(client.options['retry_writes']).to be false
+          end
         end
 
-        it 'sets the option' do
-          expect(client.options['retry_writes']).to eq(options[:retry_writes])
+        context 'when retry_writes is not given' do
+
+          let(:options) do
+            { retry_writes: true }
```
saghm: This test looks identical to `when retry_writes is true`. I think it should set the options to the empty hash?

p-mongo: Good catch, fixed.

PR #1297, opened 2019-03-08
Do not restart monitors when monitoring io is false

p-mongo:


p-mongo:
Extracted from https://github.com/mongodb/mongo-ruby-driver/pull/1200.

PR #1298, opened 2019-03-08
Move snakeize methods to utility module

p-mongo:


p-mongo:
Extracted from https://github.com/mongodb/mongo-ruby-driver/pull/1200.

PR #1319, opened 2019-03-15
RUBY-1774 use add_file instead of adding a Certificate object

sgirones:
**Summary**
Use the `add_file` method in order to parse all the certificates in the file, not only the first one.

This is needed if the CA cert file contains an intermediate CA and a root CA

**Notes**
I want to add tests but I would like to know if you have a specific way to generate the certificates in https://github.com/mongodb/mongo-ruby-driver/tree/master/spec/support/certificates

If not, I can do it myself and add them to this PR.

TODO:
- [ ] Generate a new server certificate using an intermediate certificate
- [ ] Add SSL validation tests using the new intermediate certificate


p-mongo:
Thank you for the PR. We do have a tool that generates the certificates, but I'm thinking to leave the existing single-cert bundle in place and add another bundle for testing a cert chain. Can you generate those and put them in `spec/support/certificates/chain`?

Also would you mind adding a note to the tutorial stating that `:ssl_ca_cert` option accepts path to a file with CA cert as well as any intermediate certificates, as well as whether the intermediate certificates should be given before or after the CA cert.

sgirones:
sounds good 👍

sgirones:
@p-mongo I created the new certificates for the new tests, but I may need some context regarding the automated tests.

My idea is to create a new `mongod` process that uses the new certificate chain, and then point the new tests to that instance.

I'm trying to find where the main mongod + TLS process is configured, but can't find it in the .travisci.yml. I only found this https://github.com/mongodb/mongo-ruby-driver/blob/master/.travis.yml#L15, but it doesn't enable TLS.

Do you use some infrastructure not defined in this repo?

p-mongo:
Are you able to run (and pass) the test you are writing when launching the server following instructions in https://github.com/mongodb/mongo-ruby-driver/tree/master/spec#tls-without-verification, using your generated certificates ?

TLS is tested in Evergreen and it is configured via [mongo-orchestration](https://github.com/10gen/mongo-orchestration). I think ultimately I will need to either generate the intermediate cert using existing CA in the test suite or rebuild all certificates from the same CA, but we also have some more work to do related to the TLS tests performing certificate verification.

sgirones:
I added a single test to demonstrate that my change works. The commands I used to validate it locally are:
> $ mlaunch init --single --dir /tmp/mdb-ssl --sslMode requireSSL --sslPEMKeyFile $(pwd)/spec/support/certificates/chain/server.pem --sslCAFile $(pwd)/spec/support/certificates/chain/ca-chain.pem --sslClientCertificate $(pwd)/spec/support/certificates/chain/client.pem --setParameter enableTestCommands=1

> $ MONGODB_URI='mongodb://localhost:27017/?tls=true' rake spec SPEC=spec/mongo/socket/ssl_spec.rb SPEC_OPTS="-e 'when a CA certificate chain is provided'"

I added a single test as a PoC. Without knowing how the infrastructure side will look like, it's difficult to come up with relevant tests.

I'm not sure how to proceed here. I'm happy to help on whatever you need from me 👍

p-mongo:
Thank you, I will get back to you regarding the tests. Are you able to make the documentation changes mentioned? I am also curious whether the certificates can be given in any order (i.e. ca first or intermediate first).

sgirones:
I just tested running the tests with the ca-chain being: `ca + intermediate` and the test pass. So no issues there.

Also, I was checking the documentation in `docs/tutorials/ruby-driver-create-client.txt` and it already says:
```
``:ssl_ca_cert``
The file path containing concatenated certificate authority certificates used to validate certs
       passed from the other end of the connection
```
It looks like this is a bug fix instead of a new feature. I can rephrase the documentation, but I think it's well explained already.

p-mongo:
Thank you for the patch. As I mentioned in the comment on the ticket (https://jira.mongodb.org/browse/RUBY-1774?focusedCommentId=2268947&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2268947), this PR is superseded by https://github.com/mongodb/mongo-ruby-driver/pull/1367 for certificate chain support and https://github.com/mongodb/mongo-ruby-driver/pull/1368 for multiple CA certificate support.

PR #1326, opened 2019-03-19
RUBY-1778 Collection should inherit read concern from Client

saghm:


lib/mongo/session.rb
```diff
@@ -874,6 +874,21 @@ def cluster
...
+    #
+    # This is a driver style hash with underscore keys.
+    #
+    # @example Get the session's transaction read concern.
+    #   session.txn_read_concern
+    #
+    # @return [ Hash ] The read concern used for starting transactions.
+    #
+    # @since 2.9.0
+    def txn_read_concern
```
p-mongo: Was this method intentionally made part of the public API, and if so, what was the reason behind this? It does not seem to me like a method that applications need to be invoking.

saghm: Yes; it's used elsewhere in the driver besides here, and `txn_read_preference` is already part of the public API, so I didn't see why having this also be part of it would be controversial.

p-mongo: Since it is difficult to modify and remove methods which are part of public API, if there is no compelling reason to make a method a part of public API such method should not be part of public API. If a user requests this method be part of public API because they want to use it outside of the driver, let's make it public then; until such a request let's keep this method in private API.

saghm: I don't really understand why we added `txn_read_preference` as part of the public API but don't want to add `txn_read_concern`, but you seem to feel more strongly about this than I do, so I'll remove it.

p-mongo: Looking at the transactions PR which added `txn_read_preference`, many of the methods added there, and not marked api private, appear to be internal to the driver - even more so than `txn_read_preference` (e.g. `add_autocommit!`).

PR #1328, opened 2019-03-21
RUBY-1780 More strict verification in crud spec tests

p-mongo:


lib/mongo/protocol/msg.rb
```diff
@@ -82,10 +88,26 @@ def replyable?
       #
       # @since 2.5.0
       def payload
+        # Reorder keys in global_args for better logging - see
+        # https://jira.mongodb.org/browse/RUBY-1591?filter=25988.
```
saghm: No need to put `?filter=25988` here; I think that's specific to the JIRA session you were in, as when I go to that link, it redirects to the URL without the querystring.

PR #1330, opened 2019-03-21
RUBY-1781 Refactor crud spec runner/tests to support lists of operations

p-mongo:


spec/support/crud.rb
```diff
@@ -130,25 +130,29 @@ def initialize(data, test)
...
       #
       # @return [ Result, Array<Hash> ] The result(s) of running the test.
       #
       # @since 2.0.0
-      def run(collection)
+      def run(collection, num_ops)
         result = nil
-        @operations.each do |op|
-          result = op.execute(collection)
+        1.upto(num_ops) do |i|
```
saghm: This feels like a really strange way to implement this; instead of just running each operation once and checking the result after each one, we're instead running the first operation, then the first two (including the one we just did), then the first three (including the two we just did), etc. I feel like refactoring this a bit more rather than calling run N times is worth it to avoid running twice as many operations.

p-mongo: This is done intentionally to make it possible to run each example individually. SDAM tests currently run each operation once and this makes them hard to debug (because one has to skip over assertions for each operation) and awkward to parallelize (if the entire file is not run sequentially by the same process the run would fail).

saghm: Fair enough

PR #1332, opened 2019-03-25
RUBY-1668 Support "startAfter" option in the $changeStream stage

saghm:


spec/integration/change_stream_spec.rb
```diff
@@ -366,4 +366,57 @@ def clear_fail_point_before
...
+      authorized_collection.insert_one(x: 1)
+      authorized_collection.client.use(:admin).command(fail_point_base_command.merge(
+        :mode => {:times => 1},
+        :data => {:failCommands => ['getMore'], errorCode: 100}))
+      stream.to_enum.next
+
+      subscriber.started_events.select { |e| e.command_name == 'aggregate' }
+    end
+
+    it 'sends startAfter on the initial aggregation when passed in' do
```
p-mongo: This test only concerns itself with one change stream operation (initial aggregation), but it triggers three operations (initial aggregation, getmore which fails, another initial aggregation). This is not described anywhere in the prose which will make this test unnecessarily difficult to work with in the future. What do you think about making this test perform the one operation it requires and scoping the full set of three operations to the other test where they are tested?

saghm: The spec requires that we send `startAfter` on initial aggregations but not when resuming. Without having a failed getMore in between and then doing another initial aggregation, there's no way to test the resume case. While we could run separate sequences of operations for this test and the one below, it seems unnecessary when a set of operations (initial agg, failed getMore, and resume) covers all the cases that we need. I don't think I understand why this would make the test difficult to work with.

p-mongo: I think this makes a lot of sense for the second example, but there may be a way to perform all assertions that the first example currently has without the first example performing unrelated unnecessary operations.

PR #1333, opened 2019-03-26
RUBY-1767 set client key from tlsCertificateKeyFile

saghm:


p-mongo:
Since `spec/integration/ssl_uri_options.rb` is missing the _spec suffix, does this test ever get run in CI?

saghm:
I think so, but I'll push up a change just to make sure

saghm:
The test is definitely running now, and Evergreen doesn't report it as failing.

spec/integration/ssl_uri_options_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  require_no_auth
+  require_topology :single
+  require_ssl
+
+  let(:hosts) do
+    SpecConfig.instance.addresses.join(',')
+  end
+
+  let(:uri) do
+    "mongodb://#{hosts}/?tls=true&tlsInsecure=true&tlsCertificateKeyFile=#{SpecConfig.instance.client_cert_key_pem}"
```
p-mongo: Since we have a set of certificates in the test suite, why is`tlsInsecure=true` given here?

saghm: For the same reason that we use `ssl_verify: false` for the existing SSL tests: https://github.com/mongodb/mongo-ruby-driver/blob/f445b789e38b710ae77d81b514a6a5f3842f3a4c/spec/support/spec_config.rb#L230

p-mongo: Then, since there appear to be no assertions that the key in URI actually is used, if the driver ignored certificate option(s) entirely, would it still pass this test?

saghm: No; `tlsInsecure` is just the combination of `tlsAllowInvalidCertificates` and `tlsAllowInvalidHostnames`. `tlsAllowInvalidCertificates` doesn't mean that certificates don't have to be specified _at all_; it affects client-side validation of the server's certificate, but the server still expects a certificate of from the client. In this specific case, I suspect the original reason the option was provided to the driver in the case I linked above is that the server certificate is self-signed (meaning it would fail to validate if `tlsAllowInvalidCertificates` weren't present) as well as the possibility that the server's certificate might not be issued for the correct hostname (e.g. if an evergreen host did not have an alias for `localhost` in its hosts file, meaning it would fail if `tlsAllowInvalidHostnames` weren't present).

I don't really understand why we're discussing that these tests use `tlsInsecure` when literally all of the existing TLS tests we have in the driver use the same option (as I linked to above). If we think that's wrong for some reason, that sounds like a discussion potentially leading to a different ticket, not something that should be changed in this ticket, which is narrowly scoped to hooking up one specific URI option to one specific client option.

p-mongo: https://jira.mongodb.org/browse/RUBY-1767 says:

> If the current URI options are sufficient to configure TLS connections with verification, this needs to be tested in the test suite and documented. 

Acceptance criteria for RUBY-1767 is a test which connects to mongod with TLS verification, with certificates being provided via URI options. The test in this PR, since it disables TLS verification, does not fulfill the acceptance criteria.

PR #1334, opened 2019-03-27
RUBY-1562 Remove -1 limit in db view

p-mongo:
As far as I can tell the -1 limit is ignored by the driver, just like
a nil limit would be ignored. However the call on server is problematic
with the new retryable reads API since server may change during a
read operation.

### DIFF
--- a/lib/mongo/database/view.rb
+++ b/lib/mongo/database/view.rb
@@ -51,7 +51,6 @@ module Mongo
       def collection_names(options = {})
         @batch_size = options[:batch_size]
         server = next_primary(false)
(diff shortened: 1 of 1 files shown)


p-mongo:
Existing code changes what command is sent to the server based on the server the command is (going to) be sent to. With retryable reads, the server for a retry may be different than the server for the first attempt, in particular server version may change. Thus the command that was constructed for the first server may not be appropriate for the second server. Luckily, as far as I can tell the limit of -1 is no different from limit being nil, thus this PR removes the limit being set to -1 in some circumstances.

PR #1335, opened 2019-03-28
RUBY-1562 Retryable Reads implementation

p-mongo:
https://jira.mongodb.com/browse/RUBY-1562

### DIFF
--- a/lib/mongo/client.rb
+++ b/lib/mongo/client.rb
@@ -239,8 +239,9 @@ module Mongo
     #   in which reads on a mongos are retried.
     # @option options [ Symbol ] :replica_set The name of the replica set to
     #   connect to. Servers not in this replica set will be ignored.
-    # @option options [ true | false ] :retry_reads Retry reads once
-    #   (modern retry behavior)
+    # @option options [ true | false ] :retry_reads If true, modern retryable
+    #   reads are enabled (which is the default). If false, modern retryable
+    #   reads are disabled and legacy retryable reads are enabled.
     # @option options [ true | false ] :retry_writes Retry writes once when
     #   connected to a replica set or sharded cluster versions 3.6 and up.
     #   (Default is true.)
--- a/lib/mongo/client.rb
+++ b/lib/mongo/client.rb
@@ -26,7 +26,7 @@ module Mongo
     # subcomponents.
     #
     # @since 2.1.0
-    CRUD_OPTIONS = [ :database, :read, :write, :retry_writes ].freeze
+    CRUD_OPTIONS = [ :database, :read, :write, :retry_reads, :retry_writes ].freeze
 
     # Valid client options.
     #
@@ -56,6 +56,7 @@ module Mongo
       :read_concern,
       :read_retry_interval,
       :replica_set,
+      :retry_reads,
       :retry_writes,
       :scan,
       :sdam_proc,
@@ -170,7 +171,7 @@ module Mongo
     #   # connect to the replica set if given the address of a server in
     #   # a replica set
     #
-    # @param [ Array<String>, String ] addresses_or_uri The array of server addresses in the
+    # @param [ Array<String> | String ] addresses_or_uri The array of server addresses in the
     #   form of host:port or a MongoDB URI connection string.
     # @param [ Hash ] options The options to be used by the client.
     #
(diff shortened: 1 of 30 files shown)


lib/mongo/collection/view/change_stream.rb
```diff
@@ -270,17 +270,22 @@ def create_cursor!
...
+          start_at_operation_time = nil
+          @cursor = read_with_retry_cursor(session, server_selector, view) do |server|
+            result = send_initial_query(server, session)
+            if doc = result.replies.first && result.replies.first.documents.first
+              start_at_operation_time = doc['operationTime']
+            else
+              # The above may set @start_at_operation_time to nil
+              # if it was not in the document for some reason,
+              # for consistency set it to nil here as well.
+              # NB: since this block may be executed more than once, each
```
saghm: Will the block be executed more than once if the first branch of the `if/else` is taken? My assumption is that if the driver successfully receives a reply with an operation time, then the read was successful and it won't be retried. If that's the case, why do you we need to set the operation time to `nil` here if it's set to `nil` above the call to `read_with_retry_cursor`?

lib/mongo/retryable.rb
```diff
@@ -163,6 +208,52 @@ def write_with_retry(session, write_concern, ending_transaction = false, &block)
...
+      server = select_server(cluster, server_selector)
+      begin
+        yield server
+      rescue Error::SocketError, Error::SocketTimeoutError => e
+        if session.in_transaction?
+          raise
+        end
+        retry_read(e, server_selector, &block)
+      rescue Error::OperationFailure => e
+        if session.in_transaction? || !e.write_retryable?
```
saghm: Should this be `!e.read_retryable?`

lib/mongo/collection/view/change_stream.rb
p-mongo: It is possible that the cursor creation might raise an error which will cause the read to be retried, for example if a future version of the driver decides to communicate with the server during cursor creation. Since an assignment is inexpensive, being defensive and clearing `start_at_operation_time` makes this method not reliant on behavior elsewhere in the driver to function correctly.

lib/mongo/retryable.rb
p-mongo: No, per the spec in https://github.com/mongodb/specifications/blob/master/source/retryable-reads/retryable-reads.rst#retryable-error the set of retryable errors is the same for reads and writes, hence there is no `read_retryable?` method to invoke.

lib/mongo/collection/view/change_stream.rb
saghm: I'm not sure how raising an error during cursor creation could cause this block to have `start_at_operation_time` populated with a non-nil value (given that the error would be raised during the `result.replies.first && result.replies.first.documents.first` check and therefore wouldn't even get to the part where the operation time is set, and I don't think "relying on behavior elsewhere in a the driver" is a very meaningful description of "relying on the condition of the `if` branch that this `else` block is attached to, but I doubt we'll come to any agreement about this, and it's not super important.

lib/mongo/retryable.rb
saghm: The [documentation for `write_retryable?`](https://github.com/mongodb/mongo-ruby-driver/blob/ba8b70edbd660c4ec6cefd6718140aa33427b9a0/lib/mongo/error/operation_failure.rb#L93) says "Can the write operation that caused the error be retried?", and the [documentation for `retryable?`](https://github.com/mongodb/mongo-ruby-driver/blob/ba8b70edbd660c4ec6cefd6718140aa33427b9a0/lib/mongo/error/operation_failure.rb#L81) says "Can the read operation that caused the error be retried?". It's pretty confusing that we wouldn't use the method specifically documented to be used for read retries in the method that retries reads. Can you elaborate on why using `retryable?` wouldn't work?

p-mongo: `retryable?` deals with the legacy retryable read criteria and as a result is not applicable to modern retryable reads or writes.

saghm: This PR should update the documentation on that method then, since it's not at all clear from the current language

p-mongo: Made the requested changes.

lib/mongo/client.rb
```diff
@@ -238,10 +239,12 @@ def hash
     #   in which reads on a mongos are retried.
     # @option options [ Symbol ] :replica_set The name of the replica set to
     #   connect to. Servers not in this replica set will be ignored.
-    # @option options [ true, false ] :retry_writes Retry writes once when
+    # @option options [ true | false ] :retry_reads Retry reads once
```
saghm: The tutorial documentation above indicates that setting this to false will enable legacy retryable reads, which isn't obvious from the documentation here. I think we should add a sentence explaining that setting it to false will enable legacy retryable reads.

lib/mongo/retryable.rb
saghm: Thanks!

lib/mongo/client.rb
p-mongo: OK, what language would you suggest?

saghm: `:retry_reads If true, modern retryable writes are enabled (which is the default). If false, modern retryable writes are disabled and legacy retryable writes are enabled`

p-mongo: Made the suggested change, fixed to refer to retryable reads rather than retryable writes.

PR #1338, opened 2019-04-04
RUBY-1784 Document that retryable writes are now on by default


p-mongo:


docs/tutorials/ruby-driver-create-client.txt
```diff
@@ -722,30 +722,11 @@ Details on Retryable Writes
...
-elections and similar events, but will not retry writes on network errors
-like timeouts.
-
-Note that writes are always retried on transient server status-related
-errors like "not master" and "node is recovering", regardless of what
-``retry_writes`` option is set to.
-
-The following write methods used in day-to-day operations on collections will
-be automatically retried:
+As of driver version 2.9.0, the modern mechanism is used by default, and the
```
saghm: Am I understanding correctly that on server versions below 3.6 will have no retryable writes (either modern or legacy) by default, and legacy retryable writes will only be enabled on those servers if the `retryWrites: false` option is specified? It's not super clear to me from the changes in the documentation whether this is the case or if legacy retryable writes are enabled by default for older servers still, so I think we should explicitly state whichever way it is.

p-mongo: That is correct, what would you suggest as more clear language?

saghm: I'd change the sentence to "As of driver version 2.9.0, the modern mechanism is used by default on servers that support it, and the legacy mechanism is deprecated and disabled by default on all server versions."

p-mongo: Thank you, made this change.

PR #1342, opened 2019-04-09
RUBY-1626 Knob to turn off legacy retryable writes


p-mongo:


lib/mongo/client.rb
```diff
@@ -368,9 +374,45 @@ def cluster_options
         server_selection_semaphore: @server_selection_semaphore,
         # but need to put the database back in for auth...
         database: options[:database],
+
+        # Put these options in for legacy compatibility, but note that
+        # their values on the client and the cluster do not have to match -
```
saghm: What circumstances would cause the values on the client and cluster not to match? I don't think I understand why that would happen

p-mongo: Calling `with` on a client and changing these options would cause the new client and its cluster to have different values.

saghm: Ah, right

PR #1343, opened 2019-04-11
RUBY-1562 Implement command assertions for retryable reads spec tests


p-mongo:


lib/mongo/grid/fs_bucket.rb
```diff
@@ -147,6 +147,7 @@ def insert_one(file)
       #   size.
       # @option options [ String ] :write The write concern.
       # @option options [ String ] :read The read preference.
+      # @option options [ Session ] :session The session to use.
```
saghm: It's not clear to me how the gridfs changes are related to the purpose of this ticket. Did the retryable read assertions uncover bugs in the existing gridfs implementation that these changes fix?

spec/support/crud.rb
```diff
@@ -22,11 +22,76 @@
 require 'support/crud/write'
 require 'support/crud/verifier'
 
+def crud_execute_operations(spec, test, num_ops, event_subscriber, expect_error,
+  client
+)
+  cache_key = "#{test.object_id}:#{num_ops}"
+  $crud_result_cache ||= {}
+  $crud_result_cache[cache_key] ||= begin
```
saghm: Just to make sure I understand, the purpose of this cache is to avoid running the same test more than once and instead just reuse the results from the first run?

lib/mongo/grid/fs_bucket.rb
p-mongo: The implementation was not compliant with the spec tests added for retryable reads.

spec/support/crud.rb
p-mongo: That is correct.

lib/mongo/grid/fs_bucket.rb
saghm: Sounds good, thanks

spec/support/crud.rb
saghm: Awesome, thanks

PR #1345, opened 2019-04-11
Remove START_TRANSACTION and start_transaction


p-mongo:


p-mongo:
Based on the Evergreen run it appears that this constant and method are not used by anything in the driver.

PR #1347, opened 2019-04-16
RUBY-1792 Use runOn syntax to specify transaction test requirements


p-mongo:


spec/support/transactions.rb
```diff
@@ -175,6 +178,21 @@ def event_subscriber
       def run
         test_client.subscribe(Mongo::Monitoring::COMMAND, event_subscriber)
 
+        $distinct_ran ||= if @ops.any? { |op| op.name == 'distinct' }
```
saghm: What's different about distinct that requires special handling?

p-mongo: This behavior is mandated by the spec.

saghm: Can you share a link to where this is written? From looking at RUBY-1792, the only spec commit mentioned in the linked drivers ticket is https://github.com/mongodb/specifications/commit/2d2003825f2a11f360440d1b77371cff8370bf82, and I can't find anything there about distinct.

p-mongo: This requirement is described in https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#why-do-tests-that-run-distinct-sometimes-fail-with-staledbversion.

PR #1351, opened 2019-04-26
RUBY-1572 Connections survive primary step down

p-mongo:
Update driver to match current specification

### DIFF
--- a/spec/integration/step_down_spec.rb
+++ b/spec/integration/step_down_spec.rb
@@ -3,7 +3,7 @@ require 'spec_helper'
 describe 'Step down behavior' do
   require_topology :replica_set
 
-=begin Uncomment to shorten the runtime of this test
+=begin Uncommenting this will reduce the runtime of the test, but the spec currently does not allow us to do it
   before(:all) do
     # These before/after blocks are run even if the tests themselves are
     # skipped due to server version not being appropriate
--- a/spec/integration/step_down_spec.rb
+++ b/spec/integration/step_down_spec.rb
@@ -2,8 +2,8 @@ require 'spec_helper'
 
 describe 'Step down behavior' do
   require_topology :replica_set
-  min_server_fcv '4.2'
 
+=begin Uncomment to shorten the runtime of this test
   before(:all) do
     # These before/after blocks are run even if the tests themselves are
     # skipped due to server version not being appropriate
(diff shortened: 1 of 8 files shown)


spec/integration/step_down_spec.rb
```diff
@@ -2,8 +2,8 @@
 
 describe 'Step down behavior' do
   require_topology :replica_set
-  min_server_fcv '4.2'
 
+=begin Uncomment to shorten the runtime of this test
```
saghm: If this shortens the runtime of the test, why do we want it disabled by default?

spec/support/event_subscriber.rb
```diff
@@ -53,13 +53,26 @@ def failed(event)
       end
     end
 
+    def select_cmap_events(cls)
```
saghm: This doesn't seem to have anything specific to CMAP in it; maybe it should jut be named `select_succeeded_events`?

spec/integration/step_down_spec.rb
p-mongo: I would prefer to have this approach used by default, but the spec as approved describes the simpler approach which takes longer.

saghm: Ah, okay, I didn't realize that. Can you update this comment to mention that?

spec/support/event_subscriber.rb
p-mongo: The purpose of this method is to select cmap events. As CMAP implementation is incomplete, the implementation of this method is subject to change as well.

spec/integration/step_down_spec.rb
p-mongo: What would you like this comment to say?

saghm: Something about how this is only disabled by default per spec requirements; maybe something like:

```ruby
=begin Uncommenting this will reduce the runtime of the test, but the spec currently does not allow us to do it
```

spec/support/event_subscriber.rb
saghm: The current name of the method does not reflect what it does very well, given that it works fine for any event class, not just CMAP ones. If future changes change the semantics of this method, then the name can be changed then to reflect it. This is code only used internally in the tests, so there's no issue with changing the name of it later if it needs to be changed to become CMAP-specific later (which doesn't seem like it would definitely have to be the case).

spec/integration/step_down_spec.rb
p-mongo: Made the suggested change.

spec/support/event_subscriber.rb
p-mongo: This method exists to select cmap events, and that is what it does. The fact that it may be used for another purpose does not necessarily make it appropriate for other purposes. If in the future this method ends up being used for something else, its name and/or its implementation may be adjusted. As you pointed out, this is easy to do.

saghm: I still think it's pretty silly to arbitrarily claim that the method only works for CMAP events when it very clearly doesn't, but since this isn't a public API, it's not worth blocking these changes on, so I'll resolve this conversation.

spec/integration/step_down_spec.rb
saghm: Thanks!

PR #1356, opened 2019-05-10
RUBY-1807 Client side errors should not change transaction state


p-mongo:


spec/support/transactions/test.rb
```diff
@@ -63,14 +63,22 @@ def initialize(data, test, spec)
...
-          result = o['result']
-          next result unless result.class == Hash
-
-          # Change maps of result ids to arrays of ids
-          result.dup.tap do |r|
-            r.each do |k, v|
-              next unless ['insertedIds', 'upsertedIds'].include?(k)
-              r[k] = v.to_a.sort_by(&:first).map(&:last)
+          if o.key?('error')
+            if o['error']
```
saghm: Is the `error` value ever actually initialized to a falsy value? I don't see it happening anywhere, but I might just be missing it

p-mongo: There are no tests specifying `error: false` if this is what you are asking.

saghm: Why do we need to check both `o.key?('error')` and `o['error']` then? It seems redundant

p-mongo: Since currently the runner does not handle `error: false`, if such a thing was to be added in the future the runner would fail rather than succeeding while not performing the expected assertions.

saghm: The test runner still fails in this case though, since the `else` branch raises an error.

p-mongo: That is correct. Given the code as currently implemented in the PR, the test runner fails on unexpected input.

saghm: I guess I'm just wondering why the premature handling of a case that we don't know will ever be added when it still will fail even with the handling

p-mongo: What do you suggest to do instead?

saghm: It doesn't seem necessary to me to check `o.key?` as well as `o['error']`, but I guess just adding a comment explaining why this is done would clear it up enough

p-mongo: I can add a comment, what would you like it to say?

saghm: Maybe something like "We check both `o.key('error')` and `o['error']` to provide a better error message in case `error: false` is ever needed in the tests"?

p-mongo: Added the comment.

PR #1365, opened 2019-05-31
RUBY-1790 Document connection string and URI option precedence rules

HanaPearlman:


docs/tutorials/ruby-driver-create-client.txt
```diff
@@ -70,7 +70,8 @@ Client Options
 
 A number of different options can be passed to a ``Mongo::Client`` to configure driver
 behavior, either by providing them in the options hash to the constructor or by
-providing them in the URI.
+providing them in the URI. If both a URI and an options hash are provided, the options
```
p-mongo: Can you change "the options" to "the options in the hash" here for clarity please?

PR #1368, opened 2019-05-31
RUBY-1823 Support adding multiple CA certificates


p-mongo:


spec/support/spec_config.rb
```diff
@@ -189,7 +189,7 @@ def spec_root
   end
 
   def ssl_certs_dir
-    "#{spec_root}/support/certificates"
+    Pathname.new("#{spec_root}/support/certificates")
```
saghm: Why is `Pathname` needed here but not in each of the functions below where a path is returned?

p-mongo: Pathname use is for convenience only, it is not required. The certificate path code will be refactored/unified at a future time but this is not part of the current PR.

saghm: Okay, seems reasonable

PR #1370, opened 2019-05-31
RUBY-1824 Run "Local TLS" tests in Evergreen


p-mongo:


.evergreen/run-tests.sh
```diff
@@ -18,11 +18,11 @@ set -o errexit  # Exit the script with error if any of the commands fail
 set_fcv
 set_env_vars
 
-export DRIVER_TOOLS_CLIENT_CERT_PEM="${DRIVERS_TOOLS}/.evergreen/x509gen/client-public.pem"
-export DRIVER_TOOLS_CLIENT_KEY_PEM="${DRIVERS_TOOLS}/.evergreen/x509gen/client-private.pem"
-export DRIVER_TOOLS_CLIENT_CERT_KEY_PEM="${DRIVERS_TOOLS}/.evergreen/x509gen/client.pem"
-export DRIVER_TOOLS_CA_PEM="${DRIVERS_TOOLS}/.evergreen/x509gen/ca.pem"
-export DRIVER_TOOLS_CLIENT_KEY_ENCRYPTED_PEM="${DRIVERS_TOOLS}/.evergreen/x509gen/password_protected.pem"
+#export DRIVER_TOOLS_CLIENT_CERT_PEM="${DRIVERS_TOOLS}/.evergreen/x509gen/client-public.pem"
```
saghm: If these aren't necessary anymore, is there a reason not to just delete them instead of commenting them? They'll still be in the git history if for some reason we need to restore them again

p-mongo: Yes, I wouldn't know where in git history to look for these lines. But when they are commented out they are easy to locate.

saghm: It seems like it would be pretty easy to just look at the git history of this specific file, since there aren't likely to be many changes to it, but you seem to feel much more strongly about this than I do, so I won't insist

PR #1375, opened 2019-06-04
RUBY-1811 ChangeStream spec's Resumable Error definition is too broad

HanaPearlman:


lib/mongo/error/operation_failure.rb
```diff
@@ -162,6 +168,16 @@ def change_stream_resumable_code?
       end
       private :change_stream_resumable_code?
 
+      def change_stream_resumable_label?
```
p-mongo: This method would return true when there is no label, which I think is confusing. Can it be changed to `change_stream_not_resumable_label?` which would return true if 1) there is a label and 2) it is a change stream not resumable one?

lib/mongo/error/operation_failure.rb
```diff
@@ -162,6 +168,16 @@ def change_stream_resumable_code?
...
+        if labels
+          !labels.include? CHANGE_STREAM_NOT_RESUME_LABEL
```
p-mongo: Since this is the only use of the constant, the literal can be given inline.

spec/mongo/error/operation_failure_spec.rb
```diff
@@ -148,6 +148,52 @@
...
+      context 'not a getMore response' do
+        let(:error) { Mongo::Error::OperationFailure.new('no message', nil,
+          :code => 91, :code_name => 'ShutdownInProgress',
+          :labels => [ Mongo::Error::TRANSIENT_TRANSACTION_ERROR_LABEL ]) }
+
+        it 'returns false' do
+          expect(error.change_stream_resumable?).to eql(false)
+        end
+      end
+    end
```
p-mongo: Can a test also be added which tests what happens when there is no label?

spec/support/change_streams.rb
```diff
@@ -165,8 +177,12 @@ def match_result?(result)
         end
 
         def match_commands?(actual)
-          @expectations.each_with_index.all? do |e, i|
-            actual[i] && match?(e, actual[i])
+          if @expectations
```
p-mongo: I would prefer this condition be given in the spec test rather than in the matcher.

Example 1: https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/spec_tests/uri_options_spec.rb#L54

Example 2: https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/support/crud.rb#L153 (the crud tests have a bit more than usual meta programming around them)

spec/mongo/error/operation_failure_spec.rb
HanaPearlman: The other tests for `change_stream_resumable?` in this file do not have labels. I didn't want to be repetitive, but is it better to explicitly include similar tests?

p-mongo: My mistake, everything here is already correct.

p-mongo:
Unfortunately we now have a bunch of other test failures.

PR #1385, opened 2019-06-12
RUBY-1561 Changes to retryable reads/writes to support sharded transactions

p-mongo:


lib/mongo/retryable.rb
```diff
@@ -222,6 +222,41 @@ def write_with_retry(session, write_concern, ending_transaction = false, &block)
...
+    # This method is used for operations which are not supported by modern
+    # retryable writes, such as delete_many and update_many.
+    #
+    # @param [ Server ] server The server which should be used for the
+    #   operation. If not provided, the current primary will be retrieved from
+    #   the cluster.
+    # @param [ nil | Session ] session Optional session to use with the operation.
+    #
+    # @api private
+    def legacy_write_with_retry(server = nil, session = nil)
```
saghm: This method is just moved from down below without the implementation being changed, right?

p-mongo: That is correct.

saghm: Cool, just making sure

PR #1387, opened 2019-06-13
RUBY-1796 Test writes in transactions do not send writeConcern when collection write concern is w=0

HanaPearlman:


lib/mongo/bulk_write.rb
```diff
@@ -154,11 +155,39 @@ def single_statement?(operation)
...
+      if options[:write_concern]
+        WriteConcern.get(options[:write_concern])
+      else 
+        applied_collection_write_concern(session)
+      end
+    end
+
+    def applied_collection_write_concern(session)
+      wc = collection.write_concern
+      if session.respond_to?(:in_transaction?) && session.in_transaction?
```
p-mongo: The pattern we use when session may not be provided is `if session && session.in_transaction?`

lib/mongo/bulk_write.rb
```diff
@@ -154,11 +155,39 @@ def single_statement?(operation)
...
+        if wc && WriteConcern.send(:unacknowledged?, wc.options)
+          opts = wc.options.dup
+          opts.delete(:w)
```
p-mongo: I believe this should only be done if `w == 0`.

HanaPearlman: The `unacknowledged?` method just checks for `w == 0`, but if you would prefer for that to be explicit here, I can do that.

p-mongo: You are right, no changes are needed.

PR #1392, opened 2019-06-14
RUBY-1561 Move error labels into operations layer 

p-mongo:


lib/mongo/error.rb
```diff
@@ -85,17 +85,19 @@ def change_stream_resumable?
     # manually retried by the user.
     #
     # @since 2.6.0
+    # @deprecated
```
saghm: Why don't we want constants for these values? Users will need to be able to specify them in their code to handle errors properly, so I feel like having constants is better UX.

p-mongo: Driver specifications reference error labels by their names. Server output also includes label names. I found while working on this PR that translating from label names to driver-specific constants created friction and the risk of missing related code/data that used label names, such as spec tests. Users who wish to use constants instead of string literals retain the ability to define their own constants.

saghm: I think that putting a little more work into the implementation is worth it to make it easier for users to properly use the API. We provide constants in a similar way for the monitoring API, and I think the value proposition holds here as well. It seems surprising to me that the need to remember to use constants could be burdensome enough to be worth avoiding at the cost of giving users a more friendly API.

p-mongo: The easiest way to use the Ruby driver for users is to use the label names they already know, which are the ones referenced in the server error messages, our documentation examples, and driver specifications. All caps constants that only exist in the Ruby driver require more effort to use.

p-mongo:
@saghm  I understand that you have a preference for constants but I do not find them to be the best approach in this particular case. Do you have any other comments on this PR?

PR #1401, opened 2019-06-20
RUBY-1636 Resync change stream tests to test all new notification types

HanaPearlman:
Includes RUBY-1416 Update ChangeStream tests for 4.1.1 "drop" notifications

### DIFF
--- a/spec/support/change_streams/operation.rb
+++ b/spec/support/change_streams/operation.rb
@@ -50,12 +50,39 @@ module Mongo
         coll.insert_one(document)
       end
 
+      def update_one(coll)
+        coll.update_one(filter, arguments['update'])
+      end
+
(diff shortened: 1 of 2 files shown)


spec/support/change_streams/operation.rb
```diff
@@ -50,13 +50,40 @@ def insert_one(coll)
...
+      def replace_one(coll)
+        coll.replace_one(filter, arguments['replacement'])
+      end
+
+      def delete_one(coll)
+        coll.delete_one(filter)
+      end
+
+      def drop(coll)
+        coll.drop()
```
p-mongo: Parentheses are usually elided when a method call has no arguments, i.e. this would normally be spelled `coll.drop`.

PR #1406, opened 2019-06-26
RUBY-1383 RUBY-1854 RUBY-1834 test & fix read preference handling by the driver in various scenarios

p-mongo:


saghm:
Does this need to be rebased with master? It looks like it contains all the changes from the `write` -> `write_concern` PR

lib/mongo/collection.rb
```diff
@@ -228,6 +228,13 @@ def capped?
     #
     # @since 2.0.0
     def create(opts = {})
+      # Passing read options to create command causes it to break.
+      # Filter the read options out.
+      # TODO put the list of read options in a class-level constant when
```
HanaPearlman: Do we have this list?

lib/mongo/session.rb
```diff
@@ -770,7 +779,7 @@ def add_txn_num!(command)
     def add_txn_opts!(command, read)
       command.tap do |c|
         # The read preference should be added for all read operations.
-        if read && txn_read_pref = txn_read_preference
+        if false && read && txn_read_pref = txn_read_preference
```
HanaPearlman: This statement should never evaluate to true, correct?

saghm: Why is this block being disabled? if we don't want it to run, can we just remove it?

lib/mongo/collection.rb
p-mongo: No, this will need to be researched or done organically over time.

lib/mongo/session.rb
p-mongo: Right,  I forgot to remove this piece.

PR #1409, opened 2019-06-26
RUBY-1858 Prohibit :journal option in write concerns


p-mongo:


lib/mongo/write_concern/base.rb
```diff
@@ -43,8 +43,24 @@ class Base
...
+        options = Options::Mapper.transform_values_to_strings(options).freeze
+
+        if options[:w]
+          if options[:w] == 0 && options[:j]
+            raise Error::InvalidWriteConcern, "Invalid write concern options: :j cannot be true when :w is 0: #{options.inspect}"
+          elsif options[:w] == 0 && options[:fsync]
+            raise Error::InvalidWriteConcern, "Invalid write concern options: :fsync cannot be true when :w is 0: #{options.inspect}"
+          elsif options[:w].is_a?(Integer) && options[:w] < 0
+            raise Error::InvalidWriteConcern, "Invalid write concern options: :w cannot be negative (#{options[:w]}): #{options.inspect}"
+          end
```
saghm: Do we also want to validate the types of the fields that are present (e.g. that `w` is either an integer, string, or symbol, j is a boolean, etc.)?

p-mongo: I have not run into this issue to date.

saghm: The idea would be for this to benefit users, who wouldn't otherwise have a silent failure. If you don't think that's helpful, we can leave it as-is

p-mongo:
The server errors on `journal` option:

<pre>
irb(main):002:0> Mongoid.default_client['foo'].with(write:{w:9,journal:true}).insert_one(foo:1)

Mongo::Error::OperationFailure (unrecognized write concern field: journal (9))
</pre>

However, I think there is value in detecting this condition earlier in the driver as this PR proposes.

PR #1410, opened 2019-06-26
1605 bg connect wip

HanaPearlman:
This PR covers part of [RUBY-1605](https://jira.mongodb.org/browse/RUBY-1605).
Summary:

-  Sets up a background thread attached to connection pool which keeps the pool at min_size. The thread is triggered whenever the pool size decreases. This does not connect (handshake and auth) the connections; that will happen in a later PR.

- During SDAM flow, when a server is identified as data-bearing, the driver starts up the background thread for the server, bringing its connection pool up to min_size.


lib/mongo/server/connection_pool.rb
```diff
@@ -493,14 +507,44 @@ def close_idle_sockets
         end
       end
 
-      # Creates up to the min size connections.
+      # Create and add connections to the pool until the
+      # pool size is at least min_size.
       #
-      # Used by the spec test runner.
+      # Used by the spec test runner and the pool populator background thread.
```
p-mongo: The spec runner used this method because we didn't have the background thread. Now that the background thread exists, the spec runner should not be calling this method at all and should instead use the pool normally.

lib/mongo/server/connection_pool/connection_pool_populator.rb
```diff
@@ -0,0 +1,48 @@
...
+      def start!
+        @thread = Thread.new {
+          while !@pool.closed? do
+            @pool.populate
+            @pool.populate_semaphore.wait(nil)
+          end
+        }
+      end
+
+      def stop!
```
p-mongo: The stop method should take a `wait=false` argument like the server monitor does. https://github.com/mongodb/mongo-ruby-driver/pull/1415 expands the scope of what is waited for,  I suggest waiting until that PR is merged prior to making this change.

p-mongo: Which also means that, perhaps, `Pool#close` method should take a `wait: false` argument as well.

When clients are closed in test suite, the caller wants to wait for everything that is part of the client to be torn down and stopped which includes all background processing.

lib/mongo/server/connection_pool/connection_pool_populator.rb
```diff
@@ -0,0 +1,48 @@
...
+        # Kill the thread instead of signaling so that if stop! is called during
+        # populate or before the wait() on the semaphore, the thread still terminates
+        @thread.kill if @thread
+      end
+
+      def is_running?
```
p-mongo: This can be called simply `running?`.

lib/mongo/server/connection_pool/connection_pool_populator.rb
HanaPearlman: Yes, I agree with this, since when pools are closed in the test suite we will want stop the background thread, etc. However, this won't help us when a client is closed, since the `Pool#close` method is not called (at least for now).

PR #1411, opened 2019-06-27
Memoize Mongo::Operation::PolymorphicResult#result_class to speedup driver

cesarizu:
This change alone speeds up our test suit by 10%. It seems that the `PolymorphicResult` was doing a symbol lookup unnecessarily too often.


cesarizu:
Closing in favor of #1412

PR #1412, opened 2019-06-27
Memoize Mongo::Operation::PolymorphicResult#result_class to speedup driver


p-mongo:


p-mongo:
@cesarizu Thank you for the patch.

PR #1415, opened 2019-06-29
RUBY-1868 Assorted fixes to improve test suite reliability

p-mongo:


p-mongo:
This hasn't gotten Evergreen to green quite yet, unfortunately.

spec/mongo/cursor_spec.rb
```diff
@@ -273,7 +279,7 @@
     context 'when the cursor is not fully iterated and is garbage collected' do
 
       let(:documents) do
-        (1..4).map{ |i| { field: "test#{i}" }}
+        (1..6).map{ |i| { field: "test#{i}" }}
```
saghm: It's not clear at all to me why we made 3 documents before or why having 5 now is better. Why does this need to be changed?

lib/mongo/cluster.rb
```diff
@@ -150,12 +160,14 @@ def initialize(seeds, monitoring, options = Options::Redacted.new)
...
 
-      @cursor_reaper = CursorReaper.new
-      @socket_reaper = SocketReaper.new(self)
-      @periodic_executor = PeriodicExecutor.new(@cursor_reaper, @socket_reaper)
-      @periodic_executor.run!
+      if options[:cleanup] != false
+        @cursor_reaper = CursorReaper.new
+        @socket_reaper = SocketReaper.new(self)
+        @periodic_executor = PeriodicExecutor.new(@cursor_reaper, @socket_reaper)
+        @periodic_executor.run!
```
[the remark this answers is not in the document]
p-mongo: Yes it should I think, but let's make that change in a separate PR.

spec/mongo/cursor_spec.rb
p-mongo: 6 total documents mean the collection requires two getmores to iterate after the first getmore is done, which is needed for jruby as described in the comment in the next hunk.

saghm: Ah, okay. I hadn't realized that this change was also related to that comment.

PR #1420, opened 2019-07-02
RUBY-1868 Further test suite reliability improvements

p-mongo:


spec/mongo/cursor_spec.rb
```diff
@@ -319,25 +319,26 @@
           # We need to verify that the cursor was able to retrieve more documents
           # from the server so that more than one batch is successfully received
 
+          cluster.unregister_cursor(cursor.id)
```
HanaPearlman: Why do we need to do this?

p-mongo: This is what the test does - it unregisters the cursor and verifies a behavior.

I see now that this diff appears to be different from what it actually is. What happened was I moved the comment between the unregister call and the assertions, because comments apply to the assertions.

PR #1427, opened 2019-07-08
RUBY-1833 Implement missing Change Stream prose tests

HanaPearlman:


spec/integration/change_stream_spec.rb
```diff
@@ -45,6 +45,16 @@ def clear_fail_point_before
       end
     end
 
+    shared_context 'next errors with a non-resumable error' do
```
p-mongo: This should be `shared_examples_for`.

spec/integration/change_stream_spec.rb
```diff
@@ -45,6 +45,16 @@ def clear_fail_point_before
...
+      it 'errors with a non-resumable error' do
```
p-mongo: This can be something like `raises an exception`.

spec/mongo/collection/view/change_stream_spec.rb
```diff
@@ -591,6 +592,32 @@
...
+    let(:enum) do
+      change_stream.to_enum
+    end
+
+    it 'should create a new cursor and resume' do
+      document = enum.next
+
+      expect(document[:fullDocument][:a]).to eq(2)
+
+      # 'cursor' still references the cursor that was killed
```
p-mongo: This comment seems to be opposite the assertion on the next line.

PR #1433, opened 2019-07-10
Fix RUBY-1840 Legacy write retries are used when modern retryable writes are requested on some operations


p-mongo:


lib/mongo/bulk_write.rb
```diff
@@ -39,7 +39,7 @@ class BulkWrite
                    :database,
                    :cluster,
                    :write_with_retry,
-                   :legacy_write_with_retry,
+                   :nro_write_with_retry,
```
saghm: Out of curiosity, what does "nro" stand for?

p-mongo: "Non-retryable operation".

PR #1435, opened 2019-07-10
RUBY-1605 WIP

HanaPearlman:


lib/mongo/server/connection_pool.rb
```diff
@@ -523,30 +541,70 @@ def populate
...
+                  @available_connections << connection
+                  @pending_connections.delete(connection)
+
+                  # wake up one thread waiting for connections, since one was created
+                  @available_semaphore.signal
+                end
+              end
+            rescue Exception => e
+              @lock.synchronize do
+                @pending_connections.delete(connection)
```
p-mongo: Closing the connection here is probably a good idea, in case the exception happened during handshake/auth after the socket got connected.

PR #1438, opened 2019-07-13
RUBY-1655 Implement CRUD v2 spec tests

p-mongo:
https://jira.mongodb.com/browse/RUBY-1655


p-mongo:
Looks like outcome isn't checked for v2 tests, I'll repair that in a follow-up PR.

PR #1439, opened 2019-07-13
RUBY-1878 RUBY-1711 Do not send multi: false & upsert: false to the server

p-mongo:
https://jira.mongodb.com/browse/RUBY-1878

### DIFF
--- a/lib/mongo/collection/view/writable.rb
+++ b/lib/mongo/collection/view/writable.rb
@@ -230,9 +230,10 @@ module Mongo
         def replace_one(replacement, opts = {})
           update_doc = { Operation::Q => filter,
                          Operation::U => replacement,
-                         Operation::MULTI => false,
-                         Operation::UPSERT => !!opts[:upsert]
                         }
+          if opts[:upsert]
+            update_doc[:upsert] = true
+          end
           with_session(opts) do |session|
             write_concern = write_concern_with_session(session)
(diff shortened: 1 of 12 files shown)


lib/mongo/collection/view/writable.rb
```diff
@@ -317,8 +321,10 @@ def update_many(spec, opts = {})
         def update_one(spec, opts = {})
           update_doc = { Operation::Q => filter,
                          Operation::U => spec,
-                         Operation::MULTI => false,
-                         Operation::UPSERT => !!opts[:upsert] }
+                         }
+          if opts[:upsert]
+            update_doc[:upsert] = true
```
joeyAghion: I noticed upon upgrading to 2.11.2 that our software's uses of `upsert: true` failed to insert new documents, and traced it to this change. Replacing the `:upsert` symbol with `Operation::UPSERT` (which is a _string_) fixed our problem. We've downgraded to v2.9.x in the meantime.

p-mongo: @joeyAghion Can you please create a new jira ticket (https://jira.mongodb.org/browse/RUBY) describing your issue, including the following:

- a code snippet reproducing the problem
- what you expected to happen
- what actually happened

joeyAghion: https://jira.mongodb.org/browse/RUBY-2054

PR #1441, opened 2019-07-13
RUBY-1563 Support polling SRV records for mongos discovery

p-mongo:
https://jira.mongodb.com/browse/RUBY-1563


spec/integration/reconnect_spec.rb
```diff
@@ -3,7 +3,7 @@
 describe 'Client after reconnect' do
   let(:client) { authorized_client }
 
-  it 'works' do
```
egiurleo: 😂

PR #1442, opened 2019-07-15
RUBY-1605 Establish working connections in a background thread

HanaPearlman:
This also includes the change required by RUBY-1863 Ignore ConnectionReadyEvent in CMAP pool creation test


lib/mongo/server/connection_pool.rb
```diff
@@ -513,40 +530,102 @@ def close_idle_sockets
...
+
+            begin
+              connect_connection(connection)
+            rescue Exception => e
+              @lock.synchronize do
+                @pending_connections.delete(connection)
+              end
+
+              if retried
+                return false
```
p-mongo: This path does not signal the available semaphore. Do I have the following scenario correct:

- Pool is at max_size-1
- Populator starts to create another connection
- A thread wants a connection, can't create it in flow because pool is now at max_size, waits for the semaphore
- Populator fails with an error on the connection, but does not signal the semaphore
- Thread continues waiting although it should either abort with an error (current behavior we are going for as I understand it) or attempt to create the connection in flow (possible future behavior)
- Eventually the thread times out and raises the timeout exception

spec/mongo/server_spec.rb
```diff
@@ -177,6 +177,14 @@
         expect(server.monitor.instance_variable_get('@thread')).to be nil
       end
     end
+  end
+
+  describe '#scan!' do
+    include_context 'with monitoring io'
```
p-mongo: This context seems redundant now.

spec/mongo/server_spec.rb
```diff
@@ -177,6 +177,14 @@
...
+
+    let(:server_options) do
```
p-mongo: Would the test still pass if this is  moved into the `with monitoring io' context below?

lib/mongo/server/connection_pool.rb
HanaPearlman: Yes, this is a good point. If the semaphore is signaled after the retry fails, the thread will wake up and attempt to create the connection in flow, and error if that fails. I think this is preferred.

lib/mongo/server/connection_pool.rb
```diff
@@ -513,40 +530,102 @@ def close_idle_sockets
...
+
+        @lock.synchronize do
+          # If stop_populator is called while populate is running, there may be
+          # connections waiting to be connected, connections which have not yet
+          # been moved to available_connections, or connections moved to available_connections
+          # but not deleted from pending_connections. These should be cleaned up.
+          until @pending_connections.empty?
+            connection = @pending_connections.take(1).first
+            connection.disconnect!
+            @pending_connections.delete(connection)
```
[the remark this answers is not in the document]
HanaPearlman: It could not be creating new connections while this loop is executing, since this loop is locked and so is connection creation (within the populate method). However, it could be trying to connect the connection or waiting for the lock to be released.

lib/mongo/server/connection_pool.rb
```diff
@@ -421,6 +437,8 @@ def close(options = nil)
 
         options ||= {}
 
+        stop_populator(options[:wait])
```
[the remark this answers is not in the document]
HanaPearlman: I went with that strategy (always waiting for the populator) for now, and in the future intend to improve this.

lib/mongo/server/connection_pool.rb
```diff
@@ -513,40 +535,83 @@ def close_idle_sockets
...
+        begin
+          return create_and_add_connection
+        rescue Exception
+          # an error was encountered while connecting the connection,
+          # ignore this first error and try again.
+        end
+
+        begin
+          return create_and_add_connection
+        rescue Exception
```
p-mongo: ... but the semaphore signal should happen on all exceptions, regardless of whether they are retried. This is probably easiest to achieve by moving this rescue block to the top level of the method.

lib/mongo/server/connection_pool/connection_pool_populator.rb
```diff
@@ -34,8 +34,15 @@ def pre_stop
...
       def do_work
         throw(:done) if @pool.closed?
-        unless @pool.populate
-          @pool.populate_semaphore.wait
+
+        begin
+          unless @pool.populate
+            @pool.populate_semaphore.wait
+          end
+        rescue Exception
```
p-mongo: Let's change this to `rescue Error` to rescue all Mongo::Error derivatives, and `log_warn` the encountered exception prior to sleeping.

spec/mongo/server/connection_pool_spec.rb
```diff
@@ -686,6 +663,7 @@ def create_pool(min_pool_size)
       it 'does not add the connection to the pool' do
         pending 'Re-enable when connections are connected prior to being returned from check_out method'
 
+        # fails because with_connection raises the SocketError which is not caught anywhere
```
p-mongo: Note to self to look at this.

lib/mongo/server/connection_pool/connection_pool_populator.rb
```diff
@@ -39,9 +43,10 @@ def do_work
           unless @pool.populate
             @pool.populate_semaphore.wait
           end
-        rescue Exception
+        rescue Error => e
           # Errors encountered when trying to add connections to
           # pool; try again later
+          log_warn("Populator failed to connect a connection due to #{e.message}.")
```
p-mongo: The idiom to output full exception diagnostics is this:

`"Populator failed to connect a connection: #{e.class}: #{e.message}"`

PR #1443, opened 2019-07-16
1879 do_connect closes socket if an error occurs during any connect step

HanaPearlman:


lib/mongo/server/connection.rb
```diff
@@ -173,11 +173,21 @@ def connect!
...
-        socket = address.socket(socket_timeout, ssl_options,
-          connect_timeout: address.connect_timeout)
-        handshake!(socket)
-        pending_connection = PendingConnection.new(socket, @server, monitoring, options)
-        authenticate!(pending_connection)
+        socket = nil
+
+        begin
+          socket = address.socket(socket_timeout, ssl_options,
+            connect_timeout: address.connect_timeout)
```
p-mongo: If you move `begin` below this line, the rescue block won't need to check whether `socket` is not nil.

lib/mongo/server/connection.rb
```diff
@@ -173,11 +173,21 @@ def connect!
...
+          handshake!(socket)
+          pending_connection = PendingConnection.new(socket, @server, monitoring, options)
+          authenticate!(pending_connection)
+        rescue Exception => e
+          if socket
+            socket.close
+          end
+          raise e
```
p-mongo: Ruby allows an argument-less `raise` which will reraise the exception being handled, which also would make it unnecessary to assign it to a variable (`=> e` part).

lib/mongo/server/connection.rb
HanaPearlman: I thought about this, but it looks like the call on this line could also throw an exception, which is why I included it in the block.

p-mongo: Yes but if it does, the cleanup code in this method won't do anything since `socket` won't be assigned.

PR #1444, opened 2019-07-16
updates mongodb compatibility matrix

terakilobyte:
This is a small addition that updates the compatibility matrix for 2.10


p-mongo:
Thank you for the patch.

PR #1448, opened 2019-07-17
RUBY-1871 Associate connections with pools

HanaPearlman:
First part of 1871 followup work: Distinguish between connection being checked into the correct pool, but out of sequence vs connection being checked into a foreign pool

### DIFF
--- a/lib/mongo/server/connection_pool.rb
+++ b/lib/mongo/server/connection_pool.rb
@@ -348,8 +348,8 @@ module Mongo
       # @since 2.9.0
       def check_in(connection)
         @lock.synchronize do
-          unless connection.pool_object_id == object_id
-            raise ArgumentError, "Trying to check in a connection which was not checked out by this pool: #{connection}"
+          unless connection.pool == self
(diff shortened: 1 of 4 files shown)


lib/mongo/server/connection_pool.rb
```diff
@@ -326,20 +326,24 @@ def check_out
...
           # This means threads waiting for a connection to free up when
           # the pool is at max size may time out.
           # Threads that begin waiting after this method completes (with
           # the exception) should be fine.
-          publish_cmap_event(
-            Monitoring::Event::Cmap::ConnectionCheckedIn.new(@server.address, connection.id)
-          )
+
+          if @checked_out_connections.include?(connection)
+            # It is possible the connection was already checked in; we only
```
p-mongo: What is the use case for checking in a connection twice? I am inclined to think this should be an error.

HanaPearlman: This is not a strong case for it, but it would just be more tolerant of connections that are checked out without `with_connection`, which could accidentally be checked in twice if the connection reference is not handled carefully. It makes more sense to error, so I will change it.

PR #1456, opened 2019-07-19
RUBY-1884 time out background thread waits after about 10 seconds


p-mongo:


lib/mongo/server/connection_pool/connection_pool_populator.rb
```diff
@@ -24,12 +24,19 @@ class ConnectionPoolPopulator
...
       attr_reader :options
 
-      def initialize(pool)
+      # @param [ Server::ConnectionPool ] The connection pool.
+      # @param [ Hash ] options The options.
+      #
+      # @option options [ Logger ] :logger A custom logger to use.
+      def initialize(pool, options = {})
         @pool = pool
         @options = @pool.options
```
HanaPearlman: This line can be deleted

spec/mongo/server/connection_pool/populator_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  let(:server) do
+    authorized_client.cluster.next_primary
+  end
+
+  let(:pool) do
+    server.pool
+  end
+
+  let(:populator) do
+    described_class.new(pool, pool.options)
```
HanaPearlman: Should we be doing some kind of clean up for the populator after the test is run?

lib/mongo/server/connection_pool/connection_pool_populator.rb
p-mongo: I seem to have missed this, what options did the populator use before?

spec/mongo/server/connection_pool/populator_spec.rb
p-mongo: This is probably a good idea.

lib/mongo/server/connection_pool/connection_pool_populator.rb
HanaPearlman: The populator includes Loggable, so that’s what the options were used for before. Now that the Background Thread includes Loggable, I think the “includes Loggable” line also can be deleted.

p-mongo: Okay removed that too.

p-mongo:
Rebased to fix the test suite.

PR #1463, opened 2019-07-24
RUBY-1887 add AuthError and rescue it with Mongo::Error (wip)

HanaPearlman:
This does not include an entire audit of the driver; that will be in a future PR.

### DIFF
--- a/lib/mongo/error/auth_error.rb
+++ b/lib/mongo/error/auth_error.rb
@@ -17,7 +17,11 @@ module Mongo
 
     # Raised when authentication fails.
     #
-    # @since 2.7.0
+    # Note: This class is derived from RuntimeError for
+    # backwards compatibility reasons. It is subject to
+    # change in future major versions of the driver.
+    #
+    # @since 2.11.0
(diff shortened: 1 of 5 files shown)


lib/mongo/error/auth_error.rb
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module Mongo
+  class Error
+
+    # Raised when authentication fails.
+    #
+    # @since 2.7.0
```
p-mongo: This should be `@since 2.11.0`.

lib/mongo/error/auth_error.rb
```diff
@@ -0,0 +1,24 @@
...
+    class AuthError < RuntimeError
```
p-mongo: Can you please add a note that this class is intentionally derived from `RuntimeError` for backwards compatibility reasons, and this is subject to change in a future major version of the driver?

PR #1464, opened 2019-07-24
RUBY-1605 Rename Connection Pool Populator

HanaPearlman:


spec/integration/connection_pool_populator_spec.rb
```diff
@@ -266,12 +266,14 @@
     context 'when min size is provided' do
       min_server_version '2.8'
 
+      let!(:client) do
+        register_client(ClientRegistry.instance.new_local_client([SpecConfig.instance.addresses.first],
```
p-mongo: `new_local_client` already registers the returned clients for cleanup, additional registering is not needed.

PR #1481, opened 2019-08-07
RUBY-1896 Raise an actionable error message when retryWrites fails due to using an unsupported storage engine	

HanaPearlman:


lib/mongo/retryable.rb
```diff
@@ -403,5 +407,10 @@ def log_retry(e, options = nil)
       end
       Logger.logger.warn "#{message} due to: #{e.class.name} #{e.message}"
     end
+
+    # Retry writes on MMAPv1 should raise an actionable error, so we change the error message.
+    def raise_unsupported_error(e)
+      raise e, "This MongoDB deployment does not support retryable writes. Please add retryWrites=false to your connection string."
```
p-mongo: The original exception class and message need to be included in the final message following the `#{e.class}: #{e}` pattern.

lib/mongo/retryable.rb
```diff
@@ -215,9 +215,13 @@ def write_with_retry(session, write_concern, ending_transaction = false, &block)
         end
         retry_write(e, session, txn_num, &block)
       rescue Error::OperationFailure => e
-        if (session.in_transaction? && !ending_transaction) || !e.write_retryable?
+        if e.code == 20 && e.message.start_with?("Transaction numbers")
```
p-mongo: Given that OperationFailure already contains a bunch of logic for figuring out whether a specific error occurred, I think this logic should be moved to OperationFailure in a similar fashion to https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/error/operation_failure.rb#L252.

lib/mongo/retryable.rb
p-mongo: Since the class of exception is not changing, it may also make sense to preserve the backtrace which looks like this:
<pre>
new_e = Error::OperationFailure.new(new_msg)
new_e.set_backtrace(e.backtrace)
raise new_e
</pre>

spec/mongo/retryable_spec.rb
```diff
@@ -591,5 +591,18 @@ class RetryableHost
...
+      before do
+        expect(operation).to receive(:execute).and_raise(
+          Mongo::Error::OperationFailure.new('Transaction numbers are only allowed on...',
+            nil, :code=>20)).ordered
+      end
+
+      it 'raises an exception with the correct error message' do
+        expect {
+          retryable.write
+        }.to raise_error(Mongo::Error::OperationFailure, 'This MongoDB deployment does not support retryable writes. Please add retryWrites=false to your connection string.')
```
p-mongo: I think the Ruby option should also be mentioned, e.g. `Please add retryWrites=false to your connection string or retry_writes: false Ruby client option`. And we do not include trailing periods in exception messages for composability. (I am aware of the spec language mandating this exact message.)

p-mongo:
If a client has retry_reads: true set on an mmapv1 deployment, do the reads work?

HanaPearlman:
From what I can see yes, the reads succeed.

lib/mongo/retryable.rb
HanaPearlman: Is there another place is the driver where we compose multiple error messages like above? I would have assumed that the helpful error message here ("This MongoDB deployment does not support retryable writes...") should come first, but your comment about not including a trailing period is making me think it should come after the original exception class and message.

p-mongo: The pattern I usually use is when translating one exception class to another, the message of the new class goes first, but when adding information to the existing exception and/or message, the information gets added at the end.

Example exceptions which get extended error messages are:

- NoServerAvailable (https://jira.mongodb.org/browse/RUBY-1787)
- SocketError/SocketTimeoutError (https://github.com/mongodb/mongo-ruby-driver/pull/988/files, https://jira.mongodb.org/browse/RUBY-1391)
- Mongo::Auth::Unauthorized

HanaPearlman: Thank you!

PR #1485, opened 2019-08-09
RUBY-1680 sync connection string spec tests (including SPEC-1349) other than the ones for RUBY-1351


p-mongo:


egiurleo:
Do we ignore jruby failures in evergreen?

p-mongo:
For now, yes, for jruby 9.1 failures.

PR #1486, opened 2019-08-09
RUBY-1868 actually close local clients after each test

p-mongo:
https://jira.mongodb.com/browse/RUBY-1868

### DIFF
--- a/spec/mongo/operation/update/op_msg_spec.rb
+++ b/spec/mongo/operation/update/op_msg_spec.rb
@@ -160,14 +160,19 @@ describe Mongo::Operation::Update::OpMsg do
           before do
             session.instance_variable_set(:@options, { implicit: true })
             # Topology is standalone, hence there is exactly one server
-            authorized_client.cluster.servers.first.monitor.stop!
-            allow(authorized_primary.features).to receive(:sessions_enabled?).and_return(false)
+            authorized_primary.monitor.stop!
           end
 
           it 'creates the correct OP_MSG message' do
-            authorized_client.command(ping:1)
-            expect(Mongo::Protocol::Msg).to receive(:new).with([], {}, expected_global_args, expected_payload_1)
-            op.send(:message, authorized_primary)
+            RSpec::Mocks.with_temporary_scope do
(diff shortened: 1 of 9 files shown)


p-mongo:
It appears that the test suite was not actually closing local clients after every example, which caused background threads to stick around for longer than they should have.

spec/mongo/operation/update/op_msg_spec.rb
```diff
@@ -160,14 +160,19 @@
...
-            authorized_client.cluster.servers.first.monitor.stop!
-            allow(authorized_primary.features).to receive(:sessions_enabled?).and_return(false)
+            authorized_primary.monitor.stop!
           end
 
           it 'creates the correct OP_MSG message' do
-            authorized_client.command(ping:1)
-            expect(Mongo::Protocol::Msg).to receive(:new).with([], {}, expected_global_args, expected_payload_1)
-            op.send(:message, authorized_primary)
+            RSpec::Mocks.with_temporary_scope do
```
egiurleo: Could you explain why you use the `Rspec::Mocks.with_temporary_scope` here? My assumption would be that any doubles or mocks set up in this example would be cleared at the end of the example, but it's possible there's something else going on that I don't understand.

p-mongo: The purpose of the scope is to limit assertions to the operations being tested only, not to create the first scope.

`after` callbacks close local clients (this was meant to be the case prior to this PR also, but I omitted the call to do so erroneously). When clients are closed they can perform network operations, for example `endSessions` calls to clear session pools. RSpec's assertion scope by default covers all `before`/`after` callbacks, therefore assertions defined in this example fail when an `after` callback runs `endSessions` and instantiates additional `Protocol::Msg` objects.

egiurleo: Okay, I see. So the purpose of the `Rspec::Mocks.with_temporary_scope` is to create a scope that does not include `before/after`, correct?

p-mongo: That's right.

PR #1496, opened 2019-08-16
[wip] [RUBY-1745] Invalid auth mechanism is not reported until an operation is attempted on the client

egiurleo:


lib/mongo/auth.rb
```diff
@@ -70,6 +70,26 @@ def get(user)
       SOURCES[mechanism].new(user)
     end
 
+    # Raised when trying to authorize with an invalid configuration
+    #
+    # @since 2.11.0
```
egiurleo: I'm assuming this is going to be released with 2.11.0, but let me know if that's wrong.

lib/mongo/client.rb
```diff
@@ -855,14 +857,32 @@ def validate_new_options!(opts = Options::Redacted.new)
...
+      case options[:auth_mech]
+      when :mongodb_cr, :plain, :scram, :scram256
+        raise Mongo::Auth::InvalidConfiguration.new(options[:auth_mech], 'user is required') if options[:user].nil?
+      when :mongodb_x509
+        raise Mongo::Auth::InvalidConfiguration.new(:mongodb_x509, 'auth_source must be $external or nil') if !['$external', nil].include?(options[:auth_source])
+        raise Mongo::Auth::InvalidConfiguration.new(:mongodb_x509, 'password is not supported') if !options[:password].nil?
+      when nil
+        raise Mongo::Auth::InvalidConfiguration.new(:default, 'auth_source cannot be present without a user') if !options[:auth_source].nil? && options[:user].nil?
+        raise Mongo::Auth::InvalidConfiguration.new(:default, 'do not include user information delimiter (@) in URI for blank user') if !options[:user].nil? && !options[:user].length
+        raise Mongo::Auth::InvalidConfiguration.new(:default, 'do not include password delimiter (:) in URI for blank password') if !options[:password].nil? && !options[:password].length
```
egiurleo: These last two cases are related to parsing a connection string with user information but without a username or password (e.g. `"mongodb://@localhost.com/"`).

In the cases where no auth mechanism is specified, I'm passing `:default` into the error class. Let me know if you think this is clear enough.

lib/mongo/client.rb
```diff
@@ -855,14 +857,32 @@ def validate_new_options!(opts = Options::Redacted.new)
...
     # not allowed.
     def validate_options!
       if options[:write] && options[:write_concern] && options[:write] != options[:write_concern]
         raise ArgumentError, "If :write and :write_concern are both given, they must be identical: #{options.inspect}"
       end
     end
 
+    # Validates all authentication-related options after they are set on the client
+    # This method is intended to catch combinations of options which are not allowed
+    def validate_authentication_options!
```
egiurleo: It might also be good to extract this method and validate auth options passed to the `Mongo::Auth::User` object, but I think that's slightly outside the scope of [[RUBY-1745]](https://jira.mongodb.org/browse/RUBY-1745) (but open to other opinions there).

lib/mongo/uri.rb
```diff
@@ -405,27 +405,23 @@ def parse_uri_options!(string)
 
     def parse_user!(string)
       if (string && user = string.partition(AUTH_USER_PWD_DELIM)[0])
-        if user.length > 0
```
egiurleo: Removed this line to allow for a distinction between a URI like `"mongodb://localhost/"` (valid) and `"mongodb://@localhost.com/"` (invalid).

.evergreen/run-enterprise-auth-tests.sh
```diff
@@ -23,5 +23,10 @@ echo "Install dependencies"
 export BUNDLE_GEMFILE=gemfiles/mongo_kerberos.gemfile
 bundle install --gemfile="$BUNDLE_GEMFILE"
 
+unset CI
 echo "Running tests"
 bundle exec rspec spec/enterprise_auth -fd
+bundle exec rspec spec/spec_tests/uri_options_spec.rb -fd
+bundle exec rspec spec/spec_tests/connection_string_spec.rb -fd
+bundle exec rspec spec/mongo/uri/srv_protocol_spec.rb -fd
+bundle exec rspec spec/mongo/uri_spec.rb -fd
```
egiurleo: I opted to only run files that have relevant tests to keep CI as speedy as possible. That means that anybody who adds kerberos-dependent tests to a different file will have to remember to include that file.

spec/lite_spec_helper.rb
```diff
@@ -57,6 +57,7 @@
 require 'support/sdam_monitoring'
 require 'support/crud'
 require 'support/command_monitoring'
+require 'support/cluster_tools'
```
egiurleo: The tests in `spec_tests/uri_options/spec` rely on `ClusterTools` being present and will throw errors if run without this require.

spec/spec_tests/data/uri_options/auth-options.yml
```diff
@@ -1,6 +1,6 @@
 tests:
     -
-        description: "Valid auth options are parsed correctly"
+        description: "Valid auth options are parsed correctly (GSSAPI)"
```
egiurleo: Submitted a spec ticket to add this to specifications repo: [https://jira.mongodb.org/browse/SPEC-1423](https://jira.mongodb.org/browse/SPEC-1423)

.evergreen/run-enterprise-auth-tests.sh
```diff
@@ -23,5 +23,10 @@ echo "Install dependencies"
 export BUNDLE_GEMFILE=gemfiles/mongo_kerberos.gemfile
 bundle install --gemfile="$BUNDLE_GEMFILE"
 
+unset CI
```
egiurleo: When this env variable is set, `lite_spec_helper.rb` sets the formatter as `JsonExtFormatter`, which causes a `NoMethodError: undefined method `seed' for #<JsonExtFormatter:0x0000000002ec4f58>` error when rspec tries to output anything. I'm guessing there's another gem I have to require to fix this, but I'm not sure which.

spec/spec_tests/uri_options_spec.rb
```diff
@@ -1,4 +1,4 @@
-require 'lite_spec_helper'
+require 'spec_helper'
```
egiurleo: Changing this because this spec requires `support/cluster_config`.

This causes every test in `uri_options_spec` and `connection_string_spec` to fail with `NoAvailableServer` errors (see [here](https://evergreen.mongodb.com/task_log_raw/mongo_ruby_driver_enterprise_auth_tests_ubuntu__auth_and_ssl~auth_and_ssl_ruby~jruby_9.1_enterprise_auth_tests_patch_310299025045a26be3fa769dd9b45373108968f9_5d5db6e50ae6061643ad02b1_19_08_21_21_25_59/0?type=T&text=true)), only when run from `run-enterprise-auth-tests`. There is probably a gem that I need to require. If you have an idea of what that might be, feel free to let me know. I will investigate more tomorrow.

lib/mongo/auth.rb
```diff
@@ -70,10 +70,29 @@ def get(user)
...
+    class InvalidConfiguration < Mongo::Error::AuthError
+
+      # Instantiate the new error
+      #
+      # @example Instantiate the error
+      #   Mongo::Auth::InvalidConfiguration.new(:mongodb_cr, "User required")
+      #
+      # @param [ Symbol ] mechanism The provided mechanism.
+      # @param [ String ] reason The reason that this configuration is invalid.
```
p-mongo: This docstring is outdated and example can be omitted.

lib/mongo/auth.rb
```diff
@@ -70,10 +70,29 @@ def get(user)
...
+      #
+      # @since 2.11.0
+      def initialize(message)
+        super("Invalid auth configuration: #{message}")
```
p-mongo: The constructor can be omitted.

lib/mongo/client.rb
```diff
@@ -855,14 +857,42 @@ def validate_new_options!(opts = Options::Redacted.new)
...
+
+      if options[:auth_mech].nil?
+        if !options[:user].nil? && !options[:user].length
+          raise Mongo::Auth::InvalidConfiguration.new('user information delimiter (@) must not be included if user is blank')
+        elsif !options[:password].nil? && !options[:password].length
+          raise Mongo::Auth::InvalidConfiguration.new('password delimeter (:) must not be included if password is blank')
+        end
+      end
+
+      if !Mongo::Auth::SOURCES.has_key?(options[:auth_mech]) && !options[:auth_mech].nil?
```
p-mongo: Can the nil check be done first please.

lib/mongo/client.rb
```diff
@@ -855,14 +857,42 @@ def validate_new_options!(opts = Options::Redacted.new)
...
+      if options[:auth_mech] == :mongodb_x509
+        if ![:external, nil].include?(options[:auth_source])
+          raise Mongo::Auth::InvalidConfiguration.new("#{options[:auth_source]} is an invalid auth source for mechanism mongodb_x509; valid options are $external and nil")
+        elsif !options[:password].nil?
+          raise Mongo::Auth::InvalidConfiguration.new('password is not supported for mongodb_x509 auth mechanism')
+        end
+      end
+
+      if options[:auth_mech].nil?
+        if !options[:user].nil? && !options[:user].length
```
p-mongo: `!options[:user].length` is always false, suggest  `options[:user].empty?` instead.

lib/mongo/client.rb
```diff
@@ -855,14 +857,42 @@ def validate_new_options!(opts = Options::Redacted.new)
...
+          raise Mongo::Auth::InvalidConfiguration.new('user information delimiter (@) must not be included if user is blank')
```
p-mongo: The messages should also make sense when a client is constructed from host list and options, not URI.

PR #1501, opened 2019-08-22
RUBY-1868 only disconnect clusters of local clients when these cluste…

p-mongo:
…rs are not shared with global clients

### DIFF
--- a/spec/support/client_registry.rb
+++ b/spec/support/client_registry.rb
@@ -210,7 +210,14 @@ class ClientRegistry
   def close_local_clients
     @lock.synchronize do
       @local_clients.each do |client|
-        client.close(true)
+        cluster = client.cluster
+        # Disconnect this cluster if and only if it is not shared with
+        # any of the global clients we know about.
+        if @global_clients.none? { |name, global_client|
+          cluster == global_client.cluster
+        }
+          cluster.disconnect!
+        end
       end
       @local_clients = []
     end


p-mongo:
Summary of changes:

1. Change client registry to disconnect clusters rather than clients, and to only disconnect those clusters not also used by any of the known global clients. This restores performance to what it was prior to https://github.com/mongodb/mongo-ruby-driver/pull/1486 (~15 minutes for longest test vs ~1 hour on current master).
2. Adds a helper method to stop monitoring threads on a client, where a test needs a fully functioning client but performs global assertions which can be interfered with by background monitoring of said client.
3. Adds tracking of all `BackgroundThread` objects, building on refactoring done for background connection pool thread/SRV polling. This tracking allows positively asserting that no background threads are running in the helper which closes all clients.
4. Adds a local resource tracking helper for connection pools, since there are tests which create standalone pools.
5. Changes server disconnect method to use the `@pool` instance variable rather than `pool` accessor, to avoid instantiating a pool if one hadn't already existed (such as when the server never transitioned out of unknown state).
6. Changes several tests to use the tracking mentioned above.
7. Changes a couple of tests to use `ClusterConfig` for client options rather than taking them off of `authorized_client`, removing some needless client instantiations/reconnections.
8. Adds logging of object count, obtained from `ObjectSpace`, to ensure the number of outstanding objects does not grow unboundedly throughout the test suite execution.
9. Java heap space size bumped to 2G as 1G was insufficient.

After these changes jruby 9.2 appears to be green, jruby 9.1 appears to be still red as before and will be untested in a follow-up PR.

PR #1503, opened 2019-08-23
Add nil protection.

jonhyman:
We ran into this issue earlier this summer and had to restart our application to support it. There are other nil handling protections for session being nil in other files, so this adds more protection.

### DIFF
--- a/lib/mongo/retryable.rb
+++ b/lib/mongo/retryable.rb
@@ -210,7 +210,7 @@ module Mongo
         txn_num = session.in_transaction? ? session.txn_num : session.next_txn_num
         yield(server, txn_num, false)
       rescue Error::SocketError, Error::SocketTimeoutError => e
-        if session.in_transaction? && !ending_transaction
+        if session && session.in_transaction? && !ending_transaction
           raise
         end
         retry_write(e, session, txn_num, &block)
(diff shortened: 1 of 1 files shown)


p-mongo:
Thank you for the patch. As the comment on lines 200-201 states, if execution gets to line 203 the session must be present. Can you clarify why you think session may be nil on lines >= 203?

lib/mongo/retryable.rb
```diff
@@ -296,12 +296,12 @@ def modern_read_with_retry(session, server_selector, &block)
       begin
         yield server
       rescue Error::SocketError, Error::SocketTimeoutError => e
-        if session.in_transaction?
+        if session && session.in_transaction?
```
p-mongo: `modern_read_with_retry` is only called from line 117 in this file, which is the branch of an if statement conditional on session being present.

jonhyman:
I see; perhaps this wasn't the issue we had encountered. If those are guaranteed to be defined there then feel free to close out.

p-mongo:
You submitted https://github.com/mongodb/mongo-ruby-driver/pull/1048 last year. After investigating the issue, the fix we made was https://github.com/mongodb/mongo-ruby-driver/pull/1222/files. In that PR we did indeed add a check for the session being nil in error label handling code, but analysis of retryable logic showed that a session must be not nil when modern retries are performed (which makes sense because modern retryable writes require sessions in order to function).

I'll close this PR as being superseded by #1222, but if you still have this issue or other issues please let us know.

jonhyman:
Thank you for the analysis and response!

PR #1510, opened 2019-08-29
RUBY-1258 Implement auth spec tests

egiurleo:
Implements auth spec test runner. In addition, implements some rules around setting default values for the auth source and auth mechanism property options.


lib/mongo/database.rb
```diff
@@ -29,6 +29,11 @@ class Database
     # @since 2.0.0
     ADMIN = 'admin'.freeze
 
+    # The external database name.
+    #
+    # @since 2.11.0
+    EXTERNAL = :external
```
egiurleo: @p-mongo you mentioned in [RUBY-1563](https://github.com/mongodb/mongo-ruby-driver/pull/1514) preferring inline strings to class constants. I wanted to make this consistent with the ADMIN constant, but I can remove it if you don't want to add more constants.

lib/mongo/uri.rb
```diff
@@ -241,10 +241,12 @@ def self.get(string, opts = {})
     #
     # @since 2.0.0
     def client_options
-      opts = uri_options.merge(:database => database)
+      opts = default_client_options.merge(uri_options)
+      # opts = uri_options.merge(:database => database)
```
egiurleo: ignore this, removing

lib/mongo/uri.rb
```diff
@@ -241,10 +241,12 @@ def self.get(string, opts = {})
...
       @user ? opts.merge(credentials) : opts
     end
 
+
```
egiurleo: ignore this too

PR #1512, opened 2019-08-29
RUBY-1868 omit monitoring on server instances since we manually add/remove them to cluster


p-mongo:


p-mongo:
Fixes the failure in https://github.com/mongodb/mongo-ruby-driver/pull/1511.

PR #1514, opened 2019-08-30
RUBY-1563 preparation/refactoring

p-mongo:
This PR primarily deals with validation of data involved in DNS seed list processing which will also be used for SRV polling. Specifically it extracts the validation code into reusable methods/modules, improves exception messages to be more correct/informative/actionable and adds test coverage for some of the validations.


lib/mongo/uri.rb
```diff
@@ -64,6 +65,7 @@ class URI
     # Error details for an invalid scheme.
     #
     # @since 2.1.0
+    # @deprecated
     INVALID_SCHEME = "Invalid scheme. Scheme must be '#{MONGODB_SCHEME}' or '#{MONGODB_SRV_SCHEME}'".freeze
```
egiurleo: What is the benefit of keeping this constant in the codebase if it is no longer being used? Documentation purposes?

p-mongo: This constant is part of the driver's public API, thus we keep them all around until next major release in case any of the applications make use of them for some reason.

PR #1516, opened 2019-09-03
Fix selectable null handling, RUBY-1911.

jonhyman:
This pull attempts to ensure that `average_round_trip_time` does not change out from under us while evaluating.


p-mongo:
Closing in favor of https://github.com/mongodb/mongo-ruby-driver/pull/1517.

PR #1522, opened 2019-09-09
Use errorContains: E11000 in all spec tests formerly referencing DuplicateKey


p-mongo:


p-mongo:
Spec change: https://github.com/mongodb/specifications/pull/586

PR #1524, opened 2019-09-12
RUBY-1563 Poll SRV records in unknown topologies

p-mongo:
https://jira.mongodb.com/browse/RUBY-1563

### DIFF
--- a/spec/integration/reconnect_spec.rb
+++ b/spec/integration/reconnect_spec.rb
@@ -81,13 +81,17 @@ describe 'Client after reconnect' do
     end
 
     shared_examples_for 'recreates SRV monitor' do
+      # JRuby produces this error:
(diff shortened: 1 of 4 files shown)


spec/integration/reconnect_spec.rb
```diff
@@ -75,37 +72,107 @@
...
+      # This is bizarre because if thread was nil, the earlier call to
+      # thread.kill should've similarly failed, but it doesn't.
+      fails_on_jruby
+
+      it 'recreates SRV monitor' do
+        wait_for_discovery
+
+        if BSON::Environment.jruby?
+          # Wait for jruby to start SRV monitor thread
+          sleep 1
```
egiurleo: Is this necessary if we're skipping these tests because they fail on jruby?

p-mongo: No, removed this bit as well.

PR #1534, opened 2019-09-24
Fix RUBY-1944 Background threads such as the Monitor cannot restart once stopped

jonhyman:
Threads that are stopped via a disconnect can't actually be restarted:

```ruby
> server.instance_variable_get(:@monitor).instance_variable_get(:@thread).alive?
=> true
> server.disconnect!
=> true
> server.reconnect!
=> true
> server.instance_variable_get(:@monitor).instance_variable_get(:@thread).alive?
=> false
> server.instance_variable_get(:@monitor).instance_variable_get(:@thread).alive?
=> false
> server.instance_variable_get(:@monitor).instance_variable_get(:@thread).alive?
=> false
```

I can't run tests for some reason, likely due to my config perhaps I don't have some kind of environment variables set up. But I believe this test should pass based on the example code above.

```ruby
     NoMethodError:
       undefined method `map' for true:TrueClass
       Did you mean?  tap
     # ./lib/mongo/cluster.rb:129:in `initialize'
     # ./lib/mongo/client.rb:421:in `new'
     # ./lib/mongo/client.rb:421:in `initialize'
     # ./spec/support/spec_config.rb:70:in `new'
     # ./spec/support/spec_config.rb:70:in `connect_options'
     # ./spec/support/spec_config.rb:359:in `test_options'
     # ./spec/support/client_registry.rb:170:in `new_global_client'
     # ./spec/support/client_registry.rb:83:in `global_client'
     # ./spec/support/cluster_config.rb:124:in `determine_cluster_config'
     # ./spec/support/cluster_config.rb:17:in `server_version'
     # ./spec/support/cluster_config.rb:35:in `fcv_ish'
     # ./spec/spec_helper.rb:22:in `block (2 levels) in <top (required)>'
```

### DIFF
--- a/lib/mongo/background_thread.rb
+++ b/lib/mongo/background_thread.rb
@@ -114,6 +114,7 @@ module Mongo
 
     # @return [ Thread ] The created Thread instance.
     def start!
+      @stop_requested = nil
       @thread = Thread.new do
         catch(:done) do
           until @stop_requested
--- a/lib/mongo/background_thread.rb
+++ b/lib/mongo/background_thread.rb
@@ -30,6 +30,14 @@ module Mongo
     #
     # @api public for backwards compatibility only
     def run!
+      if @stop_requested && @thread
+        wait_for_stop
+        if @thread.alive?
+          log_warn("Starting a new background thread in #{self}, but the previous background thread is still running")
+          @thread = nil
+        end
+        @stop_requested = false
+      end
       if running?
         @thread
       else
@@ -81,6 +89,25 @@ module Mongo
       # the middle of an operation.
       @thread.kill
 
+      wait_for_stop
+    end
+
+    private
+
+    def start!
+      @thread = Thread.new do
+        catch(:done) do
+          until @stop_requested
+            do_work
+          end
+        end
+      end
+    end
+
+    # Waits for the thread to die, with a timeout.
+    #
+    # Returns true if the thread died, false otherwise.
+    def wait_for_stop
       # Wait for the thread to die. This is important in order to reliably
       # clean up resources like connections knowing that no background
       # thread will reconnect because it is still working.
@@ -106,21 +133,8 @@ module Mongo
         false
       else
         @thread = nil
+        @stop_requested = false
         true
-      end
-    end
-
-    private
-
-    # @return [ Thread ] The created Thread instance.
-    def start!
-      @stop_requested = nil
-      @thread = Thread.new do
-        catch(:done) do
-          until @stop_requested
-            do_work
-          end
-        end
       end
     end
 
--- a/spec/mongo/server/monitor_spec.rb
+++ b/spec/mongo/server/monitor_spec.rb
@@ -158,7 +158,7 @@ describe Mongo::Server::Monitor do
   end
 =end
 
-  describe '#restart!' do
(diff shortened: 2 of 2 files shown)


p-mongo:
Thank you for the patch.

We have documentation for running the tests at https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/README.md, are you not successful following it?

p-mongo:
I changed the patch to use `run!` which is the public API of `BackgroundThread` rather than `start!` which is its private API.

lib/mongo/background_thread.rb
```diff
@@ -30,6 +30,14 @@ module BackgroundThread
     #
     # @api public for backwards compatibility only
     def run!
+      if @stop_requested && @thread
+        wait_for_stop
+        if @thread.alive?
+          log_warn("Starting a new background thread in #{self}, but the previous background thread is still running")
+          @thread = nil
```
egiurleo: It's possible I'm misunderstanding what's going on here -- it looks like we're removing a reference to a thread that's still running, and then starting a new thread. Is that correct, and if so, is there any danger in leaving the previous background thread to run?

p-mongo: Yes, it's not great. The alternative is to not start the new thread, but then the old thread may exit eventually, or the old thread may be broken/stuck. Do you have any thoughts on this?

egiurleo: If `@stop_requested` is true, that means that `stop!` has been called on this `BackgroundThread` instance, and that `@thread.kill` has been called. How likely is it that this won't actually kill the thread? Are there particular cases where the `.kill` method doesn't work as intended?

p-mongo: It seems that on Ruby 2.6 kill succeeds when the thread is both doing CPU and I/O, as follows:

<pre>
irb(main):033:0> start=Time.now; a=Thread.new{ loop { 2**1000000 } }; a.kill; while a.alive?; sleep 0.01; end; p Time.now-start
0.010240296
=> 0.010240296
irb(main):034:0> 
</pre>

<pre>
irb(main):032:0> start=Time.now; a=Thread.new{ open('/mnt/speed/123').read }; a.kill; while a.alive?; sleep 0.01; end; p Time.now-start
0.010207387
=> 0.010207387
</pre>

However, kill only marks the thread for termination as implied by https://ruby-doc.org/core-2.6.3/Thread.html#method-c-kill and https://ruby-doc.org/core-2.6.3/Thread.html#method-c-exit, and does not immediately terminate the thread:

<pre>
irb(main):005:0> a=Thread.new{ open('abc.xyz').read }; a.kill; a.alive?
=> true
</pre>

egiurleo: Thanks for the context. One last question -- if the thread was broken/stuck, would there be any way to test for that and force it to exit?

p-mongo: I haven't been able to make a thread get stuck yet, hence how to produce this is currently a mystery to me.

As far as forcing it to exit, this is what Thread#kill is meant to do.

egiurleo: Okay, then I think this is fine. Thank you for answering all my questions!

PR #1535, opened 2019-09-24
RUBY-1913 Do not convert auth source from $external to :external because server does not recognize symbol database names

egiurleo:
Part one of [RUBY-1913](https://jira.mongodb.org/browse/RUBY-1913)

Any references to the `$external` auth source was being changed into a symbol, which is not a type that the server recognizes. This removes all references to `:external` and deletes related code.

### DIFF
--- a/spec/mongo/uri/srv_protocol_spec.rb
+++ b/spec/mongo/uri/srv_protocol_spec.rb
@@ -716,9 +716,9 @@ describe Mongo::URI::SRVProtocol do
 
         context '$external' do
           let(:source) { '$external' }
-          let(:expected) { :external }
+          let(:expected) { '$external' }
 
-          it 'sets the auth source to :external' do
+          it 'sets the auth source to '$external'' do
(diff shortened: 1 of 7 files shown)


spec/mongo/uri/srv_protocol_spec.rb
```diff
@@ -713,19 +713,6 @@
             expect(client.options[:auth_source]).to eq(source)
           end
         end
-
-        context '$external' do
```
p-mongo: Do you think there is no value in keeping these tests?

egiurleo: There are still tests starting on line 702 that test that an auth source is set on the client as passed in on the URI string. Since there's no longer any special treatment for `$external`, this test seems unnecessary.

p-mongo: OK sounds good.

PR #1536, opened 2019-09-25
RUBY-1913 Client auth option integration tests

egiurleo:
This is part two of my work for [RUBY-1913](https://jira.mongodb.org/browse/RUBY-1913).

In [RUBY-1258](https://github.com/mongodb/mongo-ruby-driver/commit/c31f62c0d217727ca466db862469ceaddf76b583#diff-e7786d37b967fce59a4cd10176f3a85aR441), I updated Mongo::URI to set some default options that it passes to the client. This works if a user is connecting to a MongoDB instance using exclusively URI options, but it doesn't help if the user wants to pass those options directly into the client and get the same defaults.

For example:
```
> client = Mongo::Client.new('mongodb://localhost:27018/?authMechanism=MONGODB-X509&tls=true&tlsCAFile=/path/to/ca&tlsCertificateKeyFile=/path/to/cert')
> client.options
=> {"database"=>"admin", "auth_source"=>"$external", "auth_mech"=>:mongodb_x509, "ssl"=>true, "ssl_ca_cert"=>"/path/to/ca", "ssl_cert"=>"/path/to/cert", "ssl_key"=>"/path/to/cert", "retry_reads"=>true, "retry_writes"=>true}
# notice that it sets the auth source and the ssl_key; this is good

> client2 = Mongo::Client.new('mongodb://localhost:27018',
   {
      auth_mech: :mongodb_x509,
      ssl: true,
      ssl_cert: '/path/to/cert', 
      ssl_ca_cert: '/path/to/ca'
   }
)
> client2.options
=> {"database"=>"admin", "auth_mech"=>:mongodb_x509, "ssl"=>true, "ssl_cert"=>"/path/to/cert", "ssl_ca_cert"=>"/path/to/ca", "retry_reads"=>true, "retry_writes"=>true}
# there's no auth source or ssl_key; the client won't be able to connect and authenticate
```

In this PR, I add integration tests for auth options in the client. These integration tests check that all authentication-related default options are set at the appropriate times. They also check that the same results are achieved whether the user is passing options through the URI or through the client options.

There are some tests that are commented out because they currently fail (see code snippet above). In my next PR, I will update the client so that those tests pass.

I will also open a PR with integration tests that check that clients can actually connect to a server and insert a document (especially with x509 authentication).

**This PR does not change client behavior**

### DIFF
--- a/spec/integration/client_options_spec.rb
+++ b/spec/integration/client_options_spec.rb
@@ -339,10 +339,10 @@ describe 'Client options' do
   context 'with no auth mechanism provided' do
     context 'with URI options' do
       context 'with no credentials' do
-          it 'creates a client with epty credentials' do
+        it 'creates a client with epty credentials' do
           expect(client.options[:user]).to be_nil
           expect(client.options[:password]).to be_nil
-          end
+        end
       end
 
       context 'with empty credentials' do
@@ -358,10 +358,10 @@ describe 'Client options' do
 
     context 'with client options' do
       context 'with no credentials' do
-          it 'creates a client with epty credentials' do
+        it 'creates a client with empty credentials' do
           expect(client.options[:user]).to be_nil
           expect(client.options[:password]).to be_nil
-          end
+        end
       end
 
       context 'with empty credentials' do
--- a/spec/integration/client_options_spec.rb
+++ b/spec/integration/client_options_spec.rb
@@ -349,7 +349,7 @@ describe 'Client options' do
       context 'with empty username' do
         let(:credentials) { '@' }
 
-        it 'does not allow a client to be created' do
+        it 'raises an exception' do
           expect {
             client
           }.to raise_error(Mongo::Auth::InvalidConfiguration, /empty username is not supported/)
@@ -359,16 +359,16 @@ describe 'Client options' do
 
     context 'with client options' do
       context 'with no credentials' do
-        it 'creates a client with empty credentials' do
+        it 'creates a client without credentials' do
           expect(client.options[:user]).to be_nil
           expect(client.options[:password]).to be_nil
         end
       end
 
-      context 'with empty credentials' do
+      context 'with empty username' do
         let(:client_opts) { { user: '', password: '' } }
 
-        it 'does not allow a client to be created with no username' do
+        it 'raises an exception' do
           expect {
             client
           }.to raise_error(Mongo::Auth::InvalidConfiguration, /empty username is not supported/)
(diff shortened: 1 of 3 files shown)


spec/integration/client_options_spec.rb
```diff
@@ -0,0 +1,441 @@
...
+        it 'creates a client with empty credentials' do
+          expect(client.options[:user]).to be_nil
+          expect(client.options[:password]).to be_nil
+        end
+      end
+
+      context 'with empty credentials' do
+        let(:client_opts) { { user: '', password: '' } }
+
+        it 'does not allow a client to be created with no username' do
```
p-mongo: Either "with empty username" or just say "does not allow a client to be created". An empty string is different from no input.

spec/integration/client_options_spec.rb
```diff
@@ -0,0 +1,441 @@
...
+          }.to raise_error(Mongo::Auth::InvalidConfiguration, /password is not supported/)
+        end
+      end
+    end
+  end
+
+  context 'with no auth mechanism provided' do
+    context 'with URI options' do
+      context 'with no credentials' do
+        it 'creates a client with epty credentials' do
```
p-mongo: "without credentials"

spec/integration/client_options_spec.rb
```diff
@@ -0,0 +1,441 @@
...
+
+        it 'recognizes the mechanism with no username' do
+          expect(client.options[:user]).to be_nil
+        end
+      end
+
+      context 'when a password is provided' do
+        let(:client_opts) { { auth_mech: :mongodb_x509, user: user, password: 'password' } }
+
+        it 'throws an error on client creation' do
```
p-mongo: In Ruby exceptions are raised. There is also a throw keyword which does something else https://stackoverflow.com/questions/51021/what-is-the-difference-between-raising-exceptions-vs-throwing-exceptions-in-ruby).

spec/integration/client_options_spec.rb
```diff
@@ -0,0 +1,441 @@
...
+      #   it 'creates a client with database as auth source' do
+      #     expect(client.options['auth_source']).to eq(database)
+      #   end
+      # end
+    end
+  end
+
+  shared_examples_for 'an auth mechanism with ssl' do
+    let(:ca_file) { '/path/to/ca.pem' }
+    let(:cert) { '/path/to/client.pem' }
```
p-mongo: Can these be renamed to `ca_file_path` and `cert_path` please?

PR #1538, opened 2019-09-26
RUBY-1945 Prohibit distinct(nil) call as server 4.4 rejects it


p-mongo:


p-mongo:
ruby-head failures are repaired in https://github.com/mongodb/mongo-ruby-driver/pull/1539.

PR #1540, opened 2019-09-26
RUBY-1948 Set Client default options on Client, not URI

egiurleo:
Before this, default authentication client options were set on the URI object. This creates inconsistencies between creating a client using URI options vs. creating a client using Ruby options. In this PR, I move the code that creates default options to the Client class and add integration tests that ensure that options are generated consistently using URI and Ruby options.


lib/mongo/client.rb
```diff
@@ -814,6 +808,41 @@ def watch(pipeline = [], options = {})
...
+        end
+
+        if options[:auth_mech] == :gssapi
+          opts[:auth_mech_properties] = { service_name: 'mongodb' }
+        end
+
+        # Since we know that the only URI option that sets :ssl_cert is "tlsCertificateKeyFile", any
+        # value set for :ssl_cert must also be set for :ssl_key.
+        if options[:ssl_cert]
+          default_options[:ssl_key] = options[:ssl_cert]
```
p-mongo: This expansion makes sense for URI processing because one URI option maps to two Ruby options. The Ruby options however, as documented in https://docs.mongodb.com/ruby-driver/master/tutorials/ruby-driver-create-client/#ruby-options, are not the same. In particular, ssl_key is not documented to accept a file containing both a key and a certificate; if this is the case, can ssl_key documentation be updated to state this?

Furthermore, the ssl_key documentation currently states:

> Note that even if the key is stored in the same file as the certificate, both need to be explicitly specified. 

Thus we are apparently already requiring a user to provide both ssl_key and ssl_cert options explicitly.

egiurleo: Thanks for all this. I'll update it!

docs/tutorials/ruby-driver-authentication.txt
```diff
@@ -101,6 +101,13 @@ derived from the distinguished subject name of this certificate.
 This authentication method requires the use of SSL connections with
 certificate validation.
 
+To authenticate the client, you will need a valid SSL certificate
+and private encryption key. These can be stored in separate files,
+or together in one file (in the PEM format). Even if the certificate
+and private key are stored in the same file, you must specify the path to
+that file by passing both the ``ssl_cert`` and ``ssl_key`` options
+to the client.
```
egiurleo: I'd love feedback on this paragraph! Is it clear enough? Should I talk about how certificate files are generated, or is that not relevant?

p-mongo: I think it is well written as it is.

spec/support/client_registry.rb
```diff
@@ -100,12 +100,20 @@ def new_global_client(name)
...
     when 'authorized'
+      client_options = {
+        database: SpecConfig.instance.test_db,
+        user: SpecConfig.instance.test_user.name,
+        password: SpecConfig.instance.test_user.password,
+      }.tap do |opts|
+        # specify that the user was created using the SCRAM-SHA-1 mechanism if
+        # auth is turned on and MongoDB version is greater than 4.0
+        # (SCRAM-SHA-256 was introduced in 4.0)
+        opts[:auth_mech] = :scram if SpecConfig.instance.user && SpecConfig.instance.supports_scram_256?
```
egiurleo: This is my proposed fix for the [authentication errors](https://evergreen.mongodb.com/task_log_raw/mongo_ruby_driver_ruby_mid__mongodb_version~4.0_topology~sharded_cluster_auth_and_ssl~auth_and_ssl_ruby~ruby_2.5_os~rhel70_test_patch_a92b967961c216d34c74fdea967adc1fa7a68903_5d922b4057e85a5c6654d8c0_19_09_30_16_20_17/0?type=T) in `connection_pool_populator_spec` and `connection_pool_spec`.

Here is why those errors were occurring:
1. I changed `Mongo::Client` to always set a default `auth_source` if you pass in user credentials (as per the authentication spec). This means that the `authorized` and `root_authorized` clients will always have an auth source set in their options.
2. In certain tests, the client options are passed to the app_metadata object via a [connection handshake](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/server/connection.rb#L309).
3. The app_metadata object sends [different options](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/server/app_metadata.rb#L76) to the server if it is given an auth source.
4. This causes the server to respond with a list of viable authentication mechanisms where it wasn't before. This is not a problem unless the server version is 4.0 or greater, where both SCRAM-SHA-1 and SCRAM-SHA-256 are both supported. In this case, the post-handshake logic will [set the default auth mech to SCRAM-SHA-256](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/server/connection.rb#L337).
5. This would not be a problem.... except mlaunch [does not create users using the SCRAM-SHA-256 auth mechanism](https://github.com/rueckstiess/mtools/blob/develop/mtools/mlaunch/mlaunch.py#L1699). So even though SCRAM-SHA-256 is the default auth mechanism for these servers, trying to authenticate with that mechanism will cause an error.

Specifying that the users are created using the SCRAM-SHA-1 mechanism fixes these errors, and it's more explicit, which I like. I'm also open to other opinions.

spec/integration/connection_spec.rb
```diff
@@ -159,7 +159,7 @@
           # now pretend an ismaster returned a different range
           features = Mongo::Server::Description::Features.new(0..3)
           # the second Features instantiation is for SDAM event publication
-          expect(Mongo::Server::Description::Features).to receive(:new).twice.and_return(features)
+          expect(Mongo::Server::Description::Features).to receive(:new).at_least(:once).and_return(features)
```
egiurleo: See comment below.

spec/support/client_registry.rb
p-mongo: If mlaunch is changed to create users with `opts = {'mechanisms': ['SCRAM-SHA-1','SCRAM-SHA-256']}`, would that fix everything?

egiurleo: Yes, I believe so, but there currently isn't a public API for those options. I'm happy to submit a pull to mlaunch at some point. What do you think is the best path forward now?

p-mongo: I have two concerns right now:

1. If a developer creates users without specifying any mechanisms, would our test suite still work? I believe it will because I am under the impression that without specifying mechanisms, a user will be created with both scram1 and scram256 mechanisms, but I could be mistaken.

2. The comment above the code being changed in this hunk no longer matches the code.

Assuming the first concern is a non-issue, adjusting the comment to reflect that we are temporarily forcing scram1 even on 4.2+ servers until mlaunch is fixed I think is sufficient for now.

egiurleo: You're correct about your first concern -- The tests will still work because users are created with both scram1 and scram256 mechanisms. I'll fix the comment now.

egiurleo: @p-mongo is the new comment clearer?

spec/support/spec_config.rb
```diff
@@ -92,6 +92,23 @@ def connect_options
     end
   end
 
+  def supports_scram_256?
```
p-mongo: This method should be in ClusterConfig which returns information about the deployment. SpecConfig is meant for information about how the test suite was configured.

spec/support/client_registry.rb
p-mongo: Yes I think this is good.

spec/support/spec_config.rb
p-mongo: This method should be in ClusterConfig which deals with how the cluster is configured rather than SpecConfig, which deals with how the test suite is configured.

spec/support/spec_config.rb
```diff
@@ -92,6 +92,23 @@ def connect_options
...
+    return @supports_scram_256 if @supports_scram_256
```
p-mongo: This is better spelled as `return @supports_scram_256 unless @supports_scram_256.nil?` which will short-circuit the rest of the method when the deployment does not support scram256.

spec/support/spec_config.rb
```diff
@@ -92,6 +92,23 @@ def connect_options
...
+
+    client = Mongo::Client.new(addresses, Mongo::Options::Redacted.new(
```
p-mongo: When this is moved to cluster config, this should use one of the global clients (`ClientRegistry.instance.global_client`) unless there's a reason why this won't work.

spec/support/client_registry.rb
```diff
@@ -1,4 +1,5 @@
 require 'singleton'
+require_relative './cluster_config'
```
egiurleo: Is there a better way to make `ClusterConfig` available here?

p-mongo:
The issue appears to be that app metadata contains one username and the connection is authenticated for another user. The two have different possible auth mechanisms, and with this PR the driver now appears to (correctly) perform auth mechanism negotiation where it previously didn't (in at least one case), and the failure was because negotiated mechanism was not appropriate for the user being authenticated.

p-mongo:
Merged as https://github.com/mongodb/mongo-ruby-driver/pull/1550.

PR #1547, opened 2019-10-02
Test users documentation

egiurleo:
Write up some documentation about the users created in our test suite and how they're used.

Preview the rendered markdown [here](https://github.com/egiurleo/mongo-ruby-driver/blob/test-users/spec/USERS.md).


spec/USERS.md
```diff
@@ -0,0 +1,70 @@
+# Test Users
```
egiurleo: I'm not sure if there are any conventions around including more than one markdown file in a folder. I named this file USERS.md, but I'm happy to change it if there's another naming scheme you prefer.

spec/USERS.md
```diff
@@ -0,0 +1,70 @@
...
+the script has verified the existence of `root_user`, it will create a client authenticated with the `root_user` and use that client to create a second user, `test_user`. Because `root_user` has the `userAdminAnyDatabase` role, it has the permissions necessary to create and destroy users on your MongoDB instance. If you have already created a user with the same credentials as `test_user` prior to running
+the `spec_setup` script, the script will delete this user and re-create it.
+
+The `root_user` is created in the `admin` database, while the `test_user` is created in the `ruby-driver`
+database.
+
+The authentication mechanism used to store the user credentials is going to change depending on the version of MongoDB running on your deployment. If you are running tests against an instance running a MongoDB server version older than 3.2, the users are created using the `MONGODB-CR` authentication mechanism. Starting in version 3.2,`MONGODB-CR` was replaced as the default authentication mechanism, so on any test instance running MongoDB 3.2 or newer, users will be created using the `SCRAM` authentication mechanism by default.
```
p-mongo: There are two scram mechanisms, right? "SCRAM mechanism" isn't actually a thing?

spec/USERS.md
p-mongo: I think it's fine for now, we could revisit in the future if more files are added.

spec/USERS.md
egiurleo: You're right! I'll adjust.

spec/USERS.md
```diff
@@ -0,0 +1,72 @@
...
+the `spec_setup` script, the script will delete this user and re-create it.
+
+The `root_user` is created in the `admin` database, while the `test_user` is created in the `ruby-driver`
+database.
+
+The authentication mechanism used to store the user credentials is going to change depending on the version of MongoDB running on your deployment. If you are running tests against a MongoDB instance with a server version older than 3.0, the users will be created using the `MONGODB-CR` authentication mechanism. If your server version is between 3.0 and 3.6 (inclusive), the test users will be created using the `SCRAM-SHA-1` mechanism, which was introduced as the new default starting in MongoDB version 3.0. If you are running a version of MongoDB newer than 4.0, test users will be authenticated using either `SCRAM-SHA-1` or `SCRAM-SHA-256`.
+
+**Note:** (m-launch)[http://blog.rueckstiess.com/mtools/mlaunch.html], the client tool we use to spin up MongoDB instances for our tests, creates users EXCLUSIVELY with the `SCRAM-SHA-1` mechanism, even when `SCRAM-SHA-256` is enabled on the test server. This should not impact your ability to run the Mongo Ruby Driver test suite.
+
```
egiurleo: @p-mongo I went more into detail about the different versions of scram. I also wrote a note about what we discussed in #1540 in case somebody runs into similar issues that I had. Let me know what you think.

PR #1550, opened 2019-10-08
Fix RUBY-1948 Default authentication options are set on URI and not on client

p-mongo:
https://jira.mongodb.com/browse/RUBY-1948


p-mongo:
The issue was incorrect test setup. The app_metadata object that was used in the mocking was built with the options of the authorized client, not the options of the test where the object was used. One of the differences between the two was the username.

PR #1554, opened 2019-10-11
RUBY-1919 Libmongocrypt binding

egiurleo:
First step in the FLE implementation: create a Ruby wrapper around libmongocrypt.

I've opted to use FFI as opposed to a Ruby extension here, but that's subject to change once I have implemented a larger proof of concept. I haven't packaged the libmongocrypt source code with the driver -- also subject to change.

To run tests, download a built version of libmongocrypt [here](https://s3.amazonaws.com/mciuploads/libmongocrypt/all/master/latest/libmongocrypt-all.tar.gz) and export a`LIBMONGOCRYPT_PATH` environment variable with a path to your libmongocrypt build.


lib/mongo/libmongocrypt.rb
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module Mongo
+  class Libmongocrypt
+    autoload(:Binding, 'mongo/libmongocrypt/binding.rb')
```
egiurleo: This is one possible way to have `Binding` only load when it's called. That way, if there's no libmongocrypt (path specified by an env variable for now, maybe a separate gem later), the user gets a specific error message when they try to use `Mongo::Libmongocrypt::Binding` rather than an error message when loading the entire driver.

Another way we could do this is by raising an error in the class and then catching it on `require` in `mongo.rb`, for example:
```
# in mongo.rb

begin
  require 'mongo/libmongocrypt'
rescue LibmongocryptError => e
  warn('Could not load Mongo::Libmongocrypt::Binding because...`)
end
```
I've honestly never used a pattern like this before in Ruby so I'm happy to chat more about tradeoffs or better ways to do it.

lib/mongo/libmongocrypt/binding.rb
```diff
@@ -0,0 +1,42 @@
...
+      rescue LoadError => e
+        if ENV['LIBMONGOCRYPT_PATH']
+          raise "Cannot load Mongo::Libmongocrypt::Binding because the path to " +
+            "libmongocrypt specified in the LIBMONGOCRYPT_PATH environment variable " +
+            "is invalid: #{ENV['LIBMONGOCRYPT']}\n\n#{e.message}"
+        else
+          raise "Cannot load Mongo::Libmongocrypt::Binding because there is no path " +
+            "to libmongocrypt specified in the LIBMONGOCRYPT_PATH environment variable."
+        end
+      end
```
egiurleo: This could change depending on the way this file gets loaded.

.evergreen/run-fle-tests.sh
```diff
@@ -0,0 +1,21 @@
...
+
+set -o xtrace   # Write all commands first to stderr
+set -o errexit  # Exit the script with error if any of the commands fail
+
+. `dirname "$0"`/functions.sh
+
+wget "https://s3.amazonaws.com/mciuploads/libmongocrypt/all/master/latest/libmongocrypt-all.tar.gz"
+tar -xvf libmongocrypt-all.tar.gz
+
+export MONGODB_URI='mongodb://localhost:27017'
```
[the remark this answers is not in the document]
egiurleo: No reason -- I'll remove.

lib/mongo/libmongocrypt/binding.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      begin
+        ffi_lib ENV['LIBMONGOCRYPT_PATH']
+      rescue LoadError => e
+        raise "Cannot load Mongo::Libmongocrypt::Binding because the path to " +
+          "libmongocrypt specified in the LIBMONGOCRYPT_PATH environment variable " +
+          "is invalid: #{ENV['LIBMONGOCRYPT']}\n\n#{e.class}: #{e.message}"
+      end
+
+      attach_function :mongocrypt_version, [:pointer], :string
```
p-mongo: Documentation for these methods.

spec/mongo/libmongocrypt/binding_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'lite_spec_helper'
+
+describe 'Mongo::Libmongocrypt::Binding' do
+  before do
+    unless ENV['LIBMONGOCRYPT_PATH']
```
p-mongo: Consider adding a method to SpecConfig instead of checking ENV in each test.

lib/mongo/libmongocrypt/binding.rb
```diff
@@ -0,0 +1,42 @@
...
+
+require 'ffi'
+
+module Mongo
+  class Libmongocrypt
+
+    # A Ruby binding for the libmongocrypt C library
+    #
+    # @since 2.12.0
+    class Binding
```
p-mongo: Is this class part of the driver's public API?

PR #1557, opened 2019-10-25
RUBY-1965 Bind mongocrypt_binary_t functions

egiurleo:
mongocrypt_binary_t is a non-owning byte buffer (read more in [libmongocrypt.h](https://github.com/mongodb/libmongocrypt/blob/master/src/mongocrypt.h.in#L69)). This is the first step of creating an idiomatic Ruby API in support of an explicit encryption POC.


lib/mongo/libmongocrypt.rb
```diff
@@ -13,7 +13,10 @@
...
 
 module Mongo
-  class Libmongocrypt
+  module Libmongocrypt
     autoload(:Binding, 'mongo/libmongocrypt/binding.rb')
   end
 end
+
+require 'mongo/libmongocrypt/binary.rb'
+require 'mongo/libmongocrypt/mongocrypt_error.rb'
```
egiurleo: I'm wondering if it would make more sense to autoload these as well so that FFI doesn't need to be loaded until it's actually going to be used (see how I included ffi in the enterprise auth and kerberos gemfiles above). Open to opinions.

p-mongo: I think it's a fine idea, and I believe the `.rb` suffix is unnecessary in the requires.

spec/mongo/libmongocrypt/binary_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+end
+
+describe Mongo::Libmongocrypt::Binary do
+  require_libmongocrypt
+
+  let(:bytes) { [104, 101, 108, 108, 111] }
+  let(:binary) { described_class.new(bytes) }
+
+  after(:each) do
+    binary.close if binary
```
p-mongo: This instantiates the binary, therefore the condition doesn't do anything meaningful.

lib/mongo/libmongocrypt/binary.rb
```diff
@@ -0,0 +1,77 @@
...
+        data.read_array_of_type(FFI::TYPE_UINT8, :read_uint8, len)
+      end
+
+      # Releases allocated memory and cleans up resources
+      #
+      # @return [ true ] Always true.
+      #
+      # @since 2.12.0
+      def close
+        Binding.mongocrypt_binary_destroy(@bin) if @bin
```
p-mongo: I think this class should also define a finalizer which would call this method.

lib/mongo/libmongocrypt/binary.rb
```diff
@@ -0,0 +1,77 @@
...
+      #
+      # @example Instantiate a Binary object
+      #   Mongo::Libmongocrypt::Binary.new([73, 76, 111, 118, 101, 82, 117, 98, 121])
+      #
+      # @param [ Array<Int> ] data An array of uint-8 bytes
+      #
+      # @since 2.12.0
+      def initialize(data)
+        unless data
+          raise MongocryptError.new('Cannot create new Binary object with no data.')
```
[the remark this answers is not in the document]
egiurleo: I'm happy with `Crypt`! I'll change the error class to `CryptError` in this PR and then update the `Libmongocrypt` module name in the next PR.

lib/mongo/libmongocrypt/binary.rb
```diff
@@ -0,0 +1,77 @@
...
+        # FFI::MemoryPointer automatically frees memory when it goes out of scope
+        @data_p = FFI::MemoryPointer.new(data.length)
+                  .write_array_of_type(FFI::TYPE_UINT8, :put_uint8, data)
+
+        @bin = Binding.mongocrypt_binary_new_from_data(@data_p, data.length)
+      end
+
+      # Returns the data stored as a byte array
+      #
+      # @return [ Array<Int> ] Byte array stored in mongocrypt_binary_t
```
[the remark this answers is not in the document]
egiurleo: Yeah, that sounds good. I'm not sure there's ever a case where the data needs to be exposed as a string, but I think that will reveal itself in the implementation.

spec/mongo/libmongocrypt/binary_spec.rb
egiurleo: 🤦‍♀ right

lib/mongo/libmongocrypt/binary.rb
egiurleo: We talked on Slack -- will revisit this once it's more obvious how this is used.

PR #1558, opened 2019-10-28
RUBY-1965 Add a binding for mongocrypt_status_t object

egiurleo:
The `mongocrypt_status_t` is a struct used to indicate the status of a `mongocrypt_t` object, either `ok`, `error_client`, or `error_kms`.

I haven't created an idiomatic API around it because I'm not sure I want to expose this to users as a `Status` class or if I want to do something with the `Mongo::Error::CryptError` class? I'm going to create the wrapping for the `mongocrypt_t` methods and see if that makes it clearer to me.

I also changed the name of the libmongocrypt module to `Crypt` and separated the `binding_spec` tests into multiple files.


spec/mongo/crypt/binding/binary_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'mongo'
+require 'support/lite_constraints'
+
+RSpec.configure do |config|
+  config.extend(LiteConstraints)
+end
+
+describe 'Mongo::Crypt::Binding' do
+  describe 'binary_t bindings' do
```
egiurleo: These were moved from the `spec/mongo/crypt/binding_spec` file because it was getting too big to read easily.

lib/mongo/crypt/binding.rb
```diff
@@ -61,6 +61,47 @@ class Binding
...
+      # the length of the returned string.
+      attach_function :mongocrypt_status_message, [:pointer, :pointer], :string
+
+      # Takes a pointer to a mongocrypt_status_t object and returns whether or not
+      # the status type is ok
+      attach_function :mongocrypt_status_ok, [:pointer], :bool
+
+      # Takes a pointer to a mongocrypt_status_t object and destroys the
+      # reference to that status
+      attach_function :mongocrypt_status_destroy, [:pointer], :void
```
egiurleo: I haven't created an idiomatic Ruby wrapping around these methods yet because I want to understand how they're used first.

PR #1559, opened 2019-10-29
Fix RUBY-1980 Ensure cursor IDs are always encoded as 64-bit integers in BSON for getMore


p-mongo:


spec/integration/get_more_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  let(:scope) do
+    collection.find.batch_size(1).each
+  end
+
+  before do
+    collection.delete_many
+    collection.insert_one(a: 1)
+    #collection.insert_one(a: 2)
```
egiurleo: Did you mean to leave this comment here?

p-mongo: Yes, the second insert can be helpful for ensuring there is enough data.

PR #1561, opened 2019-10-29
RUBY-1965 Introduce a block API for the Mongo::Crypt::Binary object

egiurleo:
It isn't great API design to force users of the libmongocrypt binding to do their own memory management, so I'm introducing an API that creates a new `Mongo::Crypt::Binary` object in the context of a block and then closes it upon completion of that block (or upon rescuing an error).

### DIFF
--- a/lib/mongo/crypt/binary.rb
+++ b/lib/mongo/crypt/binary.rb
@@ -84,9 +84,8 @@ module Mongo
       # @since 2.12.0
       def self.with_binary(data)
         binary = self.new(data)
-
         begin
-          yield binary
+          yield(binary)
         ensure
           binary.close
         end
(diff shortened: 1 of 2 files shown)


lib/mongo/crypt/binary.rb
```diff
@@ -72,6 +72,28 @@ def close
...
+      # @since 2.12.0
+      def self.with_binary(data)
+        binary = self.new(data)
+
+        begin
+          yield(binary)
+        rescue => e
+          binary.close
+          raise e
+        end
```
p-mongo: The pattern for this can be simpler:

<pre>
begin
  yield binary
ensure
  binary.close
end
</pre>

PR #1563, opened 2019-10-29
RUBY-1965 Add idiomatic Status API around mongocrypt_status_t object

egiurleo:


lib/mongo/crypt/status.rb
```diff
@@ -0,0 +1,125 @@
...
+      # Set a label, code, and message on the Status
+      #
+      # @param [ Symbol ] label One of :ok, :error_client, or :error_kms
+      # @param [ Integer ] code
+      # @param [ String ] message
+      #
+      # @return [ Status ] returns self
+      #
+      # @since 2.12.0
+      def update(label, code, message)
```
egiurleo: I might include these params in the initializer instead, but I'll wait to see how this is used.

lib/mongo/crypt.rb
```diff
@@ -16,5 +16,6 @@ module Mongo
   module Crypt
     autoload(:Binding, 'mongo/crypt/binding.rb')
     autoload(:Binary, 'mongo/crypt/binary.rb')
+    autoload(:Status, 'mongo/crypt/status.rb')
```
p-mongo: It is my impression that `autoload(:Status, 'mongo/crypt/status')` will work.

PR #1564, opened 2019-10-30
RUBY-1965 Create Mongocrypt handle binding

egiurleo:
This is the last PR for RUBY-1965 -- after this I should be able to initialize a ClientEncryption object and start using some of these objects.

In this PR:
- Change `Mongo::Crypt::Binary` to take a string as input rather than an array of bytes.
- Add binding for mongocrypt_t methods
- Add an API around mongocrypt_t called `Mongo::Crypt::Handle` (this is what it's called in the spec)
- Change the API of `Mongo::Error::CryptError` to be an error specifically returned from libmongocrypt

### DIFF
--- a/lib/mongo/crypt/handle.rb
+++ b/lib/mongo/crypt/handle.rb
@@ -25,17 +25,12 @@ module Mongo
 
       # Creates a new Handle object and initializes it with options
       #
-      # @example Instantiate a Handle object with local KMS provider options
-      #   Mongo::Crypt::Handle.new({
-      #     local: { key: 'MASTER-KEY' }
-      #   })
-      #
       # @param [ Hash ] kms_providers A hash of KMS settings. The only supported key
       # is currently :local. Local KMS options must be passed in the format
-      # { local: { key: 'MASTER-KEY' } } where the master key is a 96-byte, base64
+      # { local: { key: <master key> } } where the master key is a 96-byte, base64
       # encoded string.
       #
-      # @since 2.12.0
+      # There will be more arguemnts to this method once automatic encryption is introduced.
       def initialize(kms_providers)
         @mongocrypt = Binding.mongocrypt_new
 
@@ -43,6 +38,9 @@ module Mongo
           set_kms_providers(kms_providers)
           initialize_mongocrypt
         rescue => e
+          # Setting options or initializing mongocrypt_t could cause validation/status
+          # errors; if that happens, make sure the reference to the mongocrypt_t object
+          # is destroyed before passing on the error
           self.close
           raise e
         end
@@ -53,8 +51,6 @@ module Mongo
       # clean up resources
       #
       # @return [ true ] Always true
-      #
-      # @since 2.12.0
       def close
         Binding.mongocrypt_destroy(@mongocrypt) if @mongocrypt
         @mongocrypt = nil
@@ -68,7 +64,7 @@ module Mongo
       # information on the underlying mongocrypt_t object
       def set_kms_providers(kms_providers)
         unless kms_providers
-          raise ArgumentError.new("The kms_providers option must not be blank")
+          raise ArgumentError.new("The kms_providers option must not be nil")
         end
 
         unless kms_providers.key?(:local) || kms_providers.key?(:aws)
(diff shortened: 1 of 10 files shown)


lib/mongo/crypt/handle.rb
```diff
@@ -0,0 +1,129 @@
...
+      #     local: { key: 'MASTER-KEY' }
+      #   })
+      #
+      # @param [ Hash ] kms_providers A hash of KMS settings. The only supported key
+      # is currently :local. Local KMS options must be passed in the format
+      # { local: { key: 'MASTER-KEY' } } where the master key is a 96-byte, base64
+      # encoded string.
+      #
+      # @since 2.12.0
+      def initialize(kms_providers)
```
egiurleo: There will be one or two more params here, but they're not required for explicit encryption.

lib/mongo/crypt/handle.rb
```diff
@@ -0,0 +1,129 @@
...
+        @mongocrypt = Binding.mongocrypt_new
+
+        begin
+          set_kms_providers(kms_providers)
+          initialize_mongocrypt
+        rescue => e
+          self.close
+          raise e
```
egiurleo: If the user passes in an invalid KMS master key, or the kms_providers option is incorrectly formatted, there could be an error before this object is done initializing. I want to make sure that `@mongocrypt` ref is destroyed even if this happens.

lib/mongo/crypt/handle.rb
```diff
@@ -0,0 +1,129 @@
...
+    # A handle to the libmongocrypt library that wraps a mongocrypt_t object,
+    # allowing clients to set options on that object or perform operations such
+    # as encryption and decryption
+    class Handle
+
+      # Creates a new Handle object and initializes it with options
+      #
+      # @example Instantiate a Handle object with local KMS provider options
+      #   Mongo::Crypt::Handle.new({
+      #     local: { key: 'MASTER-KEY' }
```
egiurleo: Is this example helpful? How can I best communicate that the master key must be a 96-byte base64-encoded string?

lib/mongo/crypt/handle.rb
```diff
@@ -0,0 +1,129 @@
...
+        Binary.with_binary(Base64.decode64(master_key)) do |binary|
+          success = Binding.mongocrypt_setopt_kms_provider_local(@mongocrypt, binary.ref)
+          raise_from_status unless success
+        end
+      end
+
+      # Initialize the underlying mongocrypt_t object and raise an error if the operation fails
+      def initialize_mongocrypt
+        success = Binding.mongocrypt_init(@mongocrypt)
+        # There is currently no test for this code path
```
egiurleo: I think you can get this to fail based on other options you can pass in to `mongocrypt_t`, but I don't think it's possible to get to that state right now.

lib/mongo/crypt/handle.rb
p-mongo: Can comments like these be made in the source please rather than in the PR, so that they are available to anyone looking at the source in the future?

lib/mongo/crypt/handle.rb
p-mongo: Usually I tend to put this type of guidance/examples in the tutorial, where there is more flexibility with respect to expanding on a particular aspect of the driver.

egiurleo: That's really helpful, thanks :)

lib/mongo/crypt/handle.rb
egiurleo: Sure thing. I'll move over all my comments.

lib/mongo/error/crypt_error.rb
```diff
@@ -13,7 +13,23 @@ class Error
...
+      attr_accessor :code
+
+      def initialize(code, message)
+        @code = code
+        super("Code #{code}: #{message}")
+      end
+    end
+
+    # A libmongocrypt error relating to the client
+    class CryptClientError < CryptError; end
```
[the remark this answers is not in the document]
egiurleo: Agreed. I'll move them over once I start adding more to them.

spec/mongo/crypt/handle_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    context 'with valid local kms_providers' do
+      after do
+        handle.close
+      end
+
+      let(:kms_providers) do
+        {
+          local: {
+            key: Base64.encode64('ruby' * 24)
```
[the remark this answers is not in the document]
egiurleo: That's great.

PR #1566, opened 2019-10-31
RUBY-1925 ClientEncryption#initialize

egiurleo:
Use all the `Mongo::Crypt` code written so far to create an initializer for the `ClientEncryption` class, including instance variables for a key vault client and a `Mongo::Crypt::Handle` object.


lib/mongo/client_encryption.rb
```diff
@@ -0,0 +1,77 @@
...
+  class ClientEncryption
+
+    # Create a new ClientEncryption object with the provided options.
+    #
+    # @param [ Mongo::Client ] client A Mongo::Client
+    #   that is connected to the MongoDB instance where the key vault
+    #   collection is stored.
+    # @param [ Hash ] options The ClientEncryption options
+    #
+    # @option options [ String ] key_vault_namespace The name of the
```
p-mongo: We use `:key_vault_namespace` etc. to indicate that options must be passed as symbols.

lib/mongo/client_encryption.rb
```diff
@@ -0,0 +1,77 @@
...
+    #   key vault collection in the format "database.collection"
+    # @option options [ Hash ] kms_providers A hash of key management service
+    #   configuration information. Valid hash keys are :local or or :aws. There may be
```
p-mongo: `or` is duplicated.

spec/client_encryption_spec.rb
```diff
@@ -0,0 +1,71 @@
+require 'lite_spec_helper'
+require 'mongo'
+
+require 'base64'
+
+RSpec.configure do |config|
+  config.extend(LiteConstraints)
```
p-mongo: Why is this needed here?

lib/mongo/client_encryption.rb
egiurleo: Thanks!

spec/client_encryption_spec.rb
egiurleo: It isn't! Thanks.

egiurleo: It isn't! Thanks.

PR #1567, opened 2019-11-01
RUBY-1983 Create wrapper around mongocrypt_ctx_t

egiurleo:
mongocrypt_ctx_t manages the state machine for data key creation, encryption, and decryption. In this PR, I created an initial wrapper around this class with an initialization method that sets up a context specifically for creating data keys. Next, I'll wrap the methods necessary running the context's state machine.


lib/mongo/crypt/status.rb
```diff
@@ -84,6 +84,23 @@ def ref
         @status
       end
 
+      # Raises a Mongo::Error:CryptError corresponding to the
+      # information stored in this status
+      #
+      # Does nothing if self.ok? is true
+      def raise_crypt_error
```
egiurleo: I'm not sure if there's a better pattern I could use to accomplish this -- maybe 
```
raise Mongo::Error::CryptError.from_status(status)
```
Open to opinions here.

lib/mongo/crypt/handle.rb
```diff
@@ -108,16 +108,7 @@ def initialize_mongocrypt
...
-          error = case status.label
-          when :error_kms
-            # There is currently no test for this code path
-            Error::CryptKmsError.new(status.code, status.message)
-          when :error_client
-            Error::CryptClientError.new(status.code, status.message)
-          end
-
-          raise error
+          status.raise_crypt_error
```
egiurleo: I'm not sure if there's a better pattern I could use to accomplish this -- maybe
```
raise Mongo::Error::CryptError.from_status(status)
```
Open to opinions here.

lib/mongo/crypt/status.rb
```diff
@@ -84,6 +84,23 @@ def ref
...
+        return if ok?
+
+        error = case label
+        when :error_kms
+          Error::CryptKmsError.new(code, message)
+        when :error_client
+          Error::CryptClientError.new(code, message)
+        end
+
+        raise error
```
p-mongo: Symbols cannot be raised in Ruby, this will need to be revisited in the future to perhaps raise a base class of CryptError.

spec/mongo/crypt/data_key_context_spec.rb
```diff
@@ -0,0 +1,98 @@
+require 'mongo'
+require 'support/lite_constraints'
+
+RSpec.configure do |config|
+  config.extend(LiteConstraints)
+end
```
p-mongo: Also unneeded?

egiurleo: This one is needed because I've only imported `mongo` and `lite_constraints`, since I don't really need anything else in the spec helpers for this test. I can always import at least `lite_spec_helper` if you'd prefer!

p-mongo: I see, it's fine as is then

PR #1568, opened 2019-11-05
Fix RUBY-2004 Indicated attempt not always correct

p-mongo:
https://jira.mongodb.com/browse/RUBY-2004

### DIFF
--- a/spec/integration/retryable_errors_spec.rb
+++ b/spec/integration/retryable_errors_spec.rb
@@ -209,6 +209,8 @@ describe 'Failing retryable operations' do
       include_context 'read operation'
 
       context 'modern read retries' do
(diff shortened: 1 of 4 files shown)


spec/integration/retryable_errors_spec.rb
```diff
@@ -59,41 +45,223 @@
...
+
+        exception
+      end
+
+      let(:events) do
+        EventSubscriber.command_started_events('find')
+      end
+
+      it 'sends reads to different servers' do
+      end
```
egiurleo: Is this example meant to be empty?

p-mongo: This is in fact handled elsewhere, thank you for noticing.

PR #1569, opened 2019-11-05
RUBY-1984 ClientEncryption#create_data_key Proof of Concept

egiurleo:
In order to perform explicit encryption, Client Encryption first has to generate a key in the key vault database. This PR completes the work necessary to call `ClientEncryption#create_data_key`.

Usage:
```
require 'base64'

client = Mongo::Client.new(['localhost:27107'])
client_encryption_opts = {
  kms_providers: { local: { key: Base64.encode64("ru\xfe\x00" * 24) } },
  key_vault_namespace: 'test_db.keys'
}

client_encryption = Mongo::ClientEncryption.new(client, client_encryption_opts)
client_encryption.create_data_key 
# Generates key and inserts it into key vault namespace
# => Returns BSON::Binary object wrapping _id of new key
```


lib/mongo/client_encryption.rb
```diff
@@ -58,6 +58,27 @@ def close
...
+
+      Crypt::DataKeyContext.with_context(@crypt_handle.ref) do |context|
+        result = context.run_state_machine
+      end
+
+      data_key = Hash.from_bson(
+        BSON::ByteBuffer.new.put_bytes(result)
+      )
+
+      insert_result = @client.use(@key_vault_db)[@key_vault_coll].insert_one(data_key)
```
p-mongo: I expect "data key" to be a literal value, like a string. I created https://jira.mongodb.org/browse/SPEC-1490 since I was unable to locate the type of data key in the spec.

lib/mongo/crypt/binding.rb
```diff
@@ -148,6 +148,27 @@ class Binding
...
+
+      # mongocrypt_ctx_state_t type
+      enum :mongocrypt_ctx_state, [
+        :error,               0,
+        :need_mongo_collinfo, 1,
+        :need_mongo_markings, 2,
+        :need_mongo_keys,     3,
+        :need_kms,            4,
+        :ready,               5,
+        :done,                6
```
p-mongo: Would you mind adding trailing commas to lists like this?

lib/mongo/crypt/context/context.rb
```diff
@@ -42,6 +42,43 @@ def close
...
+      def state
+        Binding.mongocrypt_ctx_state(@ctx)
+      end
+
+      # Runs the mongocrypt_ctx_t state machine and handles
+      # all I/O on behalf of libmongocrypt
+      #
+      # @return [ String|nil ] A BSON string representing the outcome
+      #   of the state machine. This string could represent different
+      #   values depending on how the context was initialized.
```
p-mongo: This method also raises on `:error`, which I think should be documented in the docstring.

lib/mongo/client_encryption.rb
```diff
@@ -35,7 +35,7 @@ def initialize(client, options = {})
       validate_key_vault_namespace(options[:key_vault_namespace])
 
       @client = client
-      @key_vault_namespace = options[:key_vault_namespace]
+      @key_vault_db, @key_vault_coll = options[:key_vault_namespace].split('.')
```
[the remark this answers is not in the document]
p-mongo: In light of https://jira.mongodb.org/browse/SPEC-1490, perhaps `options[:data_keys_full_collection_name]` is yet another choice worthy of consideration.

lib/mongo/client_encryption.rb
```diff
@@ -58,6 +58,27 @@ def close
...
+    # @return [ BSON::Binary ] UUID representing the data key _id
+    def create_data_key
+      result = nil
+
+      Crypt::DataKeyContext.with_context(@crypt_handle.ref) do |context|
+        result = context.run_state_machine
+      end
+
+      data_key = Hash.from_bson(
+        BSON::ByteBuffer.new.put_bytes(result)
```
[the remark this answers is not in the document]
egiurleo: This is helpful. I didn't quite understand the distinction between buffers for reading and buffers for writing, but that makes complete sense.

lib/mongo/client_encryption.rb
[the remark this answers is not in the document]
egiurleo: I like `options[:data_keys_full_collection_name]`! Thanks for the suggestion.

One question -- this is one of the options specified by the FLE spec (check out the [`ClientEncryptionOpts` type](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#clientencryption)). Do I have to name this option in accordance with the spec or can I deviate if I think a different name might be clearer?

Also, the initializer validates that this string only has one dot in the `validate_key_vault_namespace` method (which I will rename accordingly depending on what name we choose for the option).

lib/mongo/client_encryption.rb
egiurleo: Thanks! I'll write a comment to make that clearer.

lib/mongo/crypt/binding.rb
egiurleo: Sure thing.

lib/mongo/client_encryption.rb
egiurleo: Actually, I'll rename the variable to make it clear that it's a document.

lib/mongo/client_encryption.rb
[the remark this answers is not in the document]
p-mongo: Given that https://docs.mongodb.com/manual/reference/glossary/ defines namespace as :

> The namespace is a combination of the database name and the name of the collection or index, like so: [database-name].[collection-or-index-name]

I think it's fine to keep "key vault namespace" for now, though I made https://jira.mongodb.org/browse/SPEC-1488 to potentially decide on one of "key vault" and "key vault collection".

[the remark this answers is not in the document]
egiurleo: Okay, cool. I'll update depending on the outcome of that issue.

PR #1570, opened 2019-11-07
RUBY-1929 Explicit encryption/decryption POC

egiurleo:
This is a bit of a big PR, so here's how I think you should read it:
1. [Explicit encryption/decryption integration test](https://github.com/mongodb/mongo-ruby-driver/pull/1570/files#diff-bbc4b8927d05ff439eb97bdbc4f1baaaR3)
2. [ClientEncryption changes](https://github.com/mongodb/mongo-ruby-driver/pull/1570/files#diff-b19741740d9c675f399c173d4879229fR22)
3. [Context changes](https://github.com/mongodb/mongo-ruby-driver/pull/1570/files#diff-5835158994eecd1d2cd85fa60d7d5d74R29)
4. [ExplicitEncryptionContext class](https://github.com/mongodb/mongo-ruby-driver/pull/1570/files#diff-c737ecd7efd1c99852369e0a664a5e07R1) and associated [specs](https://github.com/mongodb/mongo-ruby-driver/pull/1570/files#diff-f462b4cafb8e13ad5bc683b4088d5066R1)
5. [ExplicitDecryptionContext class](https://github.com/mongodb/mongo-ruby-driver/pull/1570/files#diff-991b45ef0bf97971d42b7d99b1c8e82eR1) and [associated specs](https://github.com/mongodb/mongo-ruby-driver/pull/1570/files#diff-e8a71240d37f560b9f3056f4bf001ebaR1)
6. Finally, [binding changes](https://github.com/mongodb/mongo-ruby-driver/pull/1570/files#diff-6bd3a05a897f2b96cd467e84165bf42bR148) and [associated specs](https://github.com/mongodb/mongo-ruby-driver/pull/1570/files#diff-21d2bcb251baf83ec283295390086147R8)


spec/integration/explicit_encryption_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+describe 'Explicit Encryption' do
+  require_libmongocrypt
+
+  let(:client) { ClientRegistry.instance.new_local_client(['localhost:27017']) }
+  let(:key_vault_namespace) { 'test.keys' }
+
+  let(:client_encryption_opts) do
+    {
+      kms_providers: {
+        local: { key: Base64.encode64("ru\xfe\x00" * 24) }
```
p-mongo: The spec in https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#kmsproviders specifies that key is a 96 byte string. Why is it base64 encoded here?

spec/integration/explicit_encryption_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    }
+  end
+
+  it 'encrypts and decrypts a value' do
+    client_encryption = Mongo::ClientEncryption.new(
+      client,
+      client_encryption_opts
+    )
+
+    data_key_id = client_encryption.create_data_key
```
p-mongo: create_data_key documents return value thusly:

<pre>
    # @return [ BSON::Binary ] UUID representing the data key _id
</pre>

Is the data key id something a user is expected to store somewhere, such as in an environment variable? If so, how does the user convert the data key id to and from a simple string?

spec/mongo/crypt/explicit_decryption_context_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+end
+
+describe Mongo::Crypt::ExplicitDecryptionContext do
+  require_libmongocrypt
+
+  let(:context) { described_class.new(mongocrypt, io, value) }
+
+  let(:mongocrypt) { Mongo::Crypt::Binding.mongocrypt_new }
+  let(:io) { double("Mongo::ClientEncryption::IO") }
+  let(:value) { BSON::Binary.new("o\x00\x00\x00\x05v\x00b\x00\x00\x00\x06\x01\xDF2~\x89\xD2+N}\x84;i(\xE5\xF4\xBF \x024\xE5\xD2\n\x9E\x97\x9F\xAF\x9D\xC7\xC9\x1A\a\x87z\xAE_;r\xAC\xA9\xF6n\x1D\x0F\xB5\xB1#O\xB7\xCA\xEE$/\xF1\xFA\b\xA7\xEC\xDB\xB6\xD4\xED\xEAMw3+\xBBv\x18\x97\xF9\x99\xD5\x13@\x80y\n{\x19R\xD3\xF0\xA1C\x05\xF7)\x93\x9Bh\x8AA.\xBB\xD3&\xEA\x00") }
```
p-mongo: What is this data? Can it be replaced by code generating it?

spec/integration/explicit_encryption_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    encrypted = client_encryption.encrypt(
+      'Hello, world!',
+      {
+        key_id: data_key_id,
+        algorithm: 'AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic'
```
p-mongo: How does the user choose the algorithm?

spec/integration/explicit_encryption_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      }
+    )
+
+    expect(encrypted).to be_a_kind_of(BSON::Binary)
```
p-mongo: I think the result of encryption should be in a basic Ruby type like a string.

p-mongo:
Points to consider:

- Does data key id contain printable characters only, i.e. can it be stringified and put into environment variable?
- Is there a unambiguous conversion from a string with data key id to BSON::Binary object of the correct type?
- Can the driver API take & return data key id as string?
- Can the spec be changed to allow or mandate data key id being  a string in driver api?
- Is client encryption a lightweight object, i.e.: in  a web app, can it reasonably be created once per each encrypt/decrypt operation? Can it be created once per request? Should it be created once per process and be reused over multiple requests?
- If client encryption ideally should be reused over multiple requests, an api that would permit this easily would be to reference client encryption off a client rather than construct a CE object in code needing it via a constructor. Potentially will need a way to store client encryption options on client which is currently not defined in CSE spec.
- Consider having client encryption object encapsulate both data key id and algorithm, such that encrypt call only takes plain text and no options to simplify the calling code (configure in one place, call in multiple places to encrypt with fewer parameters).
- If one encrypts a Integer and decrypts the result, do they get an Integer back? Need a test case.

spec/integration/explicit_encryption_spec.rb
egiurleo: This is a good question! It's actually in the FLE tutorial on the documentation page (https://docs.mongodb.com/manual/tutorial/manage-client-side-encryption-data-keys/)

> Configuring client-side field level encryption for a locally managed key requires specifying a base64-encoded 96-byte string with no line breaks.

I'm going to make a SPEC ticket for this right now.

egiurleo: spec ticket: https://jira.mongodb.org/browse/SPEC-1501

spec/integration/explicit_encryption_spec.rb
egiurleo: That's a good point. I'm changing the API to accept a data key as a string.

spec/integration/explicit_encryption_spec.rb
egiurleo: Good point.

spec/mongo/crypt/explicit_decryption_context_spec.rb
egiurleo: This is a binary string representing a value that has previously been encrypted by libmongocrypt. I agree that this should be replaced by code generating this value. That'shard to do without actually doing the encryption because libmongocrypt validates the structure of this value, and I'm not sure how to replicate it yet. I'm going to add a comment explaining what this is and then put it on my list of improvements for the future.

egiurleo:
@p-mongo could you take another look?

Changes include:
- passing values around as strings rather than BSON::Binary objects
- adding an integration test case with an integer

Follow ups:
- I will update some specs with code demonstrating how to generate certain values (for example, key id or encrypted binary string) rather than just putting the values in the spec
- I will create a convenient API for ClientEncryption

lib/mongo/client_encryption.rb
```diff
@@ -19,6 +19,33 @@ module Mongo
...
   class ClientEncryption
 
+    # A class that implements I/O methods between the driver and
+    # the MongoDB server or mongocryptd.
+    #
+    # This class should have its own file, just leaving it here
+    # for simplicity.
+    #
+    # @api private
+    class IO
```
p-mongo: I would like to reconsider this class name prior to release, since IO is also a standard library module.

lib/mongo/client_encryption.rb
```diff
@@ -62,7 +90,7 @@ def close
     # that key in the KMS collection. The generated key is encrypted with
     # the KMS master key.
     #
-    # @return [ BSON::Binary ] UUID representing the data key _id
+    # @return [ String ] UUID representing the data key _id
```
p-mongo: Should this be "base64 encoded uuid representing the data key id" ?

lib/mongo/client_encryption.rb
```diff
@@ -71,9 +99,53 @@ def create_data_key
...
+
+      return insert_result.inserted_id.data
+    end
+
+    # Encrypts a value using the specified encryption key and algorithm
+    #
+    # @param [ String|Numeric ] value The value to encrypt
+    # @param [ Hash ] opts
+    #
+    # @option [ String ] :key_id The UUID of the encryption key as
```
p-mongo: base64 encoded?

lib/mongo/client_encryption.rb
```diff
@@ -71,9 +99,53 @@ def create_data_key
...
+    # @return [ String ] The encrypted value
+    #
+    # This method is not currently unit tested.
+    # Find tests in spec/integration/explicit_encryption_spec.rb
+    def encrypt(value, opts={})
+      result = nil
+      value = { 'v': value }.to_bson.to_s
+
+      Crypt::ExplicitEncryptionContext.with_context(@crypt_handle.ref, @io, value, opts) do |context|
+        result = context.run_state_machine
```
p-mongo: `with_context` should return the value of the block, and I believe it already works this way, making explicit assignment to result unnecessary as far as I can tell.

lib/mongo/client_encryption.rb
egiurleo: Yeah that's a great point. I will make a todo to change.

PR #1590, opened 2019-11-26
RUBY-1921: Update Mongo::Client to take options relevant to auto-encryption

egiurleo:
I have updated `Mongo::Client` to take options related to auto-encryption. Because `Client` and `ClientEncryption` take many of the same options, I have extracted the validation of those options, as well as the setting of encryption-related instance variables, into a module called `Mongo::Client::Encrypter`. Similarly, I have extracted auto-encryption-related logic into a module called `Mongo::Client::AutoEncrypter`. Though AutoEncrypter is only used in one class (Client), it has the benefit of removing most of the encryption-related code from the Client class, which is already pretty overloaded and hard to read.

That being said, I'm not married to this way of doing things and I am open to feedback.

This PR *DOES NOT* address:
- updating Client options re: encryption
- passing in encryption options via url

I will make tickets for follow-up work there.


lib/mongo/client.rb
```diff
@@ -53,6 +54,7 @@ class Client
       :auth_mech,
       :auth_mech_properties,
       :auth_source,
+      :auto_encryption_opts,
```
p-mongo: We generally only use "opts" when "options" is already taken. For example methods take `opts` as argument if `options` is an attribute on the class.

lib/mongo/crypt/auto_encrypter.rb
```diff
@@ -0,0 +1,67 @@
...
+    #
+    # @api private
+    module AutoEncrypter
+      include Encrypter
+
+      # A Hash of default values for the :extra_options option
+      DEFAULT_EXTRA_OPTIONS = {
+        mongocryptd_uri: 'mongodb://localhost:27020',
+        mongocryptd_bypass_spawn: false,
+        mongocryptd_spawn_path: '',
```
p-mongo: Why is this an empty string and not nil?

lib/mongo/crypt/auto_encrypter.rb
```diff
@@ -0,0 +1,67 @@
...
+      #   with encrypted fields.
+      # @option [ Boolean | nil ] :bypass_auto_encryption When true, disables
+      #   auto-encryption. Default is false.
+      # @option [ Hash | nil ] :extra_options Options related to spawning
+      #   mongocryptd. These are set to default values if no option is passed in.
+      #
+      # @raise [ ArgumentError ] If required options are missing or incorrectly
+      #   formatted.
+      def setup_encrypter(options = {})
+        extra_options = options.delete(:extra_options) || {}
```
p-mongo: Methods should not mutate arguments, which includes passed options. E.g. call `options.dup` before mutating.

lib/mongo/crypt/auto_encrypter.rb
```diff
@@ -0,0 +1,67 @@
...
+        extra_options = DEFAULT_EXTRA_OPTIONS.merge(extra_options)
+
+        super(
+          options.tap do |options|
+            options[:bypass_auto_encryption] = options[:bypass_auto_encryption] || false
+          end.merge(extra_options)
+        )
+
+        @mongocryptd_client = Client.new(@encryption_options[:mongocryptd_uri])
```
p-mongo: Is this going to create a client with monitoring i/o when the encrypter is made for a client without monitoring i/o?

lib/mongo/crypt/encrypter.rb
```diff
@@ -0,0 +1,74 @@
...
+      #
+      # @raise [ ArgumentError ] If required options are missing or incorrectly
+      #   formatted.
+      def setup_encrypter(options = {})
+        @encryption_options = options
+
+        validate_key_vault_namespace!
+        validate_key_vault_client!
+
+        @encryption_options.freeze
```
p-mongo: This also mutates the argument, use `options.dup.freeze` to avoid that.

lib/mongo/client.rb
```diff
@@ -665,6 +684,7 @@ def write_concern
     # @since 2.1.0
     def close
       @cluster.disconnect!
+      teardown_encrypter
```
egiurleo: Not sure if this is a good way to go about this.

egiurleo: Might it make more sense to create and destroy the mongocryptd_client as needed? I know clients are usually intended to be long-running objects, but perhaps this is a reasonable exception? Open to feedback.

p-mongo: The purpose of close call is to release resources, for example in a process that will no longer use the client. In light of this, it makes sense to me that close destroys crypt bits as well.

If there isn't already, there should be a test added that close-reconnect sequence produces working crypto.

lib/mongo/client.rb
egiurleo: Good to know! Thanks.

lib/mongo/client.rb
egiurleo: That's a good point! This currently performs no crypto (that's the next couple of tickets), but I will make myself a note to add that as a test case.

lib/mongo/crypt/auto_encrypter.rb
egiurleo: That's a good question -- normally, you'd spawn mongocryptd from your system path, which is why I made the default value of this option an empty string, but I think `mongocryptd` would be a better default value.

egiurleo: (My original idea was that you'd probably spawn mongocryptd at a path like `#{mongocryptd_spawn_path}/mongocryptd`, but that's not necessarily a good assumption -- it would be better to let the user specify the entire path.

lib/mongo/crypt/auto_encrypter.rb
egiurleo: You're right. Fixing!

PR #1594, opened 2019-11-27
Fix RUBY-2030 Client/Cluster construction fails when duplicate seed addresses are provided

jonhyman:
We made a configuration change for our connection strings and mistakenly duplicated a mongoS host, which then broke processes when they tried to start up. The issue is that Cluster#add will return nil if the address already exists, and you'll receive this error (taken from the test without the patch).

```
NoMethodError: undefined method `start_monitoring' for nil:NilClass

  0) Mongo::Cluster initialize when there are duplicate addresses does not error out
     Failure/Error:[0mserver.start_monitoring

     NoMethodError:
       undefined method `start_monitoring' for nil:NilClass
     # ./lib/mongo/cluster.rb:188:in `block in initialize'
     # ./lib/mongo/cluster.rb:187:in `each'
```

### DIFF
--- a/spec/mongo/cluster_spec.rb
+++ b/spec/mongo/cluster_spec.rb
@@ -31,8 +31,8 @@ describe Mongo::Cluster do
         described_class.new(addresses, monitoring, SpecConfig.instance.test_options)
       end
 
-      it 'does not error out' do
-        cluster_with_dup_addresses
+      it 'does not raise an exception' do
+        expect { cluster_with_dup_addresses }.not_to raise_error
       end
     end
 
--- a/spec/mongo/cluster_spec.rb
+++ b/spec/mongo/cluster_spec.rb
@@ -20,6 +20,24 @@ describe Mongo::Cluster do
 
   let(:cluster) { cluster_without_io }
 
+  describe 'initialize' do
+
+    context 'when there are duplicate addresses' do
+
+      let(:addresses) do
+        SpecConfig.instance.addresses + SpecConfig.instance.addresses
+      end
+      let(:cluster_with_dup_addresses) do
+        described_class.new(addresses, monitoring, SpecConfig.instance.test_options)
+      end
+
+      it 'does not error out' do
+        cluster_with_dup_addresses
+      end
+    end
+
+  end
+
   describe '#==' do
 
     context 'when the other is a cluster' do
--- a/spec/mongo/cluster_spec.rb
+++ b/spec/mongo/cluster_spec.rb
@@ -31,6 +31,10 @@ describe Mongo::Cluster do
         described_class.new(addresses, monitoring, SpecConfig.instance.test_options)
       end
 
+      after do
+        cluster_with_dup_addresses.disconnect!
+      end
+
       it 'does not raise an exception' do
         expect { cluster_with_dup_addresses }.not_to raise_error
       end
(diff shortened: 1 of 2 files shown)


spec/mongo/cluster_spec.rb
```diff
@@ -20,6 +20,24 @@
...
+      let(:addresses) do
+        SpecConfig.instance.addresses + SpecConfig.instance.addresses
+      end
+      let(:cluster_with_dup_addresses) do
+        described_class.new(addresses, monitoring, SpecConfig.instance.test_options)
+      end
+
+      it 'does not error out' do
+        cluster_with_dup_addresses
+      end
```
egiurleo: Would you mind making this test more specific by:
1. Changing the description to `it 'does not raise an exception'`
2. Updating the test body to specifically test that the code doesn't raise an error
```
expect { cluster_with_dup_address }.not_to raise_error
```

spec/mongo/cluster_spec.rb
```diff
@@ -20,6 +20,24 @@
...
 
+  describe 'initialize' do
+
+    context 'when there are duplicate addresses' do
+
+      let(:addresses) do
+        SpecConfig.instance.addresses + SpecConfig.instance.addresses
+      end
+      let(:cluster_with_dup_addresses) do
+        described_class.new(addresses, monitoring, SpecConfig.instance.test_options)
```
egiurleo: I believe your tests are failing because you're allowing monitoring_io on this cluster. Should be fixed if you merge `monitoring_io: false` with `SpecConfig.instance.test_options`.

spec/mongo/cluster_spec.rb
jonhyman: Thanks, I made these.

spec/mongo/cluster_spec.rb
jonhyman: Hi @egiurleo thank you for the rapid feedback. Which tests are failing? Are they ones outside of this file? When I run the spec file locally I get a 100% pass rate.

Unfortunately if I merge in `monitoring_io: false` into the test options, the test won't fail in the first place without the patch.

egiurleo: Yeah, the tests are outside of this file because background threads are still running even after all the tests have terminated. Of course, as you said, the test will be pointless with `monitoring_io: false` -- thanks for pointing that out.

Could you try disconnecting the cluster you create after the test is run? Something like:

```
after do
  cluster_with_dup_addresses.disconnect!
```

I'll also defer to @p-mongo on whether that's the right solution.

jonhyman: You got it, added.

egiurleo: Thank you! I'll authorize the test run.

p-mongo: The way this should be handled is with `register_cluster` - see e.g. https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/mongo/cluster_spec.rb#L10.

p-mongo:
https://jira.mongodb.org/browse/RUBY-2030

spec/mongo/cluster_spec.rb
jonhyman: I updated to do that.

PR #1596, opened 2019-11-29
Suppress the warning "warning: URI.unescape is obsolete".

junaruga:
Hello.

URI.decode (alias of URI.unescape) is obsolete in latest Ruby 2.5, 2.6. coming 2.7.

Following the way to test with standalone mongod: https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/README.md#standalone , you see many warnings `URI.unescape is obsolete` in Ruby 2.6.5 and Ruby 2.7.0 preview3 in the process of `rake`.

```
$ bundle exec rake
...
/home/jaruga/git/mongo-ruby-driver/lib/mongo/uri.rb:457: warning: URI.unescape is obsolete
```

References: 
* https://ruby-doc.org/stdlib-2.6.5/libdoc/uri/rdoc/URI/Escape.html#method-i-unescape
* https://ruby-doc.org/stdlib-2.5.7/libdoc/uri/rdoc/URI/Escape.html#method-i-unescape

I referred this page: https://github.com/hanami/utils/pull/130 .

### DIFF
--- a/lib/mongo/uri.rb
+++ b/lib/mongo/uri.rb
@@ -454,7 +454,7 @@ module Mongo
     end
 
     def decode(value)
-      ::URI.decode(value)
+      ::URI.decode_www_form_component(value)
     end
 
     def encode(value)


p-mongo:
These two methods do not have the same behavior:

<pre>
irb(main):001:0> URI.decode('a+b')
=> "a+b"
irb(main):002:0> URI.decode_www_form_component('a+b')
=> "a b"
</pre>

https://tools.ietf.org/html/rfc3986 does not mention anything about "form components". The encode_www_form_component documentation in https://ruby-doc.org/stdlib-2.6.5/libdoc/uri/rdoc/URI.html#method-c-encode_www_form_component links to https://www.w3.org/TR/2013/CR-html5-20130806/forms.html#url-encoded-form-data which is the HTML5 standard, not RFC 3986. As I understand it our connection strings are mostly RFC 3986 compliant (although not completely), but I do not see anything in our connection string spec (https://github.com/mongodb/specifications/tree/master/source/connection-string) referencing form components or form data.

Based on this, I believe it would be incorrect for the driver to use `URI.decode_www_form_component` as proposed in this PR.

It is unfortunate that Ruby does not appear to have a standard library method to perform URI decoding (as per any URI RFC) that is free of warnings. I would consider this a bug in the Ruby standard library and report it to the Ruby issue tracker.

A possible solution would be for the Ruby driver to implement URL decoding internally which would be Ruby warning-free, however it appears that the connection spec does not currently define which URL decoding specification to use. I made a note to bring this up as part of our specification process discussion.

In the mean time, I think continuing with `URI.decode` is probably the most sensible approach. It is commonly used in virtually all Ruby applications and I cannot imagine it will be removed from the standard library without an adequate replacement (which the currently suggested alternatives - `URI.decode_www_form_component` and `CGI.decode` - are not).

Lastly, with respect to:

> Following the way to test with standalone mongod: https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/README.md#standalone , you see many warnings URI.unescape is obsolete in Ruby 2.6.5 and Ruby 2.7.0 preview3 in the process of rake.

These warnings are only produced if `$VERBOSE` is set which is not the default Ruby configuration.

We already have a ticket in the backlog to make the driver warning-free (https://jira.mongodb.org/browse/RUBY-1775). I added URI.decode and this PR there as a link.

junaruga:
Hmm, I see. Thank you for checking.

```
irb(main):001:0> require 'uri'
=> false

irb(main):002:0> URI.decode('a+b')
(irb):2: warning: URI.unescape is obsolete
=> "a+b"

irb(main):003:0> URI.decode_www_form('a+b')
=> [["a b", ""]]

irb(main):004:0> URI.decode_www_form_component('a+b')
=> "a b"

irb(main):005:0> require 'cgi'
=> true

irb(main):006:0> CGI.unescape('a+b')
=> "a b"
```

> I would consider this a bug in the Ruby standard library and report it to the Ruby issue tracker.

Thank you for that!

> These warnings are only produced if $VERBOSE is set which is not the default Ruby configuration.
> 
> We already have a ticket in the backlog to make the driver warning-free (https://jira.mongodb.org/browse/RUBY-1775). I added URI.decode and this PR there as a link.

I checked the ticket now.
It seems that the unset `$VERBOSE` better than current situation for `rake`.

Feel free to close this PR, if you like.
Thank you.

junaruga:
> These warnings are only produced if $VERBOSE is set which is not the default Ruby configuration.

Are you sure about it? It seems the warnings are produced even if $VERBOSE is not set on Ruby 2.7.0 preview3.

```
$ which irb
/usr/local/ruby-2.7.0-preview3/bin/irb

$ irb

irb(main):001:0> require 'uri'
=> false

irb(main):002:0> p $VERBOSE
false
=> false

irb(main):003:0> URI.decode('a+b')
(irb):3: warning: URI.unescape is obsolete
=> "a+b"
```

```
$ which ruby
/usr/local/ruby-2.7.0-preview3/bin/ruby

$ ruby -e "puts $VERBOSE; puts URI.decode('a+b')"

-e:1: warning: URI.unescape is obsolete
a+b
```

p-mongo:
You are right, I see this warning in our ruby-head builds but we need to solve this with a warning-free URI decoder rather than an HTML form decoder.

p-mongo:
This has been addressed by https://github.com/mongodb/mongo-ruby-driver/pull/1614.

PR #1597, opened 2019-12-02
RUBY-1923 Spawn mongocryptd on client startup

egiurleo:
If the client has `auto_encryption_options`, it should spawn mongocryptd on startup (unless the `mongocryptd_bypass_spawn` argument is true or the client is not monitoring i/o). On close, the client will terminate the mongocryptd process if it is still running.


lib/mongo/crypt/auto_encrypter.rb
```diff
@@ -66,14 +72,52 @@ def setup_encrypter(options = {})
                                 monitoring_io: mongocryptd_client_monitoring_io,
                               )
 
+        unless @encryption_options[:mongocryptd_bypass_spawn] || !mongocryptd_client_monitoring_io
```
egiurleo: I think it makes sense to not spawn anything if `monitoring_io` is false on the client. Happy to discuss if you disagree.

lib/mongo/crypt/auto_encrypter.rb
```diff
@@ -66,14 +72,52 @@ def setup_encrypter(options = {})
...
+
+      # Abort the spawned mongocryptd process. Will not raise an error,
+      # even if the process has already been terminated.
+      #
+      # @return [ true ] Always true
+      def kill_mongocryptd
+        return true if @mongocryptd_pid.nil?
+
+        begin
+          Process.kill('ABRT', @mongocryptd_pid)
```
egiurleo: I don't know much about process kill signals. Abort seemed like the appropriate signal to use in this case, but I'm open to guidance.

lib/mongo/crypt/auto_encrypter.rb
```diff
@@ -66,14 +72,52 @@ def setup_encrypter(options = {})
...
+          Process.wait(@mongocryptd_pid)
+        rescue Errno::ESRCH
+         log_warn("There was no mongocrypd process running at #{@mongocryptd_pid}.")
```
p-mongo: Who is responsible for managing the pid file, in particular deleting it when the process finishes? Is it always LMCD?

lib/mongo/crypt/auto_encrypter.rb
```diff
@@ -66,14 +72,52 @@ def setup_encrypter(options = {})
...
+
+        true
+      end
+
       # Close the resources created by the AutoEncrypter
       #
       # @return [ true ] Always true
       def teardown_encrypter
         @mongocryptd_client.close if @mongocryptd_client
+        kill_mongocryptd
```
p-mongo: I thought the daemon process would live forever and just the client would be closed here.

lib/mongo/crypt/auto_encrypter.rb
```diff
@@ -66,14 +72,52 @@ def setup_encrypter(options = {})
...
+      # Spawn a new mongocryptd process using the mongocryptd_spawn_path
+      # and mongocryptd_spawn_args passed in through the extra auto
+      # encrypt options. Stdout and Stderr of this new process are written
+      # to /dev/null.
+      #
+      # @return [ Integer ] The process id of the spawned process
+      def spawn_mongocryptd
+        @mongocryptd_pid = Process.spawn(
+                            @encryption_options[:mongocryptd_spawn_path],
+                            *@encryption_options[:mongocryptd_spawn_args],
```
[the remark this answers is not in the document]
p-mongo: Can a test be added that specifies empty args and `echo hello world` as the path? The expected behavior is `echo hello world` would be attempted to be launched as a binary and fail.

lib/mongo/crypt/auto_encrypter.rb
```diff
@@ -66,14 +72,52 @@ def setup_encrypter(options = {})
...
+          self.spawn_mongocryptd
+        end
+
         # TODO: use all the other options for auto-encryption/auto-decryption
       end
 
+      # Spawn a new mongocryptd process using the mongocryptd_spawn_path
+      # and mongocryptd_spawn_args passed in through the extra auto
+      # encrypt options. Stdout and Stderr of this new process are written
+      # to /dev/null.
```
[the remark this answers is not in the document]
egiurleo: From the spec:

> The stdout and stderr of the spawned process MUST not be exposed in the driver (e.g. redirect to /dev/null). Users can pass the argument --logpath to extraOptions.mongocryptdSpawnArgs if they need to inspect mongocryptd logs.

I think the spec considers mongocryptd an implementation detail that should never be exposed to the user, so the output is permanently discarded unless the user specifies a logpath. I can add a note about the logpath.

lib/mongo/crypt/auto_encrypter.rb
egiurleo: You're completely right. From the spec:
> The mongocryptd process is responsible for self terminating after idling for a time period.

I'll remove everything having to do with terminating it.

lib/mongo/crypt/auto_encrypter.rb
egiurleo: What is LMCD?

lib/mongo/crypt/auto_encrypter.rb
```diff
@@ -54,21 +54,56 @@ def setup_encrypter(options = {})
...
+      #
+      # @note To capture the mongocryptd logs, add "--logpath=/path/to/logs"
+      #   to auto_encryption_options -> extra_options -> mongocrpytd_spawn_args
+      #
+      # @return [ Integer ] The process id of the spawned process
+      def spawn_mongocryptd
+        mongocryptd_spawn_args = @encryption_options[:mongocryptd_spawn_args]
+
+        if  mongocryptd_spawn_args.nil? || mongocryptd_spawn_args.empty?
+          mongocryptd_spawn_args = ['--idleShutdownTimeoutSecs=0']
```
egiurleo: @p-mongo -- Thank you for your comment regarding `Process.spawn` executing any command if the list of arguments is empty. I came up with two potential ways of solving this problem:

1. Always pass an argument to the mongocryptd process. By default, mongocryptd sets the `idleShutdownTimeoutSecs` argument to 0, so there is no harm (currently) in passing that in as a default. If the user passes in a path that isn't an executable, `Process.spawn` will error. Of course, this could break something if server team changes how mongocryptd works in later versions of the server.

2. (Try to) validate that the spawn path is the path to an executable. I wrote a method below called `validate_extra_options!`, which checks that the path is an executable file or that it is referencing an executable in the user's PATH variable. There is a similar method already written in the mkmf module ([reference here](https://ruby-doc.org/stdlib-2.0.0/libdoc/mkmf/rdoc/MakeMakefile.html#method-i-find_executable)) but that method writes to stdout, and it would require adding another dependency to the project.

Let me know what you think!

spec/integration/client_auto_encryption_options_spec.rb
```diff
@@ -129,32 +136,94 @@
...
+      end
     end
+  end
 
-    it 'sets bypass_auto_encryption to false' do
-      expect(client.encryption_options[:bypass_auto_encryption]).to be false
+  describe 'Spawning mongocryptd' do
+    require_enterprise
+
+    let(:client) do
```
egiurleo: @p-mongo It seems like this client is not getting cleaned up by `close_local_clients` and it is causing errors -- am I doing something wrong here when creating the new local client?

lib/mongo/crypt/auto_encrypter.rb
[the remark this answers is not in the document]
p-mongo: :+1:

lib/mongo/crypt/auto_encrypter.rb
p-mongo: Will follow up as https://jira.mongodb.org/browse/RUBY-2034.

p-mongo:
Since the mcd is supposed to be spawned every time server selection to it fails, there should eventually be an integration test that:

- performs a query/operation requiring mcd
- ensures success
- kills running mcd
- ensures there is no running mcd
- performs the operation again
- ensures success
- ensures there is a running mcd

PR #1600, opened 2019-12-03
RUBY-1551 Uninformative error message when auth source is empty


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-1551

PR #1604, opened 2019-12-14
Synchronize client connection/disconnection to avoid leaking background threads


p-mongo:


p-mongo:
This PR should also fix the map/reduce failures on latest server.

PR #1606, opened 2019-12-15
RUBY-1849 refactor cursor logic for clarity

p-mongo:


p-mongo:
This PR seems to trigger connection reconnect lint failures rather consistently, but they also happen (at a significantly lower frequency) on master.

PR #1607, opened 2019-12-18
RUBY-2014 Send logging messages from libmongocrypt to Mongo::Logger

egiurleo:
Also change the context api (data key context, explicit encryption context, explicit decryption context) to be more consistent with the rest of the `Crypt` API that's been built so far.

### DIFF
--- a/lib/mongo/crypt/handle.rb
+++ b/lib/mongo/crypt/handle.rb
@@ -52,9 +52,9 @@ module Mongo
 
       private
 
-      # TODO: documentation
+      # Send the logs from libmongocrypt to the Mongo::Logger
       def set_logger
-        @log_callback = Proc.new do |level, msg, len=nil, ctx=nil|
+        @log_callback = Proc.new do |level, msg|
           logger = Mongo::Logger.logger
 
           case level
--- a/lib/mongo/crypt/handle.rb
+++ b/lib/mongo/crypt/handle.rb
@@ -28,12 +28,16 @@ module Mongo
       # is currently :local. Local KMS options must be passed in the format
       # { local: { key: <master key> } } where the master key is a 96-byte, base64
       # encoded string.
+      # @param [ Hash ] options A hash of options
+      #
+      # @option [ Logger ] :logger A Logger object to which libmongocrypt logs
+      #   will be sent
       #
       # There will be more arguemnts to this method once automatic encryption is introduced.
-      def initialize(kms_providers, logger=nil)
+      def initialize(kms_providers, options={})
         # FFI::AutoPointer uses a custom release strategy to automatically free
         # the pointer once this object goes out of scope
-        @logger = logger
+        @logger = options[:logger]
         @mongocrypt = FFI::AutoPointer.new(
           Binding.mongocrypt_new,
           Binding.method(:mongocrypt_destroy)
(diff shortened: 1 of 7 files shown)


spec/mongo/crypt/explicit_encryption_context_spec.rb
```diff
@@ -79,6 +78,40 @@
...
+          # option; none of the pre-built packages on Evergreen have been built with logging enabled.
+          #
+          # It is still useful to be able to run these tests locally to confirm that logging is working
+          # while debugging any problems.
+          #
+          # For now, skip this test by default and revisit once we have determined how we want to
+          # package libmongocrypt with the Ruby driver.
+          skip "These tests require libmongocrypt to be built with the '-DENABLE_TRACE=ON' cmake option." +
+            " They also require the MONGOCRYPT_TRACE environment variable to be set to 'ON'."
+        end
```
egiurleo: @p-mongo let me know if you're okay leaving the tests like this for now.

lib/mongo/crypt/handle.rb
```diff
@@ -52,6 +52,31 @@ def ref
...
+          when :fatal
+            logger.fatal(msg)
+          when :error
+            logger.error(msg)
+          when :warning
+            logger.warn(msg)
+          when :info
+            logger.info(msg)
+          when :trace
+            logger.debug(msg)
```
p-mongo: Would it make sense to have the level symbols match the Ruby logger methods, to simplify the mapping?

lib/mongo/crypt/handle.rb
```diff
@@ -52,6 +52,31 @@ def ref
 
       private
 
+      # Send the logs from libmongocrypt to the Mongo::Logger
+      def set_logger
+        @log_callback = Proc.new do |level, msg|
+          logger = Mongo::Logger.logger
```
p-mongo: This takes the global logger, not the client's logger for any particular client. This is an existing issue in the driver affecting a number of call sites, but is the client's logger conveniently available here? Or, is the logger here intended to be the global one?

spec/mongo/crypt/explicit_encryption_context_spec.rb
p-mongo: I think it's fine, maybe link to the jira ticket for packaging if you are going to make other changes.

lib/mongo/crypt/handle.rb
egiurleo: Nope, you're right, I meant it to be the client's logger. I'll fix it. Thanks for pointing that out.

lib/mongo/crypt/handle.rb
egiurleo: That's awesome, thank you.

lib/mongo/crypt/handle.rb
```diff
@@ -22,23 +22,28 @@ module Crypt
...
+      #
+      # @option [ Logger ] :logger A Logger object to which libmongocrypt logs
+      #   will be sent
       #
       # There will be more arguemnts to this method once automatic encryption is introduced.
-      def initialize(kms_providers)
+      def initialize(kms_providers, options={})
         # FFI::AutoPointer uses a custom release strategy to automatically free
         # the pointer once this object goes out of scope
+        @logger = options[:logger]
```
egiurleo: @p-mongo I changed the API to accept a logger as an option, with the intention of passing it down from the Client or ClientEncryption object. Does that seem reasonable?

PR #1609, opened 2019-12-20
Fix RUBY-2055 Driver sends null pwd field in createUser when password is not specified

p-mongo:
https://jira.mongodb.com/browse/RUBY-2055

### DIFF
--- a/docs/tutorials/user-management.txt
+++ b/docs/tutorials/user-management.txt
@@ -38,6 +38,31 @@ For more information about users and user management, see MongoDB's
 :manual:`online documentation </core/security-users>`.
 
 
+Users and Databases
+-------------------
+
+When a client connects to the server, MongoDB distinguishes the database
+that the client will perform operations on from the `auth source <auth-source>`_
+which is the database storing the user that the client is authenticating as.
+
+In many cases, the auth source is the same as the database. When they differ,
+user management operations must be done on the auth source database. For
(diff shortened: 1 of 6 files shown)


docs/tutorials/user-management.txt
```diff
@@ -34,12 +37,39 @@ In this example, all operations would be performed on the ``users`` database.
...
+
+  client.use('$external').database.users.create(
+    'C=US,ST=New York,L=New York City,O=MongoDB,OU=x509,CN=localhost',
+    roles: [{role: 'read', db: 'admin'}],
+  )
+
+Note that the auth source is not specified for creating the user - auth source
+is only used during the authentication process. If ``#create`` is invoked with
+a ``User`` object with ``auth_source`` set, the auth source is ignored for
+the purposes of user management.
```
egiurleo: Could you mention that the user's auth source will be set to a default value depending on what mechanism you use to create the user?

p-mongo: I suppose the language could be made more clear but my current understanding is that "auth source" is something that only exists during authentication. When performing user management there is no concept of auth source at all.

p-mongo: So, to respond to your comment:

> the user's auth source will be set to a default value

Set *where* exactly?

egiurleo: Having looked at the code a bit more, I agree with you. Thanks for pointing that out. Merge away.

PR #1610, opened 2019-12-20
Fix RUBY-2054 upsert option is not sent to legacy servers with w:0 due to string/symbol key mismatch


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2054

spec/integration/crud_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    collection.delete_many
+  end
+
+  describe 'upsert' do
+    context 'with default write concern' do
+      it 'upserts' do
+        collection.count_documents({}).should == 0
+
+        res = collection.find(_id: 'foo').update_one({'$set' => {foo: 'bar'}}, upsert: true)
+        p res
```
egiurleo: Did you mean to keep this in here?

p-mongo: No, thank you for catching that.

PR #1611, opened 2019-12-20
RUBY-1924: Schema Map

egiurleo:
Add the ability for a libmongocrypt Handle to accept a schema map, which specifies the fields to be encrypted/decrypted during auto encryption.

### DIFF
--- a/lib/mongo/crypt/handle.rb
+++ b/lib/mongo/crypt/handle.rb
@@ -25,24 +25,28 @@ module Mongo
       # Creates a new Handle object and initializes it with options
       #
       # @param [ Hash ] kms_providers A hash of KMS settings. The only supported key
-      # is currently :local. Local KMS options must be passed in the format
-      # { local: { key: <master key> } } where the master key is a 96-byte, base64
-      # encoded string.
+      #   is currently :local. Local KMS options must be passed in the format
+      #   { local: { key: <master key> } } where the master key is a 96-byte, base64
+      #   encoded string.
+      # @param [ Hash | nil ] schema_map A hash representing the JSON schema of the collection
+      #   that stores auto encrypted documents.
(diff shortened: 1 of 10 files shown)


lib/mongo/crypt/handle.rb
```diff
@@ -25,24 +25,28 @@ class Handle
...
+      # @param [ Hash | nil ] schema_map A hash representing the JSON schema of the collection
+      #   that stores auto encrypted documents.
       # @param [ Hash ] options A hash of options
       #
       # @option [ Logger ] :logger A Logger object to which libmongocrypt logs
       #   will be sent
       #
       # There will be more arguemnts to this method once automatic encryption is introduced.
-      def initialize(kms_providers, options={})
+      def initialize(kms_providers, schema_map, options={})
```
p-mongo: Passing nil to schema_map everywhere seems awkward, maybe default it to nil or use a keyword argument defaulted to nil?

lib/mongo/crypt/handle.rb
```diff
@@ -25,24 +25,28 @@ class Handle
...
+      # @param [ Hash | nil ] schema_map A hash representing the JSON schema of the collection
+      #   that stores auto encrypted documents.
       # @param [ Hash ] options A hash of options
       #
       # @option [ Logger ] :logger A Logger object to which libmongocrypt logs
       #   will be sent
       #
       # There will be more arguemnts to this method once automatic encryption is introduced.
-      def initialize(kms_providers, options={})
+      def initialize(kms_providers, schema_map: nil, options: {})
```
egiurleo: @p-mongo Thank you for the feedback. Is this what you were asking for?

p-mongo: Yes this looks good!

PR #1616, opened 2019-12-30
RUBY-2028 Allow user to update auto_encryption_options using Client#update_options

egiurleo:
I realized that the way I was creating the key vault client meant that it was trying to auto encrypt, which it should not do. I also updated the encrypter to keep track of a key vault client that it spawns and close it on teardown.

### DIFF
--- a/lib/mongo/client.rb
+++ b/lib/mongo/client.rb
@@ -861,7 +861,9 @@ module Mongo
       opts_copy = @options[:auto_encryption_options].dup
 
       opts_copy.tap do |opts|
-        opts[:key_vault_client] ||= self
+        opts[:key_vault_client] ||= self.dup
+        opts[:key_vault_client].update_options({ auto_encryption_options: nil })
+
         opts[:extra_options] ||= {}
 
         opts[:extra_options][:mongocryptd_client_monitoring_io] = self.options[:monitoring_io]
--- a/lib/mongo/client.rb
+++ b/lib/mongo/client.rb
@@ -121,6 +121,9 @@ module Mongo
     # @return [ Hash ] options The configuration options.
     attr_reader :options
 
+    # TODO: documentation
+    attr_reader :addresses_or_uri
+
     # Delegate command and collections execution to the current database.
     def_delegators :@database, :command, :collections
 
@@ -384,6 +387,8 @@ module Mongo
     # @since 2.0.0
     def initialize(addresses_or_uri, options = nil)
       options = options ? options.dup : {}
+
+      @addresses_or_uri = addresses_or_uri
 
       srv_uri = nil
       if addresses_or_uri.is_a?(::String)
--- a/lib/mongo/client.rb
+++ b/lib/mongo/client.rb
@@ -446,16 +446,7 @@ module Mongo
       remove_instance_variable('@monitoring')
 
       if @options[:auto_encryption_options]
-        opts_copy = @options[:auto_encryption_options].dup
-
-        opts_copy.tap do |opts|
-          opts[:key_vault_client] ||= self
-          opts[:extra_options] ||= {}
-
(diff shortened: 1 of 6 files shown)


lib/mongo/client.rb
```diff
@@ -657,6 +648,14 @@ def update_options(new_options)
         @options = options.freeze
         validate_options!
         validate_authentication_options!
+
+        if @options.key?(:auto_encryption_options) && @options[:auto_encryption_options].nil?
```
p-mongo: Can a comment be added explaining why the teardown happens when options are nil only?

lib/mongo/client.rb
```diff
@@ -856,6 +855,19 @@ def default_options(options)
       end
     end
 
+    # Provides some default encryption options and sets up data necessary
+    # for auto-encryption
+    def set_auto_encryption_options
+      opts_copy = @options[:auto_encryption_options].dup
+
+      opts_copy.tap do |opts|
```
p-mongo: The tap here is no longer needed, it can be removed and the two nested lines deindented.

lib/mongo/crypt/auto_encrypter.rb
```diff
@@ -53,6 +53,17 @@ module AutoEncrypter
       def setup_encrypter(options = {})
         opts = set_default_options(options.dup)
 
+        unless opts[:key_vault_client]
+          # If no key vault client is passed in, create one by duplicating the
+          # Mongo::Client used for encryption. Update options so that key vault
+          # client does not perform auto-encryption/decryption, and keep a reference
+          # to it so it is destroyed later.
+          @key_vault_client = self.dup
```
p-mongo: Is this dup immune to the issue described in https://jira.mongodb.org/browse/RUBY-1506?jql=project%20%3D%20RUBY%20AND%20status%20%3D%20Open%20AND%20text%20~%20dup ?

p-mongo: Does `#with` work to change `auto_encryption_options` to nil?

lib/mongo/client.rb
egiurleo: Sure!

lib/mongo/crypt/auto_encrypter.rb
egiurleo: RE: your first comment -- you are very right. I didn't know about that. Thank you so much for catching it.

egiurleo: I'm adding a test that makes sure key_vault_client is a functional client and fixing this now 👍

egiurleo:
@p-mongo I updated this PR:
1. Test that the auto encryption options are updated using `#with`
2. Don't duplicate the client for key_vault_client, use `with` instead.
3. Test that the default key_vault_client is functional

lib/mongo/crypt/auto_encrypter.rb
```diff
@@ -53,6 +53,16 @@ module AutoEncrypter
       def setup_encrypter(options = {})
         opts = set_default_options(options.dup)
 
+        unless opts[:key_vault_client]
+          # If no key vault client is passed in, create one by copying the
+          # Mongo::Client used for encryption. Update options so that key vault
+          # client does not perform auto-encryption/decryption, and keep a reference
+          # to it so it is destroyed later.
+          @key_vault_client = self.with({ auto_encryption_options: nil })
```
p-mongo: Could you also add a test to client_construction_spec.rb which illustrates whether changing AEO creates a new cluster.

p-mongo: And whether it is possible to use the same client instance to access key vault, since auto encryption should be completely governed by the schema and the key vault collection is separate from user collections (?) such that the schema should not specify encryption for the key vault collection.

PR #1620, opened 2020-01-02
RUBY-2065 Reset autoload config when Crypt::Binding fails to load

p-mongo:


lib/mongo/crypt/binding.rb
```diff
@@ -12,28 +12,36 @@
...
-
       begin
         ffi_lib ENV['LIBMONGOCRYPT_PATH']
       rescue LoadError => e
-        raise "Cannot load Mongo::Crypt::Binding because the path to " +
+        Crypt.reset_autoload
+        raise LoadError, "Cannot load Mongo::Crypt::Binding because the path to " +
           "libmongocrypt specified in the LIBMONGOCRYPT_PATH environment variable " +
-          "is invalid: #{ENV['LIBMONGOCRYPT']}\n\n#{e.class}: #{e.message}"
+          "is invalid: #{ENV['LIBMONGOCRYPT_PATH']}\n\n#{e.class}: #{e.message}"
```
egiurleo: Thanks for catching that 🙏

PR #1621, opened 2020-01-03
RUBY-2060 Pin mlaunch version on evergreen

egiurleo:


spec/integration/client_authentication_options_spec.rb
```diff
@@ -38,7 +38,7 @@
     end
   end
 
-  shared_examples_for 'auth mechanism that uses database or default auth source' do |default_auth_source:|
```
p-mongo: Would `do |default_auth_source: nil|` fix this also?

egiurleo: I was going to try that next. Let's see!

egiurleo: I tried that out, and rspec seems to think that I'm not passing in any value for default_auth_source. I know that keyword arguments are changing in Ruby 2.7/3, but I believe the existing syntax should still work, so I'm not sure how that's related.

How would you feel about just removing the keyword argument for now?

PR #1623, opened 2020-01-07
RUBY-1939 Bind Ruby crypto hooks to libmongocrypt

egiurleo:
Crypto functionality now works with a nocrypto build of libmongocrypt!

I have also made some API changes:
- Make the entire crypto API private except for the things designated as public by the client-side encryption spec
- Change the `Binary` and `Status` APIs to allow those objects to be initialized using an existing pointer to an underlying C object (for use in crypto hooks)
- Change the `Handle` API to be more consistent (make every argument that is optional part of `options` rather than its own argument)

### DIFF
--- a/lib/mongo/crypt/binary.rb
+++ b/lib/mongo/crypt/binary.rb
@@ -20,12 +20,14 @@ module Mongo
     # A wrapper around mongocrypt_binary_t, a non-owning buffer of
     # uint-8 byte data. Each Binary instance keeps a copy of the data
     # passed to it in order to keep that data alive.
+    #
+    # @api private
     class Binary
       # Create a new Binary object that wraps a byte string
       #
       # @param [ String ] data The data string wrapped by the
       #   byte buffer (optional)
-      def initialize(data=nil)
+      def initialize(data: nil, pointer: nil)
         if data
           # Represent data string as array of uint-8 bytes
           bytes = data.unpack('C*')
@@ -40,6 +42,11 @@ module Mongo
             Binding.mongocrypt_binary_new_from_data(@data_p, bytes.length),
             Binding.method(:mongocrypt_binary_destroy)
           )
+        elsif pointer
+          # If the Binary class is used this way, it means that the pointer
+          # for the underlying mongocrypt_binary_t object is allocated somewhere
+          # else. The Binary object is not responsible for deallocating data.
+          @bin = pointer
         else
           # FFI::AutoPointer uses a custom release strategy to automatically free
           # the pointer once this object goes out of scope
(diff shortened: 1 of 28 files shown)


spec/support/utils.rb
```diff
@@ -1,4 +1,49 @@
...
+
+    if doc.is_a?(Array)
+      return doc.each.with_index do |el, idx|
+        doc[idx] = parse_extended_json(el)
+      end
+    end
+
+    return doc
+  end
+  module_function :parse_extended_json
```
egiurleo: It would be really helpful for me to have a limited subset of extended JSON parsing to load data keys/schema maps for the FLE spec. Let me know if you think I should approach this some other way.

lib/mongo/crypt/binary.rb
```diff
@@ -50,6 +63,60 @@ def initialize(data=nil)
...
+      # original memory allocated to the underlying mongocrypt_binary_t
+      # object. Do NOT use this method unless required to do so by libmongocrypt.
+      #
+      # @param [ String ] data The new string data to be wrapped by this binary object
+      #
+      # @return [ true | false ] Whether the operation was successful
+      def write(data)
+        # Cannot write a string that's longer than the space currently allocated
+        # by the mongocrypt_binary_t object
+        return false if Binding.mongocrypt_binary_len(@bin) < data.length
```
p-mongo: How come this does not raise an exception?

lib/mongo/crypt/binary.rb
```diff
@@ -50,6 +63,60 @@ def initialize(data=nil)
...
+          @data_p.clear if @data_p
+
+          # # FFI::MemoryPointer automatically frees memory when it goes out of scope
+          @data_p = FFI::MemoryPointer.new(bytes.length)
+                      .write_array_of_uint8(bytes)
+
+          Binding.mongocrypt_binary_data(@bin)
+                    .write_array_of_uint8(bytes)
+        rescue FFI::NullPointerError
+          return false
```
p-mongo: When would this case happen?

egiurleo: If you initialize a mongocrypt_binary_t object with no data and then try to write data to it. I'll write a comment.

lib/mongo/crypt/binary.rb
egiurleo: Good question. That would actually make more sense. I'll change it to do that right now.

lib/mongo/crypt/binary.rb
egiurleo: I've realized this isn't necessary because of the conditional above it. I'll remove this case.

p-mongo: Sounds good - I would think that it is a bug in the driver to make libmongocrypt reference null pointers (for practical reasons like it interferes with debugging, even if LMC could/should provide better diagnostics in such cases).

PR #1624, opened 2020-01-09
RUBY-1934 Auto encryption and decryption unit tests

egiurleo:
This PR implements the methods necessary for auto encryption/decryption and tests those methods.

Upcoming PRs will actually integrate those methods into the required database commands.

### DIFF
--- a/lib/mongo/crypt/auto_encrypter.rb
+++ b/lib/mongo/crypt/auto_encrypter.rb
@@ -71,6 +71,7 @@ module Mongo
         @mongocryptd_client = Client.new(
                                 @encryption_options[:mongocryptd_uri],
                                 monitoring_io: mongocryptd_client_monitoring_io,
+                                server_selection_timeout: 3,
                               )
 
         @encryption_io = EncryptionIO.new(
--- a/lib/mongo/crypt/auto_encrypter.rb
+++ b/lib/mongo/crypt/auto_encrypter.rb
@@ -73,9 +73,11 @@ module Mongo
                                 monitoring_io: mongocryptd_client_monitoring_io,
                               )
 
-        # @encryption_io = EncryptionIO.new(options[:key_vault_client], options[:key_vault_namespace])
-
-        # TODO: use all the other options for auto-encryption/auto-decryption
+        @encryption_io = EncryptionIO.new(
+                          client: self,
+                          mongocryptd_client: @mongocryptd_client,
+                          key_vault_collection: build_key_vault_collection
+                         )
       end
 
       # Spawn a new mongocryptd process using the mongocryptd_spawn_path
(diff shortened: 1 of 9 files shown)


lib/mongo/crypt/encryption_io.rb
```diff
@@ -51,7 +54,37 @@ def find_keys(filter)
...
+      # Get collection info for a collection matching the provided filter
+      #
+      # @param [ Hash ] filter
+      #
+      # @return [ Hash ] The collection information
+      def collection_info(filter)
+        result = @client.database.list_collections
+
+        name = filter['name']
+        result.find { |r| r['name'] == name }
```
egiurleo: @p-mongo I was trying to do something like Python's list_collections(filter) method -- https://api.mongodb.com/python/current/api/pymongo/database.html#pymongo.database.Database.list_collections. What do you think is the best approach?

lib/mongo/crypt/context.rb
```diff
@@ -106,6 +114,11 @@ def raise_from_status
         status.raise_crypt_error
       end
 
+      # Indicate that state machine is done feeding I/O responses back to libmongocrypt
+      def mongo_done
```
p-mongo: `mongo_xxx` method names to me imply that they are interacting with the server. If they are interacting with mongocryptd/libmongocryptd can the prefix be changed to `mongocrypt`, `lmc`, `mc`, etc.?

lib/mongo/crypt/encryption_io.rb
p-mongo: The server command supports a filter argument (https://docs.mongodb.com/manual/reference/command/listCollections/). Our helper doesn't apparently (https://api.mongodb.com/ruby/current/Mongo/Database.html#collections-instance_method). Amending the helper to pass the filter seems like a valid strategy, but if the inline filtering here works it seems fine to do this as well.

lib/mongo/crypt/auto_encrypter.rb
```diff
@@ -71,9 +73,52 @@ def setup_encrypter(options = {})
...
                                 @encryption_options[:mongocryptd_uri],
                                 monitoring_io: mongocryptd_client_monitoring_io,
+                                server_selection_timeout: mongocryptd_server_selection_timeout, # For testing purposes
                               )
 
-        # TODO: use all the other options for auto-encryption/auto-decryption
+        # Tests fail with live background threads if the @mongocryptd client is not closed at the
+        # end of every test run. Better to write one line registering it as a local client than to
+        # do it separately in every test.
+        ClientRegistry.instance.register_local_client(@mongocryptd_client) if defined?(ClientRegistry)
```
p-mongo: Can you please make a ticket to investigate why this is needed and preferably remove it before release?

egiurleo: https://jira.mongodb.org/browse/RUBY-2069

PR #1627, opened 2020-01-09
RUBY-1924 use auto encryption in insert_one and find commands [WIP]

egiurleo:
This is a first attempt at adding auto encryption to insert_one and find. This is a result of me digging around the command pipeline and trying to figure out where best to put the encryption decryption.

I'd love some feedback/guidance. Mainly, I'm wondering -- is there a more central location where I can get the command for any operation in the correct format? Why doesn't insert have a builder like find does? Should it?


egiurleo:
Work completed in #1646

PR #1631, opened 2020-01-10
RUBY-2071 Separate crypto and libmongocrypt interactions in crypto hooks

p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2071

PR #1632, opened 2020-01-10
RUBY-2072 refactor Binding to encapsulate more of the boilerplate required for libmongocrypt

p-mongo:


lib/mongo/crypt/binary.rb
```diff
@@ -142,7 +142,7 @@ def ref
       def self.wrap_string(str)
         binary_p = Binding.mongocrypt_binary_new_from_data(
           FFI::MemoryPointer.from_string(str),
-          str.length,
+          str.bytesize,
```
p-mongo: Great catch.

p-mongo: Could you also check the remaining code under crypt/ to see if this change needs to be made elsewhere?

lib/mongo/crypt/binding.rb
```diff
@@ -185,7 +185,7 @@ def self.setopt_schema_map(handle, schema_map_doc)
 
       def self.init(handle)
         check_status(handle) do
-          mongocrypt_init(handle)
+          mongocrypt_init(handle.ref)
```
p-mongo: :+1:

PR #1636, opened 2020-01-14
Parse Extended JSON using BSON::ExtJSON

egiurleo:


spec/mongo/client_auto_encryption_spec.rb
```diff
@@ -5,7 +5,7 @@
   require_enterprise
   clean_slate
 
-  let(:schema_map) { Utils.parse_extended_json(JSON.parse(File.read('spec/mongo/crypt/data/schema_map.json'))) }
+  let(:schema_map) { BSON::ExtJSON.parse_obj(JSON.parse(File.read('spec/mongo/crypt/data/schema_map.json'))) }
```
p-mongo: There is also a `parse` method available so that the following should work:

<pre>
 BSON::ExtJSON.parse(File.read('spec/mongo/crypt/data/schema_map.json'))
</pre>

egiurleo: Done, thank you!

PR #1639, opened 2020-01-15
Instructions on running FLE tests in test README

egiurleo:


spec/README.md
```diff
@@ -269,6 +269,45 @@ To run the test suite against such a server, run:
...
+Option 2: Build from source
+- To build libmongocrypt from source, follow the instructions in the README on the libmongocrypt GitHub repo: https://github.com/mongodb/libmongocrypt
+
+In one terminal, launch MongoDB:
+
+NOTE: You must be running MongoDB 4.2 or higher. All auto-encryption features
+will require an enterprise build of MongoDB, but you can still run
+explicit encryption tests without enterprise.
+
+Download different versions of MongoDB here: https://www.mongodb.com/download-center/community
```
egiurleo: This is the first set of tests that requires an enterprise build. Should I go more in depth about downloading and setting up a different MDB version?

p-mongo: Should the link be https://www.mongodb.com/download-center/enterprise for the enterprise version?

spec/README.md
```diff
@@ -269,6 +269,45 @@ To run the test suite against such a server, run:
...
+
+```
+mkdir /tmp/mdb
+mongod --dbpath /tmp/mdb --setParameter enableTestCommands=
```
p-mongo: Missing `1` at the end of this line.

spec/README.md
```diff
@@ -269,6 +269,45 @@ To run the test suite against such a server, run:
...
+```
+
+In another terminal, install bson-ruby from master, then run the tests, making
```
p-mongo: I would say it is not necessary to mention bson-ruby & master since this will need to be changed eventually anyway. The reason I asked for this PR is to cover the libmongocrypt aspect of it.

spec/README.md
```diff
@@ -269,6 +269,45 @@ To run the test suite against such a server, run:
...
+sure to set the `LIBMONGOCRYPT_PATH` environment variable to the location
+where you saved your libmongocrypt build.
```
p-mongo: I would say something like "to the full path to the .so/.dll/.dylib" to make it clear this isn't the path to the root of the extracted tree.

spec/README.md
```diff
@@ -269,6 +269,45 @@ To run the test suite against such a server, run:
...
+```
+bundle install --gemfile=gemfiles/bson_master.gemfile
+LIBMONGOCRYPT_PATH=/path/to/your/libmongocrypt/build bundle exec rake
```
p-mongo: Similarly here, something like

` LIBMONGOCRYPT_PATH=/path/to/your/libmongocrypt/nocrypto/libmongocrypt.so bundle exec rake`

Where the `nocrypto/libmongocrypt.so` is the actual relative path in the archive.

spec/README.md
```diff
@@ -269,6 +269,45 @@ To run the test suite against such a server, run:
...
+- Move that file to wherever you want to keep it on your machine.
+
+Option 2: Build from source
+- To build libmongocrypt from source, follow the instructions in the README on the libmongocrypt GitHub repo: https://github.com/mongodb/libmongocrypt
+
+In one terminal, launch MongoDB:
+
+NOTE: You must be running MongoDB 4.2 or higher. All auto-encryption features
+will require an enterprise build of MongoDB, but you can still run
+explicit encryption tests without enterprise.
```
p-mongo: Perhaps "using community edition" is more clear than "without enterprise"?

spec/README.md
```diff
@@ -269,6 +269,45 @@ To run the test suite against such a server, run:
...
+    libmongocrypt.dylib or libmongocrypt.dll file, depending on your OS.
+- Move that file to wherever you want to keep it on your machine.
+
+Option 2: Build from source
+- To build libmongocrypt from source, follow the instructions in the README on the libmongocrypt GitHub repo: https://github.com/mongodb/libmongocrypt
+
+In one terminal, launch MongoDB:
+
+NOTE: You must be running MongoDB 4.2 or higher. All auto-encryption features
+will require an enterprise build of MongoDB, but you can still run
```
p-mongo: Also remove the "will" from the beginning of this line.

spec/README.md
egiurleo: Good catch, thanks

PR #1645, opened 2020-01-20
RUBY-2078 encrypt symbols in a way that round-trips them to symbols


p-mongo:


spec/integration/explicit_encryption_spec.rb
```diff
@@ -50,12 +50,9 @@
...
-  # A change to BSON master has caused this test to fail.
-  # TODO: investigate and get working again
-  #
-  # context 'value is an symbol' do
-  #   let(:value) { :hello_world }
-
-  #   it_behaves_like 'an explicit encrypter'
-  # end
+  context 'value is an symbol' do
+    let(:value) { BSON::Symbol::Raw.new(:hello_world) }
```
p-mongo: Since the current preferred behavior is to store symbols as strings in the database, I think it's OK that the symbol raw type must be explicitly given to obtain a symbol round-trip when fle is involved.

egiurleo: I totally agree! Would you mind updating the `ClientEncryption#encrypt` docstring with the proper type?

p-mongo: I think that type would be `Object`, since it is my understanding that FLE in principle supports all types (and then excludes some due to a blacklist). Do you agree?

egiurleo: 👍 Yes I think that's a good call.

PR #1646, opened 2020-01-21
RUBY-1924 Auto-encryption and Auto-decryption

p-mongo:


spec/support/spec_config.rb
```diff
@@ -363,15 +363,24 @@ def retry_writes_options
   # Base test options.
   def base_test_options
     {
-      max_pool_size: 1,
+      # Automatic encryption tests require a minimum of two connections,
+      # because the driver checks out a connection to build a command,
+      # and then may need to encrypt the command which could require a
+      # query to key vault connection triggered from libmongocrypt.
+      # In the worst case using FLE may end up doubling the number of
+      # connections that the driver uses at any one time.
```
p-mongo: This should probably make its way into our documentation in some way.

lib/mongo/protocol/msg.rb
```diff
@@ -182,6 +182,35 @@ def documents
...
+        if client && client.encryption_options && !client.encryption_options[:bypass_auto_encryption]
+          db_name = @main_document[DATABASE_IDENTIFIER]
+          cmd = merge_sections
+          enc_cmd = client.encrypt(db_name, cmd)
+          if cmd.key?('$db') && !enc_cmd.key?('$db')
+            enc_cmd['$db'] = cmd['$db']
+          end
+          if enc_cmd['txnNumber'].is_a?(Integer) && cmd[:txnNumber].is_a?(BSON::Int64)
+            enc_cmd['txnNumber'] = BSON::Int64.new(enc_cmd[:txnNumber])
+          end
```
p-mongo: The $db and txnNumber fixes need to be investigated.

lib/mongo/crypt/context.rb
```diff
@@ -88,7 +88,7 @@ def run_state_machine
             filter = Binding.ctx_mongo_op(self)
 
             result = @encryption_io.collection_info(filter)
-            mongocrypt_feed(result)
+            mongocrypt_feed(result||{})
```
p-mongo: This appears to not match the LMC integration guide - if result is nil don't feed it at all.

lib/mongo/crypt/auto_encrypter.rb
```diff
@@ -160,7 +160,7 @@ def spawn_mongocryptd
       #
       # @return [ true ] Always true
       def teardown_encrypter
-        @mongocryptd_client.close if @mongocryptd_client
+        #@mongocryptd_client.close if @mongocryptd_client
```
p-mongo: This should be investigated also.

egiurleo: What exactly needs to be investigated here?

p-mongo: If the comment is removed, is the CI still green?

egiurleo: Yup!

spec/support/spec_config.rb
egiurleo: I have added a line to the [Client documentation](https://github.com/mongodb/mongo-ruby-driver/pull/1646/files#diff-9b873200a76992a56a5c5fa0fd7914b2R392). Let me know if you'd like to see this in other parts of the documentation as well.

lib/mongo/crypt/context.rb
egiurleo: I added a comment referencing a follow-up ticket.

lib/mongo/protocol/msg.rb
```diff
@@ -182,6 +182,36 @@ def documents
         [@main_document]
       end
 
+      def maybe_encrypt(client)
```
p-mongo: Can this and maybe_decrypt methods have docstrings added please?

spec/support/spec_config.rb
p-mongo: This could go into the tutorial but since we don't currently have an fle section in the tutorial let's come back to this later.

p-mongo:
:+1:

lib/mongo/protocol/msg.rb
egiurleo: I'm going to do this right now in a follow-up PR.

PR #1651, opened 2020-01-27
RUBY-1612 Compression integration test


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-1612

lib/mongo/protocol/message.rb
```diff
@@ -95,21 +95,44 @@ def replyable?
...
       #
       # @param [ String, Symbol ] compressor The compressor to use.
       # @param [ Integer ] zlib_compression_level The zlib compression level to use.
       #
-      # @return [ self ] Always returns self. Other message types should override this method.
+      # @return [ self ] Always returns self. Other message types should
+      #   override this method.
       #
       # @since 2.5.0
-      def compress!(compressor, zlib_compression_level = nil)
```
egiurleo: Is there any concern about changing the API here? Should we deprecate this method?

p-mongo: There are many methods in the driver that should be api-private but are not marked as such. I would say most if not all of the operations layer should be api-private. To this end I marked the replacement method api-private. I see that I am missing the corresponding labels on the overridden methods which I will add now.

egiurleo: Great, thanks.

PR #1653, opened 2020-01-28
RUBY-2007 Add AWS KMS options to Data Key Context

egiurleo:
The second in a series of a few pull requests.

### DIFF
--- a/lib/mongo/crypt/binding.rb
+++ b/lib/mongo/crypt/binding.rb
@@ -423,7 +423,7 @@ module Mongo
         end
       end
 
-      # Configure the ctx to take a masterkey from AWS
+      # Configure the ctx to take a master key from AWS
       #
       # @param [ FFI::Pointer ] ctx A pointer to a mongocrypt_ctx_object
       # @param [ String ] region The AWS region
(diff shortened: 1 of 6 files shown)


lib/mongo/crypt/binding.rb
```diff
@@ -423,6 +423,72 @@ def self.ctx_setopt_algorithm(context, name)
         end
       end
 
+      # Configure the ctx to take a masterkey from AWS
```
p-mongo: "master key" please

lib/mongo/crypt/data_key_context.rb
```diff
@@ -25,18 +25,104 @@ class DataKeyContext < Context
...
+            "Valid options are 'aws' and 'local'"
+          )
+        end
+
+        super(mongocrypt, io)
+
+        Binding.ctx_setopt_masterkey_local(self) if kms_provider == 'local'
+
+        if kms_provider == 'aws'
+          masterkey_opts = options[:masterkey]
```
p-mongo: `master_key` please in both

lib/mongo/crypt/data_key_context.rb
```diff
@@ -25,18 +25,104 @@ class DataKeyContext < Context
...
+        unless ['aws', 'local'].include?(kms_provider)
+          raise ArgumentError.new(
+            "#{kms_provider} is an invalid kms provider. " +
+            "Valid options are 'aws' and 'local'"
+          )
+        end
+
+        super(mongocrypt, io)
+
+        Binding.ctx_setopt_masterkey_local(self) if kms_provider == 'local'
```
p-mongo: Can this and the following if statement be organized into a single statement, either case/when or if/elsif please.

lib/mongo/crypt/data_key_context.rb
egiurleo: 👍 Will audit all my future pull requests for this.

PR #1658, opened 2020-01-29
RUBY-2106 Mongocryptd Spawning Improvements + Some Error Cleanup

egiurleo:
This PR does a few things:
1. Always set mongocryptd timeout to 1
2. Create a class for errors while spawning mongocryptd and improve the error message that is displayed.
3. Clean up the FLE-related error classes, including removing the error code (which might mean something in the context of libmongocrypt but is meaningless to users), cleaning up class names, and standardizing file names.


p-mongo:
>  including removing the error code (which might mean something in the context of libmongocrypt but is meaningless to users)

I think it is helpful to provide as much diagnostic information as possible. It could be that someone is debugging libmongocrypt and doing it through the Ruby driver and the code means something to them.

To clarify that the code is not produced by the driver, how about using language to the effect of "libmongocrypt code X"?

egiurleo:
I like that suggestion. I'll make that change shortly.

PR #1661, opened 2020-01-31
RUBY-2007 DRY up encryption specs

egiurleo:
There was a lot of copy + paste in the specs, so I wanted to create a helper with some shared variables and test behavior. I'm worried about making the tests hard to follow (i.e. a developer has to reference this helper in order to understand the tests), so I'm open to feedback on whether this actually improves things.


spec/support/crypt.rb
```diff
@@ -0,0 +1,77 @@
...
+    let(:data_key) do
+      BSON::ExtJSON.parse(File.read('spec/mongo/crypt/data/key_document_aws.json'))
+    end
+
+    let(:schema_map) do
+      BSON::ExtJSON.parse(File.read('spec/mongo/crypt/data/schema_map_aws.json'))
+    end
+  end
+
+  def self.included(context)
```
p-mongo: Can all of these methods be moved into a RSpec shared context, and then the Crypt module can be included globally in rspec config.

spec/mongo/client_auto_encryption_spec.rb
```diff
@@ -18,7 +18,7 @@
   end
 
   let(:data_key) do
-    BSON::ExtJSON.parse(File.read('spec/mongo/crypt/data/key_document.json'))
+    BSON::ExtJSON.parse(File.read('spec/support/crypt/data_keys/key_document.json'))
```
p-mongo: :+1:

spec/mongo/crypt/auto_decryption_context_spec.rb
```diff
@@ -3,18 +3,12 @@
 
 describe Mongo::Crypt::AutoDecryptionContext do
   require_libmongocrypt
+  include_context 'define shared FLE helpers'
```
p-mongo: :+1:

spec/mongo/crypt/auto_decryption_context_spec.rb
```diff
@@ -29,6 +23,8 @@
   describe '#initialize' do
     context 'with valid options' do
       context 'when mongocrypt is initialized with local KMS provider options' do
+        include_context 'with local kms_providers'
```
p-mongo: :+1:

PR #1662, opened 2020-01-31
SPEC-1517 Commit the right test

p-mongo:


p-mongo:
Committed an old version of the test by mistake.

PR #1670, opened 2020-02-05
RUBY-2118 Make exception file names match class names + Improve CryptError API

egiurleo:


lib/mongo/error/crypt_error.rb
```diff
@@ -17,7 +17,7 @@ class CryptError < Mongo::Error
       #
       # @param [ Integer | nil ] code The optional libmongocrypt error code
       # @param [ String ] message The error message
-      def initialize(code, message)
+      def initialize(message, code: nil)
```
p-mongo: :+1:

PR #1673, opened 2020-02-06
RUBY-2116 Verify that the data based to libmongocrypt is always in the form of a BSON::Document

egiurleo:
I realized that there were still a lot of places where we were passing plain Hashes to libmongocrypt, so I fixed those as well.

### DIFF
--- a/lib/mongo/client_encryption.rb
+++ b/lib/mongo/client_encryption.rb
@@ -79,7 +79,7 @@ module Mongo
     # @return [ BSON::Binary ] A BSON Binary object of subtype 6 (ciphertext)
     #   representing the encrypted value
     def encrypt(value, opts={})
-      doc = BSON::Document.new('v': value)
+      doc = { 'v': value }
 
       Crypt::ExplicitEncryptionContext.new(
         @crypt_handle,
@@ -96,7 +96,7 @@ module Mongo
     #
     # @return [ Object ] The decrypted value
     def decrypt(value)
-      doc = BSON::Document.new('v': value)
+      doc = { 'v': value }
 
       result = Crypt::ExplicitDecryptionContext.new(
         @crypt_handle,
--- a/lib/mongo/client_encryption.rb
+++ b/lib/mongo/client_encryption.rb
@@ -79,7 +79,7 @@ module Mongo
     # @return [ BSON::Binary ] A BSON Binary object of subtype 6 (ciphertext)
     #   representing the encrypted value
     def encrypt(value, opts={})
-      doc = { 'v': value }
+      doc = BSON::Document.new('v': value)
 
       Crypt::ExplicitEncryptionContext.new(
         @crypt_handle,
@@ -96,7 +96,7 @@ module Mongo
     #
     # @return [ Object ] The decrypted value
     def decrypt(value)
-      doc = { 'v': value }
+      doc = BSON::Document.new('v': value)
 
       result = Crypt::ExplicitDecryptionContext.new(
         @crypt_handle,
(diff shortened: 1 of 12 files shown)


lib/mongo/client_encryption.rb
```diff
@@ -79,7 +79,7 @@ def create_data_key(kms_provider, options={})
     # @return [ BSON::Binary ] A BSON Binary object of subtype 6 (ciphertext)
     #   representing the encrypted value
     def encrypt(value, opts={})
-      doc = { 'v': value }
+      doc = BSON::Document.new('v': value)
```
p-mongo: Actually this was fine the way it was. The principal issue with Hash vs Document is when using symbol keys and then using the `#keys` method or iterating the hash, since Document stringifies the keys.

The Document way is fine as well, though I imagine Hash is more performant since keys aren't converted.

lib/mongo/crypt/auto_decryption_context.rb
```diff
@@ -29,7 +29,7 @@ class AutoDecryptionContext < Context
       #   state machine
       # @param [ String ] db_name The name of the database against which
       #   the command is being made
-      # @param [ Hash ] command The command to be encrypted
+      # @param [ BSON::Document ] command The command to be encrypted
```
p-mongo: What this to me is saying is this method expects to be able to access `command` indifferently (i.e. either with symbol or string keys). If so there may need to be a check added that the argument is in fact a Document and not just a Hash. Specifying Hash in the docstring is also legitimate, in which case the method and the caller must agree on whether keys are strings or symbols. I think both approaches are valid, they have different pros/cons.

lib/mongo/crypt/binding.rb
```diff
@@ -557,13 +558,17 @@ def self.ctx_datakey_init(context)
...
-      # @raise [ Error::CryptError ] If initialization fails
-      def self.ctx_encrypt_init(context, db_name, an_int, command)
+      # @raise [ Mongo::Error::CryptError ] If initialization fails
+      def self.ctx_encrypt_init(context, db_name, command)
+        validate_document(command)
         data = command.to_bson.to_s
         Binary.wrap_string(data) do |data_p|
           check_ctx_status(context) do
-            mongocrypt_ctx_encrypt_init(context.ctx_p, db_name, an_int, data_p)
+            mongocrypt_ctx_encrypt_init(context.ctx_p, db_name, -1, data_p)
```
p-mongo: :+1:

lib/mongo/crypt/binding.rb
```diff
@@ -1089,6 +1098,26 @@ def self.check_ctx_status(context)
...
       end
+
+      # Checks that the specified data is a BSON::Document before serializing
+      # it to BSON to prevent errors from libmongocrypt
+      #
+      # @param [ Object ] data The data to be passed to libmongocrypt
+      #
+      # @raise [ Mongo::Error::CryptError ] If the data is not a BSON::Document
+      def self.validate_document(data)
+        return if data.is_a?(BSON::Document)
```
p-mongo: Although I realize I said "BSON::Document" in the ticket, I believe Hash here will work also (with slightly different behavior in the driver implementation but both approaches are valid).

egiurleo:
@p-mongo Thanks for clarifying your intentions with the ticket. I've changed the validator to allow Hashes. Can you let me know if this is closer to what you had in mind?

lib/mongo/crypt/binding.rb
```diff
@@ -1089,6 +1098,28 @@ def self.check_ctx_status(context)
...
+      # @raise [ Mongo::Error::CryptError ] If the data is not a Hash
+      def self.validate_document(data)
+        return if data.is_a?(Hash)
+
+        if data.nil?
+          message = "Attempted to pass nil data to libmongocrypt. " +
+            "Data must be a BSON::Document"
+        else
+          message = "Attempted to pass invalid data to libmongocrypt: #{data} " +
+            "Data must be a BSON::Document"
```
p-mongo: `Hash` here right?

egiurleo: Yup, thanks for catching that.

PR #1674, opened 2020-02-06
RUBY-2099 Clarify behavior when "me" field doesn't match any values in "hosts" array


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2099

PR #1677, opened 2020-02-07
RUBY-1816 Specify behavior where connection string contain auth database but no credentials


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-1816

PR #1678, opened 2020-02-10
RUBY-2125 Validate write concern options when parsing URI


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2125

PR #1682, opened 2020-02-10
Assorted evergreen fixes/improvements

p-mongo:


egiurleo:
Are the things you commented out left in the code for a reason?

p-mongo:
Shell tracing output is useful when diagnosing problems, but it gets in the way when everything is working. I left those lines commented out so that if a problem needs to be diagnosed it is easy to uncomment them instead of figuring out where tracing needs to be enabled.

.evergreen/config.yml
```diff
@@ -434,7 +434,7 @@ pre:
   - func: "fetch source"
   - func: "create expansions"
   - func: "prepare resources"
-  - func: "windows fix"
+  #- func: "windows fix"
```
p-mongo: This might need to be reinstated if we start testing on windows in evergreen.

PR #1683, opened 2020-02-10
RUBY-2007 Clean up SSLSocket usage

egiurleo:
* Improve URI parsing
* Do not use the driver-specific implementation of SSL sockets
* Make sure to set nodelay option, perform SNI, and set a timeout
* Improve error handling

### DIFF
--- a/lib/mongo/crypt/binding.rb
+++ b/lib/mongo/crypt/binding.rb
@@ -793,7 +793,7 @@ module Mongo
           mongocrypt_kms_ctx_message(kms_context.kms_ctx_p, binary.ref)
         end
 
-        return binary.to_string
+        return binary.to_s
       end
 
       # Get the hostname with which to connect over TLS to get information about
--- a/lib/mongo/crypt/encryption_io.rb
+++ b/lib/mongo/crypt/encryption_io.rb
@@ -20,6 +20,12 @@ module Mongo
     #
     # @api private
     class EncryptionIO
+
+      # Timeout used for SSL socket connection, reading, and writing.
+      # There is no specific timeout written in the spec. See SPEC-1394
+      # for a discussion and updates on what this timeout should be.
+      SOCKET_TIMEOUT = 10
+
       # Creates a new EncryptionIO object with information about how to connect
       # to the key vault.
       #
@@ -96,11 +102,23 @@ module Mongo
       #   to send on that connection.
       def feed_kms(kms_context)
         with_ssl_socket(kms_context.endpoint) do |ssl_socket|
-          ssl_socket.puts(kms_context.message)
 
-          while kms_context.bytes_needed > 0 do
-            bytes = ssl_socket.sysread(kms_context.bytes_needed)
+          Timeout.timeout(SOCKET_TIMEOUT, Error::SocketTimeoutError,
+            'Socket write operation timed out'
+          ) do
+            ssl_socket.syswrite(kms_context.message)
+          end
+
+          bytes_needed = kms_context.bytes_needed
+          while bytes_needed > 0 do
+            bytes = Timeout.timeout(SOCKET_TIMEOUT, Error::SocketTimeoutError,
+              'Socket read operation timed out'
+            ) do
+              ssl_socket.sysread(bytes_needed)
+            end
+
             kms_context.feed(bytes)
+            bytes_needed = kms_context.bytes_needed
           end
         end
       end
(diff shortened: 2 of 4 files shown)


lib/mongo/crypt/encryption_io.rb
```diff
@@ -95,29 +95,58 @@ def mark_command(cmd)
...
         # There is no specific timeout written in the spec. See SPEC-1394
         # for a discussion and updates on what this timeout should be.
         socket_timeout = 10
+        default_port = 443
 
-        host, port = endpoint.split(':')
-        port ||= 443
+        unless endpoint.start_with?('https://') || endpoint.start_with?('//')
+          endpoint = "//#{endpoint}"
+        end
```
egiurleo: My goal here is to make sure that URIs are correctly parsed in as many situations as possible. By default, LMC returns the KMS endpoint in the format `kms.us-east-1.amazonaws.com`, which confuses Ruby's URI parser because there is no scheme. Is there a better or more reliable way to handle this case?

lib/mongo/crypt/binding.rb
```diff
@@ -793,7 +793,7 @@ def self.kms_ctx_message(kms_context)
           mongocrypt_kms_ctx_message(kms_context.kms_ctx_p, binary.ref)
         end
 
-        return binary.to_string
+        return binary.to_s
```
p-mongo: Can this fix be in a separate PR please with the ticket reference to the to_string -> to_s change?

lib/mongo/crypt/encryption_io.rb
p-mongo: https://github.com/mongodb/libmongocrypt/blob/master/integrating.md#state-mongocrypt_ctx_need_kms says:

> Create/reuse a TLS socket connected to the endpoint indicated by mongocrypt_kms_ctx_endpoint. The endpoint string is a host name with optional port number separated by a colon

So, this is actually supposed to be a URI, but just the host name, right?

egiurleo: Okay. In that case, do you think it's fair for me to just split the string at the colon and then raise an exception if the result doesn't have two parts? I can also specify in the documentation that any user-specified endpoint should take this same format.

p-mongo: Yes per our discussion just now.

lib/mongo/crypt/encryption_io.rb
```diff
@@ -95,29 +95,58 @@ def mark_command(cmd)
       #   the endpoint at which to establish a TLS connection and the message
       #   to send on that connection.
       def feed_kms(kms_context)
-        endpoint = kms_context.endpoint
-        message = kms_context.message
+        with_ssl_socket(kms_context.endpoint) do |ssl_socket|
+          ssl_socket.puts(kms_context.message)
```
p-mongo: s/puts/write/, puts adds another newline

p-mongo: The write call should have a timeout.

lib/mongo/crypt/encryption_io.rb
```diff
@@ -95,29 +95,58 @@ def mark_command(cmd)
...
 
+          while kms_context.bytes_needed > 0 do
+            bytes = ssl_socket.sysread(kms_context.bytes_needed)
```
p-mongo: This read call should also have a timeout.

lib/mongo/crypt/encryption_io.rb
```diff
@@ -95,29 +95,58 @@ def mark_command(cmd)
       #   the endpoint at which to establish a TLS connection and the message
       #   to send on that connection.
       def feed_kms(kms_context)
-        endpoint = kms_context.endpoint
-        message = kms_context.message
+        with_ssl_socket(kms_context.endpoint) do |ssl_socket|
+          ssl_socket.puts(kms_context.message)
 
+          while kms_context.bytes_needed > 0 do
```
p-mongo: Please save bytes needed into a variable.

lib/mongo/crypt/encryption_io.rb
```diff
@@ -95,29 +101,68 @@ def mark_command(cmd)
...
 
-          kms_context.feed(bytes.pack('C*'))
-          num_bytes_needed = kms_context.bytes_needed
+          yield(ssl_socket)
+        rescue => e
+          raise Error::KmsError, "Error decrypting data key. #{e.class}: #{e.message}"
+        ensure
+          # If there is an error during socket creation, the
+          # ssl_socket object won't exist in this scope
+          ssl_socket.sysclose if ssl_socket
```
p-mongo: This close could conceivably also fail, especially since it is an ssl socket and close may perform i/o. Either putting the close call under the rescue raising KmsError above or adding `rescue nil` should be reasonable measures to deal with this. I don't know if the close call should also be guarded with a timeout; I suppose it won't hurt if you wish to do this.

PR #1701, opened 2020-02-14
SPEC-1248 RUBY-2126 Unify behavior around configuration for replica set discovery 

p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2126

PR #1702, opened 2020-02-14
RUBY-2008 Allow user to specify key alt names when creating a data key

egiurleo:
This is the first PR of a few, with the purpose of making it possible to refer to encryption keys by alternate names.

### DIFF
--- a/lib/mongo/client_encryption.rb
+++ b/lib/mongo/client_encryption.rb
@@ -44,7 +44,7 @@ module Mongo
     #   "aws" and "local".
     # @params [ Hash ] options
     #
-    # @option [ Hash ] :master_key Information about the AWS master key. Required
+    # @option options [ Hash ] :master_key Information about the AWS master key. Required
     #   if kms_provider is "aws".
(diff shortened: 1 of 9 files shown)


lib/mongo/client_encryption.rb
```diff
@@ -53,6 +53,8 @@ def initialize(key_vault_client, options = {})
     #     by a colon (e.g. "kms.us-east-1.amazonaws.com" or
     #     "kms.us-east-1.amazonaws.com:443"). An endpoint in any other format
     #     will not be properly parsed.
+    # @option [ Array ] :key_alt_names An optional array of strings specifying
```
p-mongo: ` @option options [ Array<String> ] `

p-mongo: There's also the "options" between @option and the type.

egiurleo: I've been declaring these incorrectly 😆Just fixed all of them.

p-mongo: :+1:

PR #1705, opened 2020-02-15
RUBY-2008 Test auto and explicit encryption with key alt names

egiurleo:


spec/integration/auto_encryption_spec.rb
```diff
@@ -440,6 +442,44 @@
...
+          expect(result).to be_ok
+          expect(result.inserted_ids.length).to eq(1)
+
+          id = result.inserted_ids.first
+
+          document = client[:users].find(_id: id).first
+          document.should_not be_nil
+          # Auto-encryption with key alt names only works with random encryption,
+          # so it will not generate the same result on every test run.
+          expect(document['ssn']).to be_a_kind_of(BSON::Binary)
```
p-mongo: I think this is an example of a situation where it makes sense to write an "is ciphertext" matcher, which would check the type and the bson subtype.

egiurleo: Yeah I like that!

PR #1710, opened 2020-02-17
RUBY-2010 Raise an error when attempting to encrypt on wire version < 8

egiurleo:


lib/mongo/operation/shared/executable.rb
```diff
@@ -54,7 +54,7 @@ def get_result(server, client)
       def dispatch_message(server, client)
         server.with_connection do |connection|
           message = build_message(server)
-          message = message.maybe_encrypt(client)
+          message = message.maybe_encrypt(client, server)
```
p-mongo: Can the argument order be flipped here to match `dispatch_message` signature?

lib/mongo/protocol/msg.rb
```diff
@@ -195,10 +195,18 @@ def documents
...
-      def maybe_encrypt(client)
+      def maybe_encrypt(client, server)
         # TODO verify compression happens later, i.e. when this method runs
         # the message is not compressed.
         if client && client.encryption_options && !client.encryption_options[:bypass_auto_encryption]
+          if server.max_wire_version < 8
+            raise Error::CryptError.new(
+              "Cannot perform encryption against a server with MongoDB wire " +
+              "version less than 8. Currently connected to server with wire " +
+              "version #{server.max_wire_version}"
```
p-mongo: I think this will be more user-friendly if phrased as: *Cannot perform encryption against a MongoDB server older than 4.2 (wire version less than 8). Currently connected to server with max wire version X*

I also added a "max" at the end.

spec/integration/auto_encryption_spec.rb
```diff
@@ -4,6 +4,7 @@
 
 describe 'Auto Encryption' do
   require_libmongocrypt
+  min_server_version '4.2'
```
p-mongo: We generally use `min_server_fcv` in preference to `min_server_version` since most times a server with a lower fcv doesn't work. I am not sure whether this matters for encryption.

lib/mongo/operation/shared/executable.rb
egiurleo: Yup!

lib/mongo/protocol/msg.rb
egiurleo: Agreed that that's a lot clearer.

PR #1719, opened 2020-02-21
add timeouts to srv monitoring spec


p-mongo:


p-mongo:
This should help with the intermittent failures in this test as well as reduce its runtime potentially.

PR #1721, opened 2020-02-21
Fix most tests on jruby

p-mongo:


egiurleo:
Thank you for fixing this 🎉

PR #1723, opened 2020-02-24
Migrate regular test configurations to use mlaunch

p-mongo:


.evergreen/config.yml
```diff
@@ -614,284 +584,296 @@ axes:
...
+      topology: "*"
+      os: ubuntu1204
+    display_name: "${mongodb-version} ${topology} ${auth-and-ssl} ${ruby}"
+    tasks:
+      - name: "test-mlaunch"
+
+  - matrix_name: "mongo-latest"
+    matrix_spec:
+      auth-and-ssl: ["auth-and-ssl", "noauth-and-nossl"]
+      ruby: "ruby-2.7"
```
egiurleo: How did you choose which mongo/ruby version pairs to use?

p-mongo: I wanted to combine coverage of all server versions and coverage of all ruby versions. The exact pairings are to some extent arbitrary, e.g. if ruby 2.7 is tested with latest then 4.2 could be tested with either 2.7 or 2.6 and 4.0 could be tested with the other one.

egiurleo: Makes sense, thanks!

PR #1724, opened 2020-02-25
RUBY-2107 Make AutoEncrypter and ExplicitEncrypter their own classes

egiurleo:
**Previous design of encryption:**
`AutoEncrypter` and `Encrypter` were both modules that were included in `ClientEncryption` and `Client`, thus exposing their methods to these user-facing classes. `AutoEncrypter` was a subclass of `Encrypter.`

**New design:**
`AutoEncrypter` and `ExplicitEncrypter` are both classes that encapsulate encryption behavior. `Client` has an instance variable called `@encrypter` that is an instance of `AutoEncrypter`, and similarly, `ClientEncryption` has an instance variable called `@encrypter` that is an instance of `ExplicitEncrypter`. Some validation that was previously shared by the encrypters has been moved to `EncryptionIO`. Spawning mongocryptd has also been moved to `EncryptionIO`. 

I also added a test to make sure that command monitoring includes events from the key vault client.


lib/mongo/crypt/auto_encrypter.rb
```diff
@@ -52,46 +55,48 @@ module AutoEncrypter
...
+          mongocryptd_options: @options[:extra_options]
         )
       end
 
-      # Encrypt a database command
+      # Whether this encrypter should perform encryption (returns false if
+      # the :bypass_auto_encryption option is set to true).
+      #
+      # @return [ Boolean ] Whether to perform encryption.
+      def should_encrypt
```
p-mongo: Since this is a predicate method it should have a `?` at the end, and perhaps `encrypt?` would be sufficient.

egiurleo: Good point! Thanks.

PR #1725, opened 2020-02-25
RUBY-2142 Add a filter option to list_collections

egiurleo:
Integrating with libmongocrypt requires listCollections to be performed with a filter, which Ruby hasn't supported. I've been hacking around it, but that makes the FLE command monitoring specs fail, so I figured I would fix this if it was easy and didn't break the user-facing API. Let me know if you think this is okay and I'll make a JIRA ticket for it as well.

### DIFF
--- a/lib/mongo/database.rb
+++ b/lib/mongo/database.rb
@@ -130,8 +130,8 @@ module Mongo
     #   collection in the database.
     #
     # @since 2.0.5
-    def list_collections
-      View.new(self).list_collections
+    def list_collections(filter={})
+      View.new(self).list_collections(filter)
     end
 
     # Get all the non-system collections that belong to this database.
--- a/lib/mongo/database.rb
+++ b/lib/mongo/database.rb
@@ -126,16 +126,18 @@ module Mongo
     #   information hash per collection, depends on the MongoDB server
     #   version that fulfills the request.
     #
-    # @param [ Hash ] filter A filter on the collections returned by
-    #   this command. See https://docs.mongodb.com/manual/reference/command/listCollections/
-    #   for documentation on the filter option.
+    # @param [ Hash ] options
+    #
+    # @option options [ Hash ] :filter A filter on the collections returned.
+    #   See https://docs.mongodb.com/manual/reference/command/listCollections/
+    #   for more information and usage.
     #
     # @return [ Array<Hash> ] Array of information hashes, one for each
     #   collection in the database.
     #
     # @since 2.0.5
-    def list_collections(filter={})
-      View.new(self).list_collections(filter: filter)
+    def list_collections(**options)
+      View.new(self).list_collections(options)
     end
 
     # Get all the non-system collections that belong to this database.
--- a/lib/mongo/database.rb
+++ b/lib/mongo/database.rb
@@ -137,7 +137,7 @@ module Mongo
     #
     # @since 2.0.5
     def list_collections(**options)
-      View.new(self).list_collections(options)
+      View.new(self).list_collections(**options)
     end
 
     # Get all the non-system collections that belong to this database.
--- a/lib/mongo/database.rb
+++ b/lib/mongo/database.rb
@@ -127,14 +127,15 @@ module Mongo
     #   version that fulfills the request.
     #
     # @param [ Hash ] filter A filter on the collections returned by
-    #   this command.
+    #   this command. See https://docs.mongodb.com/manual/reference/command/listCollections/
+    #   for documentation on the filter option.
     #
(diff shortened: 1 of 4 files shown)


lib/mongo/database.rb
```diff
@@ -126,12 +126,15 @@ def collection_names(options = {})
     #   information hash per collection, depends on the MongoDB server
     #   version that fulfills the request.
     #
+    # @param [ Hash ] filter A filter on the collections returned by
```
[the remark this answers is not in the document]
p-mongo: I.e. I think this could say "See the documentation for `filter` on https://docs.mongodb.com/manual/reference/command/listCollections/".

lib/mongo/database/view.rb
```diff
@@ -81,9 +81,9 @@ def collection_names(options = {})
       # @return [ Array<Hash> ] Info for each collection in the database.
       #
       # @since 2.0.5
-      def list_collections
+      def list_collections(filter={})
```
p-mongo: Looking at https://docs.mongodb.com/manual/reference/command/listCollections/, the command accepts 3 parameters: filter, nameOnly and authorizedCollections. To allow the driver to be future-proof, can this method take `**options` as the argument which will be passed to `collections_info`, and make `filter` one of the options? We do not need to support the other options right now.

spec/mongo/database_spec.rb
```diff
@@ -180,6 +180,24 @@
       it 'returns a list of the collections info' do
         expect(result).to include('acol')
       end
+
+      context 'with more than one collection' do
```
p-mongo: There should be an assertion somewhere in this context that there is in fact more than one collection present in the database being used.

lib/mongo/database/view.rb
egiurleo: That makes a lot of sense. Thank you! 🙏

lib/mongo/database.rb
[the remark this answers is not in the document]
egiurleo: Sure thing.

spec/mongo/database_spec.rb
```diff
@@ -180,6 +180,26 @@
...
+        before do
+          database[:anothercol].drop
+          database[:anothercol].create
+
+          expect(database.collections.length).to be > 1
```
egiurleo: @p-mongo Is this the kind of assertion you were looking for?

lib/mongo/database.rb
```diff
@@ -126,12 +126,16 @@ def collection_names(options = {})
...
+    #   this command. See https://docs.mongodb.com/manual/reference/command/listCollections/
+    #   for documentation on the filter option.
+    #
     # @return [ Array<Hash> ] Array of information hashes, one for each
     #   collection in the database.
     #
     # @since 2.0.5
-    def list_collections
-      View.new(self).list_collections
+    def list_collections(filter={})
```
egiurleo: @p-mongo I'm wondering if it would make more sense to make this a double-splat argument as well, since we're going to eventually want to change this method to take multiple options.

spec/mongo/database_spec.rb
p-mongo: Yes this works.

lib/mongo/database.rb
p-mongo: Yes please.

lib/mongo/database/view.rb
```diff
@@ -78,12 +78,20 @@ def collection_names(options = {})
...
       # @return [ Array<Hash> ] Info for each collection in the database.
       #
       # @since 2.0.5
-      def list_collections
+      def list_collections(**options)
+        filter = options[:filter] || {}
+
         session = client.send(:get_session)
-        collections_info(session, ServerSelector.primary)
+        collections_info(session, ServerSelector.primary, { filter: filter })
```
p-mongo: Can this just pass `**options`?

egiurleo: Yes, you're right.

lib/mongo/database/view.rb
```diff
@@ -78,12 +78,20 @@ def collection_names(options = {})
...
+      # @option options [ Hash ] :filter A filter on the collections returned.
+      #   See https://docs.mongodb.com/manual/reference/command/listCollections/
+      #   for more information and usage.
+      #
       # @return [ Array<Hash> ] Info for each collection in the database.
       #
       # @since 2.0.5
-      def list_collections
+      def list_collections(**options)
+        filter = options[:filter] || {}
```
p-mongo: This line is no longer needed, right?

egiurleo: Right, thanks!

PR #1732, opened 2020-02-27
RUBY-2042 Additions to FLE Documentation

egiurleo:
Documentation for:
* creating a master key with AWS and local
* creating a data key with AWS and local
* creating a schema map and reading it with `BSON::ExtJson#parse`

### DIFF
--- a/docs/tutorials/client-side-encryption.txt
+++ b/docs/tutorials/client-side-encryption.txt
@@ -252,7 +252,7 @@ Local Master Key
 A local master key is a Base64-encoded, 96-byte string. It should be persisted
 on your machine as an environment variable or in a text file.
 
-.. note::
+.. warning::
 
   Using a local master key is insecure and not recommended if you plan
   to use client-side encryption in production.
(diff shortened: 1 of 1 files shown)


docs/tutorials/client-side-encryption.txt
```diff
@@ -241,12 +241,205 @@ or creating a schema map, see later sections of this tutorial.
...
+
+When you intend to use your schema map, convert it to a Ruby ``Hash`` using the
+``BSON::ExtJSON`` module in the ``bson`` Ruby gem.
+
+.. code-block:: ruby
+
+  require 'bson'
+
+  # schema_map is a Ruby Hash
+  schema_map = BSON::ExtJSON.parse(File.open('/path/to/your/file.json'))
```
p-mongo: Does this work in both default and :bson parsing mode?

egiurleo: It does! Is there one you would recommend over the other?

p-mongo: Let's leave it the way it is for now.

PR #1733, opened 2020-02-27
RUBY-2121 Calling #use on auto encryption client should produce another functional auto encryption client

egiurleo:
Also, allow clients to share `AutoEncrypter` and don't create a new cluster when `:auto_encryption_options` change.


lib/mongo/client.rb
```diff
@@ -679,23 +682,23 @@ def update_options(new_options)
...
-          should_set_new_encryption_options = true
-        end
-
         options.update(opts)
         @options = options.freeze
 
-        close_encrypter if should_close_encrypter
-        build_encrypter if should_set_new_encryption_options
+        auto_encryption_options_changed =
+          @options[:auto_encryption_options] != old_options[:auto_encryption_options]
```
egiurleo: I'm not sure how people feel about multi-line variable assignment. Open to other styles here :)

p-mongo: This looks good :+1:

lib/mongo/client.rb
```diff
@@ -679,23 +682,23 @@ def update_options(new_options)
...
+
+        # If there are new auto_encryption_options, create a new encrypter.
+        # Otherwise, allow the new client to share an encrypter with the
+        # original client.
+        #
+        # If auto_encryption_options are nil, set @encrypter to nil, but do not
+        # close the encrypter because it may still be used by the original client.
```
p-mongo: :+1:

PR #1737, opened 2020-02-28
RUBY-1926 Include $db field on all command started events

egiurleo:


spec/integration/auto_encryption_command_monitoring_spec.rb
```diff
@@ -52,14 +52,15 @@
   end
 
   let(:key_vault_list_collections_event) do
-    subscriber.succeeded_events.find do |event|
+    subscriber.started_events.find do |event|
```
egiurleo: This was a typo.

PR #1750, opened 2020-03-03
RUBY-1926 unmodified upstream spec tests

p-mongo:
https://jira.mongodb.com/browse/RUBY-1926


p-mongo:
I do not doubt that it will. The idea with this PR is to commit the tests as they exist in the specifications repo; changes to the tests to make them work with Ruby driver will be in the PR that adds the runner and all the other bits.

egiurleo:
Okay, thanks for confirming.

PR #1752, opened 2020-03-03
RUBY-2152 Implemented "data key and double encryption" prose test

egiurleo:
Also change the return type of `ClientEncryption#create_data_key` as mandated by the spec.

### DIFF
--- a/docs/tutorials/client-side-encryption.txt
+++ b/docs/tutorials/client-side-encryption.txt
@@ -284,7 +284,8 @@ Once you have created a master key, create a data key by calling the
 ``#create_data_key`` method on an instance of the ``Mongo::ClientEncryption``
 class. This method generates a new data key and inserts it into the key vault
 collection, which is the MongoDB collection in which you choose to store your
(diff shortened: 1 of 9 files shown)


docs/tutorials/client-side-encryption.txt
```diff
@@ -284,7 +284,8 @@ Once you have created a master key, create a data key by calling the
 ``#create_data_key`` method on an instance of the ``Mongo::ClientEncryption``
 class. This method generates a new data key and inserts it into the key vault
 collection, which is the MongoDB collection in which you choose to store your
-data keys. The ``#create_data_key`` method returns the UUID of the newly-created data key.
+data keys. The ``#create_data_key`` method returns id of the newly-created
+data key in the form of a BSON::Binary object.
```
p-mongo: I think if we adopt this API then we should also provide user guidance as to how to serialize and deserialize bson binaries. What do you think? Or alternatively, what was the issue with the previous API?

egiurleo: I'm happy to write that up. The only issue with the previous API is that it wasn't consistent with the spec.

PR #1756, opened 2020-03-04
RUBY-2152 Add corpus prose tests for client-side encryption

egiurleo:


spec/integration/client_side_encryption/corpus_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+          .first
+
+        # Ensure that corpus_decrypted is the same as the original corpus
+        # document by checking that they have the same set of keys, and that
+        # they have the same values at those keys (improved diagnostics).
+        expect(corpus_decrypted.keys - corpus.keys).to be_empty
+
+        corpus_decrypted.each do |key, doc|
+          if doc['value'].is_a?(Time)
+            expect(doc['value'].utc.to_s).to eq(corpus[key]['value'].utc.to_s)
```
egiurleo: This is a quick hack to fix the problem I describe below. Would love some guidance on how to resolve this:

`corpus[key]['value']` is a `Time` object that has been parsed from extended JSON: 
  `1969-12-31 19:00:12 194217733930353/562949953421312 -0500`

`doc['value']` is the result of using explicit encryption to encrypt that `Time` object and then decrypt it: 
  `1970-01-01 00:00:12.345 UTC`

They represent the same time, but when you call `#to_r` on them, different results are returned:
```
corpus[key]['value'].to_r # => 6949617174986097/562949953421312
doc['value'].to_r # => 2469/200
```

Could this have to do with the millisecond problem in our bson implementation?

spec/integration/client_side_encryption/corpus_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+
+    let(:corpus_encrypted_expected) do
+      BSON::ExtJSON.parse(File.read('spec/support/crypt/corpus/corpus_encrypted.json'))
+    end
+
+    let(:corpus_copied) do
+      # As per the instructions of the prose spec, corpus_copied is a copy of
+      # the corpus BSON::Document that encrypts all fields that are meant to
+      # be explicitly encrypted. corpus is a document containing many
+      # sub-documents, each with a value to encrypt and inforrmation about how
```
p-mongo: inforrmation typo

spec/integration/client_side_encryption/corpus_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+
+            corpus_copied[key] = doc.merge('value' => encrypted_value)
+          rescue => e
+            # If doc['allowed'] is true, it means that this field should have
+            # been encrypted without error, and thus that this error is unexpected.
+            # If doc['allowed'] is false, this error was expected and the value
+            # should be copied over without being encrypted.
+            if doc['allowed']
+              raise "Unexpected error occured in client-side encryption " +
+                "corpus tests: #{e.class}, #{e.message}"
```
p-mongo: `#{e.class}: #{e}`

spec/integration/client_side_encryption/corpus_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+          .inserted_id
+
+        corpus_decrypted = client_encrypted[:coll]
+          .find(_id: corpus_encrypted_id)
+          .first
+
+        # Ensure that corpus_decrypted is the same as the original corpus
+        # document by checking that they have the same set of keys, and that
+        # they have the same values at those keys (improved diagnostics).
+        expect(corpus_decrypted.keys - corpus.keys).to be_empty
```
p-mongo: This allows corpus_decrypted to omit keys in corpus. If the set of keys allowed to be omitted is known, can those be explicitly removed from `corpus` and then an exact comparison performed between the adjusted `corpus` and `corpus_decrypted`?

spec/integration/client_side_encryption/corpus_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+        end
+
+        corpus_encrypted_actual = client
+          .use(:db)[:coll]
+          .find(_id: corpus_encrypted_id)
+          .first
+
+        # Check that the actual encrypted document matches the expected
+        # encrypted document.
+        corpus_encrypted_actual.each do |key, value|
```
p-mongo: Can the set of keys in `corpus_encrypted` and `corpus` be also asserted to equal?

spec/integration/client_side_encryption/corpus_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+              secret_access_key: SpecConfig.instance.fle_aws_secret,
+            }
+          },
+          key_vault_namespace: 'admin.datakeys',
+        },
+      )
+    end
+
+    let(:corpus) do
+      BSON::ExtJSON.parse(File.read('spec/support/crypt/corpus/corpus.json'), mode: :bson)
```
p-mongo: Can corpus be transformed here with code similar to

<pre>

corpus.each do |key, doc|
  if doc['value'].is_a?(Time)
    doc['value'] = doc['value'].utc
  end
end
</pre>

and a link to https://jira.mongodb.org/browse/RUBY-2158 ?

spec/integration/client_side_encryption/corpus_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+        corpus_decrypted = client_encrypted[:coll]
+          .find(_id: corpus_encrypted_id)
+          .first
+
+        # Ensure that corpus_decrypted is the same as the original corpus
+        # document by checking that they have the same set of keys, and that
+        # they have the same values at those keys (improved diagnostics).
+        expect(corpus_decrypted.keys - corpus.keys).to be_empty
+
+        corpus_decrypted.each do |key, doc|
```
p-mongo: If possible remove this block and transform corpus when it is loaded.

spec/integration/client_side_encryption/corpus_spec.rb
p-mongo: Please disregard for now.

spec/integration/client_side_encryption/corpus_spec.rb
p-mongo: Please disregard for now.

spec/integration/client_side_encryption/corpus_spec.rb
p-mongo: Let's leave this the way it is for now.

PR #1769, opened 2020-03-07
Fix RUBY-2144 Redundant socket connections during address resolution

p-mongo:
https://jira.mongodb.com/browse/RUBY-2144


p-mongo:
https://jira.mongodb.org/browse/RUBY-2144

PR #1774, opened 2020-03-09
RUBY-1997 Handle absence of 'ns' field in index specifications returned from listIndexes


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-1997

PR #1776, opened 2020-03-10
RUBY-2172 Improve error reporting when nil and empty string are given to URI#get


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2172

PR #1785, opened 2020-03-13
RUBY-2175 Change all instances of "FLE/Field-Level Encryption" to say "Client-Side Encryption"


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2175

PR #1790, opened 2020-03-14
Local kerberos implementation

p-mongo:


p-mongo:
The provisioner in this PR appears to be working. The remaining work to be done later is to integrate this provisioner with our test suite, which I expect to be primarily through 1) exporting environment variables that Evergreen project sets for testing against the kerberos server there and 2) documenting where to find these (e.g. output of the running local kerberos container) and how to pass these to the test suite (e.g. the private variable mechanism recently added).

PR #1793, opened 2020-03-16
DOCSP-9414: Replace Google Groups with MongoDB Community Forum

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-9414

Please advise on what should be done with the gemspec that references the google group as the mailing list. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,8 @@ Support & Feedback
 For issues, questions or feedback related to the Ruby driver, please look into
 our [support channels](http://www.mongodb.org/about/support). Please
 do not email any of the Ruby developers directly with issues or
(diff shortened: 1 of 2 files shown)


ccho-mongodb:
@p-mongo please review when you get the chance, thanks!

p-mongo:
According to https://guides.rubygems.org/specification-reference/ the email attribute is optional, so I would say it can be removed. The gem appears to build without it.

ccho-mongodb:
Removed references from the gemspec. I'm not sure why the evergreen checks are failing, but I don't have write permissions, so can you merge this in if everything is good? Thanks.

PR #1795, opened 2020-03-17
RUBY-2147 Run FLE tests on sharded cluster

egiurleo:
This PR allows a developer to specify a non-default port on which to run mongocryptd through an environment variable. Evergreen has mongocryptd listening on port 27090, allowing us to run tests on sharded clusters. This option could eventually be replaced with a URI option.

### DIFF
--- a/spec/README.md
+++ b/spec/README.md
@@ -374,6 +374,21 @@ environment variable to the full path to the .so/.dll/.dylib
 LIBMONGOCRYPT_PATH=/path/to/your/libmongocrypt/nocrypto/libmongocrypt.so bundle exec rake
 ```
 
+If you would like to run the client-side encryption tests on a replica set or
+sharded cluster, be aware that the driver will try to spawn the mongocryptd daemon on
+port 27020 by default. If port 27020 is already in use by a mongod or mongos
+process, spawning mongocryptd will fail, causing the tests to fail as well.
+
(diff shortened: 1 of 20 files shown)


spec/support/spec_config.rb
```diff
@@ -328,6 +328,10 @@ def fle_aws_arn
     ENV['MONGO_RUBY_DRIVER_AWS_ARN']
   end
 
+  def mongocryptd_port
+    ENV['MONGO_RUBY_DRIVER_MONGOCRYPTD_PORT'] || 27020
```
p-mongo: `ENV['MONGO_RUBY_DRIVER_MONGOCRYPTD_PORT']` is a string if provided but the default value is an integer.

spec/README.md
```diff
@@ -288,6 +288,15 @@ full list of relevant test files.
 
 ## Client-Side Encryption
 
+NOTE: Client-side encryption tests require an enterprise build of MongoDB
+server version 4.2 or higher. These builds of the MongoDB server come packaged with
+mongocryptd, a daemon that is spawned by the driver during automatic encryption.
+The client-side encryption tests require the mongocryptd binary to be in the
+system path.
```
egiurleo: @p-mongo Mention that the texts expect the mongocryptd binary to be in the system path

PR #1811, opened 2020-03-23
Fix Ruby 2.3 Replica Set Errors

egiurleo:
This error was surfacing in the FLE replica set tests, but that's just the only place we test 4.2 replica sets against ruby-2.3.

### DIFF
--- a/lib/mongo/error/bulk_write_error.rb
+++ b/lib/mongo/error/bulk_write_error.rb
@@ -35,6 +35,7 @@ module Mongo
       # @since 2.0.0
       def initialize(result)
         @result = result
+        super
       end
 
       def to_s
--- a/lib/mongo/error/bulk_write_error.rb
+++ b/lib/mongo/error/bulk_write_error.rb
@@ -35,7 +35,9 @@ module Mongo
       # @since 2.0.0
       def initialize(result)
         @result = result
-        super("#{self.class}: #{build_message}")
+        # Exception constructor behaves differently for a nil argument and
+        # for no argument. Avoid passing nil explicitly.
+        super(*[build_message])
       end
 
       private
--- a/lib/mongo/error/bulk_write_error.rb
+++ b/lib/mongo/error/bulk_write_error.rb
@@ -35,23 +35,22 @@ module Mongo
       # @since 2.0.0
       def initialize(result)
         @result = result
-        super()
+        super("#{self.class}: #{build_message}")
       end
 
-      def to_s
-        messages = if errors = result['writeErrors']
-          frag = ': ' + errors[0..10].map do |error|
-            "#{error['errmsg']} (#{error['code']})"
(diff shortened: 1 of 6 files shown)


lib/mongo/error/notable.rb
```diff
@@ -47,7 +47,7 @@ def add_note(note)
 
       # @api public
       def message
-        super + notes_tail
```
egiurleo: below, `to_s` is defined as:

```
def to_s
  super + notes_tail
end
```

`message` actually calls `to_s`, so this was causing `notes_tail` to be appended to the message twice.

lib/mongo/error/bulk_write_error.rb
```diff
@@ -35,22 +35,22 @@ class BulkWriteError < Error
...
-          end
-        else
-          ''
-        end
-        "#{self.class}: #{messages}" + notes_tail
+      private
+
+      def build_message
+        errors = @result['writeErrors']
+        return '' unless errors
```
p-mongo: I think this should now be `return nil unless errors` since this returns the full message and if there isn't a message, this should produce nil rather than the empty string.

spec/mongo/error/bulk_write_error_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  describe '#message' do
+    it 'returns the formatted message' do
+      expect(error.message).to eq("#{described_class}: #{messages}#{notes_tail}")
+    end
+  end
+
+  describe '#to_s' do
+    it 'returns the error represented as a string' do
+      expect(error.to_s).to eq("#{described_class}: #{messages}#{notes_tail}")
```
p-mongo: I tried looking at the error object by itself:

<pre>
[43, 52] in /home/w/apps/ruby-driver/spec/mongo/error/bulk_write_error_spec.rb
   43:     end
   44:   end
   45: 
   46:   describe '#to_s' do
   47:     it 'returns the error represented as a string' do
   48:     byebug
=> 49:       expect(error.to_s).to eq("#{described_class}: #{messages}#{notes_tail}")
   50:     end
   51:   end
   52: end
(byebug) error
#&lt;Mongo::Error::BulkWriteError: Mongo::Error::BulkWriteError: message1 (1), message2 (2) (note1, note2) (note1, note2)>
(byebug) error.to_s
"Mongo::Error::BulkWriteError: message1 (1), message2 (2) (note1, note2)"
(byebug) 

</pre>

lib/mongo/error/bulk_write_error.rb
```diff
@@ -35,22 +35,24 @@ class BulkWriteError < Error
...
+
+        fragment = errors.first(10).map do |error|
```
p-mongo: Did not know about `first(n)`, cool.

PR #1821, opened 2020-03-25
Fix command started event matching

egiurleo:


egiurleo:
I'm holding off on merging this for now because it revealed some inconsistencies in the FLE spec tests that I'm discussing with Kevin.

PR #1829, opened 2020-03-28
RUBY-1999 test assume role temporary credentials


p-mongo:


ShaneHarvey:
At first glance I would expect this PR to add the "AssumeRole" spec test: https://github.com/mongodb/specifications/blob/master/source/auth/tests/mongodb-aws.rst#4assumerole

But I don't think that's the case because it's not using the aws_e2e_assume_role.js script. Can you explain what this change is testing?

p-mongo:
This PR verifies that the driver is able to authenticate with temporary credentials (obtained from an assume role request, but could potentially be any temporary credentials) that are provided in the environment. 

The credentials are obtained via the tooling described in https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/README.aws-auth.md#temporary-credentials---assumerole-request. It performs the function of the `aws_e2e_assume_role.js ` script but is also usable in other scenarios, for example it can be used in a local Docker container to obtain the temporary credentials.

ShaneHarvey:
Can you point me to the MONGODB-AWS auth code (the code that runs `{saslStart: 1, mechanism: 'MONGODB-AWS', ...}`)?

p-mongo:
That is a couple of PRs out.

p-mongo:
So the authentication functionality that is currently in master is the request class which can make STS requests and sign requests for the server. There isn't a way to authenticate to the server yet.

PR #1833, opened 2020-03-29
RUBY-1999 retrieval of aws credentials from ec2 instance link local endpoint


p-mongo:


lib/mongo/auth/aws/credentials_retriever.rb
```diff
@@ -0,0 +1,181 @@
...
+      # partial credentials, credential retrieval fails with an exception.
+      #
+      # @api private
+      class CredentialsRetriever
+
+        # Timeout for metadata operations, in seconds.
+        #
+        # The auth spec suggests a 10 second timeout but this seems
+        # excessively long given that the endpoint is essentially local.
+        METADATA_TIMEOUT = 5
```
ShaneHarvey, line 46: We should stick to the spec and use the 10 second timeout for consistency across drivers. Feel free to open a drivers ticket to reduce the timeout to 5 seconds.

lib/mongo/auth/aws/credentials_retriever.rb
```diff
@@ -0,0 +1,181 @@
...
+          end
+          metadata_token = resp.body
+          resp = ::Timeout.timeout(METADATA_TIMEOUT) do
+            http_get(http, '/latest/meta-data/iam/security-credentials', metadata_token)
+          end
+          if resp.code != '200'
+            return nil
+          end
+          role_name = resp.body
+          escaped_role_name = CGI.escape(role_name).gsub('+', '%20')
```
ShaneHarvey, line 122: Should this use `URI::encode`?

lib/mongo/auth/aws/request.rb
```diff
@@ -210,6 +213,38 @@ def authorization
           "AWS4-HMAC-SHA256 Credential=#{access_key_id}/#{scope}, SignedHeaders=#{signed_headers_string}, Signature=#{signature}"
         end
 
+        # Validates the credentials and the constructed request components
+        # by sending a real STS GetCallerIdentity request.
```
ShaneHarvey, line 217: This is interesting because this validation gives a much better error message to the user besides "authentication failed" but it is not part of the auth spec. I would feel more comfortable for someone on the security team to review this. If this is useful in Ruby perhaps we should open a drivers ticket to make all drivers perform this check.

lib/mongo/auth/aws/credentials_retriever.rb
p-mongo: The spec suggests a 10 second timeout but does not require it.

p-mongo: Created https://jira.mongodb.org/browse/DRIVERS-994.

lib/mongo/auth/aws/credentials_retriever.rb
p-mongo: `URI::encode` is deprecated and the Ruby driver no longer uses it.

lib/mongo/auth/aws/request.rb
p-mongo: This validation is currently only used in the test suite. I intended to add a debugging option to the driver to make it perform this check prior to authenticating, but since this involves additional network roundtrips, and because it is possible for the application to not have network access to STS due to firewall rules (but the server would have STS access), I did not intend to have this check be performed by default. I can create a drivers ticket regarding this option if you like once aws auth implementation in Ruby is complete.

lib/mongo/auth/aws/credentials_retriever.rb
ShaneHarvey: In our specs "SHOULD" usually means "MUST unless it's impossible or overly complex to implement in the lanaguage/driver". But reading the rationale in the ticket I agree we should reduce it to 5, thanks for opening.

lib/mongo/auth/aws/credentials_retriever.rb
ShaneHarvey: Oh sorry. I missed the `CGI.escape` and thought the code was _only_ escaping `+`. Please ignore.

lib/mongo/auth/aws/request.rb
ShaneHarvey: Thanks for explaining. In that case, this LGTM. I'll leave the decision of whether to open a drivers ticket up to you.

PR #1834, opened 2020-03-30
RUBY-1972 Support the hint option in update, update_many, and replace_one commands

egiurleo:


lib/mongo/collection/view/writable.rb
```diff
@@ -233,12 +236,23 @@ def replace_one(replacement, opts = {})
...
+
+              if server.max_wire_version < 5 && update_doc['hint']
+                raise Error::UnsupportedCommandOption.new(
+                  option_name: 'hint',
+                  command_name: 'replace_one',
+                  required_server_version: '3.4',
+                  required_wire_version: '5',
+                  server_max_wire_version: server.max_wire_version
+                )
+              end
```
egiurleo: This code is not very DRY, but the operations layer isn't very DRY in general. Would it be preferable for me to try and abstract this logic into a helper module?

spec/mongo/bulk_write_spec.rb
```diff
@@ -348,6 +348,30 @@
...
+                it 'raises a client error' do
+                  expect do
+                    bulk_write.execute
+                  end.to raise_error(
+                    Mongo::Error::UnsupportedCommandOption,
+                    /The hint option is only supported for bulk write with update_one commands on MongoDB server versions 3.4 and newer \(wire version 5 and greater\)/
+                  )
+                end
+              end
+            end
```
egiurleo: I have only included Ruby tests for the case where the client throws an error to verify the error message. Is there an easy way for me to test that passing the hint option to the server actually worked? Is that necessary?

p-mongo: To answer the first question, I believe such a thing is accomplished elsewhere in the test suite by setting up command monitoring and inspecting the command sent.

lib/mongo/bulk_write.rb
```diff
@@ -217,14 +217,40 @@ def insert_one(documents, server, operation_id, session, txn_num)
     end
 
     def update_one(documents, server, operation_id, session, txn_num)
+      if server.max_wire_version < 5 && hint_provided?(documents)
+        raise Error::UnsupportedCommandOption.new(
+          option_name: 'hint',
+          command_name: "bulk write with #{__callee__}",
```
p-mongo: My web search for "ruby __callee__" produced no results, can you please link to documentation for it?

egiurleo: https://apidock.com/ruby/Kernel/__callee__

PR #1837, opened 2020-03-31
RUBY-2177 Fix all API documentation warnings

egiurleo:
The culprit was mostly lib/mongo/crypt/binding.rb. Because most of the methods in that file are metaprogrammed, I had to use the `@!method` tag, which I hadn't known about before!

### DIFF
--- a/lib/mongo/crypt/binding.rb
+++ b/lib/mongo/crypt/binding.rb
@@ -56,55 +56,50 @@ module Mongo
           "is invalid: #{ENV['LIBMONGOCRYPT_PATH']}\n\n#{e.class}: #{e.message}"
       end
 
-      # Returns the version string of the libmongocrypt library
-      #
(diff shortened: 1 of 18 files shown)


lib/mongo/crypt/binding.rb
```diff
@@ -185,31 +174,30 @@ class Binding
       # A callback to the mongocrypt log function
       # Set a custom log callback with the mongocrypt_setopt_log_handler method
       #
-      # @param [ Symbol ] level The log level; possible values defined by the
+      # param [ Symbol ] level The log level; possible values defined by the
```
p-mongo: I think it is beneficial to have the callbacks' api docs also be rendered, since all of the work to document them has already been done. How about adding a note either per method or one time in the class docstring to the effect of "the entries ending in `_fn` are FFI callback definitions and not methods defined on this class"?

egiurleo: Sounds good!

PR #1840, opened 2020-04-01
RUBY-2149 Do not repeatedly resume if getMore receives the same error

p-mongo:
https://jira.mongodb.com/browse/RUBY-2149


p-mongo:
https://jira.mongodb.org/browse/RUBY-2149

p-mongo:
Our change stream tests are rather hard to read. I separated some out as otherwise the scopes were too confusing, unfortunately it probably makes the diff even less readable.

Major changes are:

- New resume logic: on 4.4+ we go by the error labels only, on pre-4.4 we go by the codes and use a whitelist. Previously we were using a blacklist and also matched by exception message, both of these strategies are now gone.
- To implement this I added the description of the connection used for a request as an attribute on the `Mongo::Operation::Result`. This is populated when the result is created, since we have the connection available there. We set the description and not the connection to not interfere with connection accounting/scopes/lifetimes.
- I would have made the description required but unfortunately mongo_kerberos copies some code that creates Result instances with the single argument. Therefore the description is optional but everything in the driver proper should be setting it.
- Killed cursors can no longer be recovered from, tests now assert that they raise errors.
- Fail point support added to change stream spec runner, since some of the new spec tests use fail points. It seems that we had existing tests that used fail points that somehow were passing even though we weren't setting any fail points.

egiurleo:
Thanks for getting the server description from the connection too!

PR #1843, opened 2020-04-02
RUBY-2196 Move Kerberos auth mechanism and conversation to driver

p-mongo:
This PR removes JRuby/Kerberos configurations until the JRuby native gssapi implementation is unbroken.


p-mongo:
https://jira.mongodb.org/browse/RUBY-2196

PR #1844, opened 2020-04-03
RUBY-2176 Raise an error if an encryption client is initialized with empty string AWS credentials

egiurleo:
When testing this change, I realized that the client would leak background threads if it raised an error while building the encrypter, so I added a fix and tests for that as well.

### DIFF
--- a/lib/mongo/client.rb
+++ b/lib/mongo/client.rb
@@ -484,7 +484,12 @@ module Mongo
 
       if @options[:auto_encryption_options]
         @connect_lock.synchronize do
-          build_encrypter
+          begin
+            build_encrypter
+          rescue
+            @cluster.disconnect!
+            raise
+          end
         end
       end
 
@@ -906,14 +911,9 @@ module Mongo
 
     # Create a new encrypter object using the client's auto encryption options
     def build_encrypter
-      begin
-        @encrypter = Crypt::AutoEncrypter.new(
-          @options[:auto_encryption_options].merge(client: self)
-        )
-      rescue
-        @cluster.disconnect!
-        raise
-      end
+      @encrypter = Crypt::AutoEncrypter.new(
+        @options[:auto_encryption_options].merge(client: self)
+      )
     end
 
     # Generate default client options based on the URI and options
--- a/lib/mongo/client.rb
+++ b/lib/mongo/client.rb
@@ -699,7 +699,9 @@ module Mongo
         # If auto_encryption_options are nil, set @encrypter to nil, but do not
         # close the encrypter because it may still be used by the original client.
         if @options[:auto_encryption_options] && auto_encryption_options_changed
-          build_encrypter
+          @connect_lock.synchronize do
+            build_encrypter
+          end
         elsif @options[:auto_encryption_options].nil?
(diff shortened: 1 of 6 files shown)


lib/mongo/client.rb
```diff
@@ -699,7 +699,9 @@ def update_options(new_options)
         # If auto_encryption_options are nil, set @encrypter to nil, but do not
         # close the encrypter because it may still be used by the original client.
         if @options[:auto_encryption_options] && auto_encryption_options_changed
-          build_encrypter
+          @connect_lock.synchronize do
+            build_encrypter
+          end
         elsif @options[:auto_encryption_options].nil?
           @encrypter = nil
```
p-mongo, line 706: This assignment should also be under the connect lock I think.

lib/mongo/client.rb
```diff
@@ -902,9 +904,14 @@ def watch(pipeline = [], options = {})
...
     def build_encrypter
-      @encrypter = Crypt::AutoEncrypter.new(
-        @options[:auto_encryption_options].merge(client: self)
-      )
+      begin
+        @encrypter = Crypt::AutoEncrypter.new(
+          @options[:auto_encryption_options].merge(client: self)
+        )
+      rescue => e
+        do_close
```
p-mongo, line 912: I am confused as to what this call is intended to do:

1. It also disconnects the cluster, why does a failure to create encrypter cause the cluster to be disconnected?
2. I don't think it would do anything as far as closing the encrypter if `@encrypter` is not assigned, therefore, if there's an exception raised in the block above this call wouldn't actually perform any encrypter cleanup?

What I think should happen is the `AutoEncrypter` constructor needs to have the begin-rescue block in it around `EncryptionIO` instantiation (or more precisely everything after the `Client` is constructed there) and if any exception is raised, the client that was made should be closed.

lib/mongo/client.rb
```diff
@@ -902,9 +904,14 @@ def watch(pipeline = [], options = {})
...
+        raise e
```
p-mongo, line 913: You can call `raise` to reraise the exception being handled.

lib/mongo/client.rb
egiurleo: We talked over Slack and agreed that:
- The client should disconnect the cluster here
- There should also be a begin... rescue block in the Encrypter initializer that closes the mongocryptd client if encryption fails

lib/mongo/client.rb
```diff
@@ -902,9 +906,14 @@ def watch(pipeline = [], options = {})
...
     def build_encrypter
-      @encrypter = Crypt::AutoEncrypter.new(
-        @options[:auto_encryption_options].merge(client: self)
-      )
+      begin
+        @encrypter = Crypt::AutoEncrypter.new(
+          @options[:auto_encryption_options].merge(client: self)
+        )
+      rescue
+        @cluster.disconnect!
```
p-mongo, line 914: Can you move this to the client constructor please so that the construction and teardown are invoked from the same place?

egiurleo: Done!

PR #1849, opened 2020-04-05
RUBY-1999 credentials retrieval from ECS task metadata

p-mongo:

### DIFF
--- a/lib/mongo/auth/aws/credentials_retriever.rb
+++ b/lib/mongo/auth/aws/credentials_retriever.rb
@@ -83,6 +83,12 @@ module Mongo
             return credentials
           end
 
+          credentials = ecs_metadata_credentials
+
+          if credentials && credentials_valid?(credentials, 'ECS task metadata')
+            return credentials
+          end
+
           credentials = ec2_metadata_credentials
 
           if credentials && credentials_valid?(credentials, 'EC2 instance metadata')
@@ -90,7 +96,7 @@ module Mongo
           end
 
           raise Auth::InvalidConfiguration,
-            "Could not locate AWS credentials (checked Client URI and Ruby options, environment variables, EC2 metadata)"
+            "Could not locate AWS credentials (checked Client URI and Ruby options, environment variables, ECS and EC2 metadata)"
         end
 
         private
@@ -135,7 +141,39 @@ module Mongo
             payload['SecretAccessKey'],
             payload['Token'],
           )
-        rescue ::Timeout::Error
+        # When trying to use the EC2 metadata endpoint on ECS:
+        # Errno::EINVAL: Failed to open TCP connection to 169.254.169.254:80 (Invalid argument - connect(2) for "169.254.169.254" port 80)
+        rescue ::Timeout::Error, IOError, Errno::EINVAL
+          return nil
+        end
+
+        def ecs_metadata_credentials
+          relative_uri = ENV['AWS_CONTAINER_CREDENTIALS_RELATIVE_URI']
+          if relative_uri.nil? || relative_uri.empty?
+            return nil
+          end
+
+          http = Net::HTTP.new('169.254.170.2')
+          # Per https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html
+          # the value in AWS_CONTAINER_CREDENTIALS_RELATIVE_URI includes
+          # the leading slash.
+          # The current language in MONGODB-AWS specification implies that
+          # a leading slash must be added by the driver, but this is not
+          # in fact needed.
+          req = Net::HTTP::Get.new(relative_uri)
+          resp = ::Timeout.timeout(METADATA_TIMEOUT) do
+            http.request(req)
+          end
+          if resp.code != '200'
+            return nil
+          end
+          payload = JSON.parse(resp.body)
+          Credentials.new(
+            payload['AccessKeyId'],
+            payload['SecretAccessKey'],
+            payload['Token'],
+          )
+        rescue ::Timeout::Error, IOError, Errno::EINVAL
           return nil
         end
 
--- a/spec/support/aws_utils/orchestrator.rb
+++ b/spec/support/aws_utils/orchestrator.rb
@@ -251,6 +251,56 @@ CMD
       end
     end
 
+    def wait_for_ecs_ready(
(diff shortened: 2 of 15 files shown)


lib/mongo/auth/aws/credentials_retriever.rb
```diff
@@ -83,14 +83,20 @@ def credentials
             return credentials
           end
 
+          credentials = ecs_metadata_credentials
```
ShaneHarvey, line 86: It looks like `ecs_metadata_credentials` returns nil whenever the http request fails and in that case the driver will fallback to `ec2_metadata_credentials`. But from my reading of the ["Obtaining Credentials" section of the spec](https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#obtaining-credentials), the driver should only ever attempt one of these. In other words, I think the logic should be:
```
if ENV['AWS_CONTAINER_CREDENTIALS_RELATIVE_URI']
    credentials = ecs_metadata_credentials
else
    credentials = ec2_metadata_credentials
fi
```

spec/support/aws_utils/orchestrator.rb
```diff
@@ -251,6 +251,56 @@ def provision_auth_ecs_task(public_key_path: nil,
...
+      ecs_client.wait_until(
+        :services_stable, {
+          cluster: cluster_name,
+          services: [service_name],
+        },
+        delay: 5,
+        max_attempts: 36,
+      )
+    end
+
```
ShaneHarvey, line 303: Why are we implementing these `aws_utils`? All other drivers (and the server) are using the drivers-evergreen-tools scripts to test MONGODB-AWS.

lib/mongo/auth/aws/credentials_retriever.rb
p-mongo: Looking at the auth spec again it seems to mandate both behaviors. I filed https://jira.mongodb.org/browse/DRIVERS-995 to clarify the situation.

The behavior implemented in this PR supports all possible deployment scenarios, including ECS tasks authenticating with instance role credentials.

lib/mongo/auth/aws/credentials_retriever.rb
```diff
@@ -83,14 +83,20 @@ def credentials
...
+
           credentials = ec2_metadata_credentials
 
           if credentials && credentials_valid?(credentials, 'EC2 instance metadata')
             return credentials
           end
 
           raise Auth::InvalidConfiguration,
-            "Could not locate AWS credentials (checked Client URI and Ruby options, environment variables, EC2 metadata)"
+            "Could not locate AWS credentials (checked Client URI and Ruby options, environment variables, ECS and EC2 metadata)"
```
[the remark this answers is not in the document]
p-mongo: I have not needed this diagnostics yet but if this changes or there is a user request for such diagnostics I'm happy to add it.

lib/mongo/auth/aws/credentials_retriever.rb
```diff
@@ -135,7 +141,39 @@ def ec2_metadata_credentials
...
+          # a leading slash must be added by the driver, but this is not
+          # in fact needed.
+          req = Net::HTTP::Get.new(relative_uri)
+          resp = ::Timeout.timeout(METADATA_TIMEOUT) do
+            http.request(req)
+          end
+          if resp.code != '200'
+            return nil
+          end
+          payload = JSON.parse(resp.body)
```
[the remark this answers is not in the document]
p-mongo: Yes, a sample response returned is provided in https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#obtaining-credentials and it does not include the code.

spec/support/aws_utils/orchestrator.rb
p-mongo: The tooling in the Ruby driver is much more extensive, flexible, modular and integrated than d-e-t. The difference between the two is similar to the difference in testing code locally and sending it to Evergreen for every change, for Evergreen to run the entire test suite and report that there is a syntax error in a file somewhere. It works but is not efficient.

For example, when the driver tooling is used to provision the cluster and task definition, it sets up CloudWatch logs for the task, such that any provisioning errors are can be quickly identified. Subsequently in  cluster so provisioned it is possible to run the tests (following instructions added in this PR under "without rebuilding the remote environment") with about 5 second turnaround time. The minimum turnaround time with Evergreen I've seen is about 3 minutes, and 5-10 minutes is more common if you consider  the number of clicks and time needed to get to the build, the logs, find the problem in the logs, etc.. This is a massive productivity difference when developing.

Thinking back to the issue I ran into with apt prompting for input in the ecs task, I am under the impression that the input prompt was not even shown in the Evergreen logs. Troubleshooting this via Evergreen would've taken an inordinate amount of time or luck. The driver tooling permits running all tests interactively and exposed this issue immediately.

To give another example, the instance role tests perform credential retrieval with the instance profile assigned to the instance as well as when the instance profile is not assigned, and verify that both scenarios work as expected.

lib/mongo/auth/aws/credentials_retriever.rb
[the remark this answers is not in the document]
ShaneHarvey: SGTM

lib/mongo/auth/aws/credentials_retriever.rb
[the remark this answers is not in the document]
ShaneHarvey: Thanks!

lib/mongo/auth/aws/credentials_retriever.rb
ShaneHarvey: I believe all other drivers have implemented the order I described above so Ruby should probably follow suit for consistency. 

> including ECS tasks authenticating with instance role credentials.

I'm not familiar with this case. Can you explain what the credential lookup flow would be and why my above suggest would not support it?

spec/support/aws_utils/orchestrator.rb
ShaneHarvey: Thanks for the detailed response. I'm convinced these scripts have value. I hope we can use these to benefit other drivers/implementors in the future.

lib/mongo/auth/aws/credentials_retriever.rb
ShaneHarvey: On second thought, I don't want to hold up your work. We can continue this discussion in Jira.

PR #1852, opened 2020-04-07
RUBY-2119 Do not lose type information when writing documents with automatic encryption

egiurleo:
**The Problem:** During automatic encryption, commands are sent to mongocryptd to have encryption placeholders added to them. The response from mongocryptd is deserialized without the `mode: :bson` flag, causing command fields that were previously BSON types (e.g. Int64s) to be deserialized into Ruby types (e.g. Ruby integers). This means that the driver can lose type information on non-encrypted fields when writing them to the database using automatic encryption.

**The Fix:** Allow commands to pass an option called `deserialize_as_bson` to the execution code and down to the protocol layer, which will then deserialize responses from the server using the `mode: :bson` option. This isn't included on all commands because it would create a significant change in driver behavior, so it is only used on the command that is sent to mongocryptd.

### DIFF
--- a/lib/mongo/database.rb
+++ b/lib/mongo/database.rb
@@ -171,19 +171,19 @@ module Mongo
     #
     # @return [ Mongo::Operation::Result ] The result of the command execution.
     def command(operation, opts = {})
-      opts_copy = opts.dup
-      execution_opts = opts_copy.delete(:execution_options) || {}
+      opts = opts.dup
+      execution_opts = opts.delete(:execution_options) || {}
 
-      txn_read_pref = if opts_copy[:session] && opts_copy[:session].in_transaction?
-        opts_copy[:session].txn_read_preference
+      txn_read_pref = if opts[:session] && opts[:session].in_transaction?
+        opts[:session].txn_read_preference
       else
         nil
       end
-      txn_read_pref ||= opts_copy[:read] || ServerSelector::PRIMARY
+      txn_read_pref ||= opts[:read] || ServerSelector::PRIMARY
       Lint.validate_underscore_read_preference(txn_read_pref)
       selector = ServerSelector.get(txn_read_pref)
 
-      client.send(:with_session, opts_copy) do |session|
+      client.send(:with_session, opts) do |session|
         server = selector.select_server(cluster, nil, session)
         op = Operation::Command.new(
           :selector => operation.dup,
--- a/lib/mongo/database.rb
+++ b/lib/mongo/database.rb
@@ -182,7 +182,9 @@ module Mongo
           :read => selector,
           :session => session
         )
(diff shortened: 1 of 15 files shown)


lib/mongo/protocol/serializers.rb
```diff
@@ -265,7 +265,7 @@ def self.serialize(buffer, value, max_bson_size = nil, validating_keys = BSON::C
           #
           # @since 2.5.0
           def self.deserialize(buffer)
-            BSON::Document.from_bson(buffer)
+            BSON::Document.from_bson(buffer, mode: :bson)
```
egiurleo, line 268: The problem was that we were sending commands to mongocryptd, which adds encryption placeholders, and then we were deserializing the responses without `mode: :bson`, so any int64s were coming back as regular integers.

lib/mongo/database.rb
```diff
@@ -162,9 +162,17 @@ def collections
...
+    # @option opts :execution_options [ Hash ] Options to pass to the code that
+    #   executes this command. This is an internal option and is subject to
+    #   change.
+    #   - :deserialize_as_bson [ Boolean ] Whether to deserialize the response
+    #     to this command using BSON types intead of native Ruby types wherever
+    #     possible.
     #
     # @return [ Mongo::Operation::Result ] The result of the command execution.
     def command(operation, opts = {})
+      execution_opts = opts.delete(:execution_options) || {}
```
p-mongo, line 174: `opts` need to be duplicated before being mutated.

spec/integration/crud_spec.rb
```diff
@@ -7,6 +7,83 @@
...
   end
 
+  describe 'insert' do
+    context 'inserting a BSON::Int64 or BSON::Int32' do
+      before do
+        collection.insert_one(int64: BSON::Int64.new(42))
+      end
+
+      it 'is stored as the correct type' do
+        result = collection.find(int64: { '$type' => 18 }).first
```
p-mongo, line 17: I believe `'$type' => 'long'` would work here and be more clear.

p-mongo: Per the docs at https://docs.mongodb.com/manual/reference/operator/query/type/ and https://docs.mongodb.com/manual/reference/operator/query/type/#document-type-available-types.

lib/mongo/database.rb
```diff
@@ -171,18 +171,19 @@ def collections
     #
     # @return [ Mongo::Operation::Result ] The result of the command execution.
     def command(operation, opts = {})
-      execution_opts = opts.delete(:execution_options) || {}
+      opts_copy = opts.dup
```
p-mongo, line 174: This can simply be `opts = opts.dup` since generally the original argument won't be used past this point.

PR #1858, opened 2020-04-10
Fix RUBY-2210 GridFS does not write data correctly when given String in UTF-8 encoding


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2210

PR #1863, opened 2020-04-13
RUBY-2110 [wip]

egiurleo:
This is a big PR, but a lot of the changes are superficial, so here's how I would read it:
* Start at `lib/mongo/retryable.rb`. This is where the bulk of the functionality change is introduced. Now, all retryable methods yield a connection rather than a server.
* Read `lib/mongo/operation/shared/executable.rb` and note that operations are now executed with a connection rather than a server.
* grep `server.with_connection` and see all the places where I've started calling `execute` with a connection.
* Glance through the other changes -- I've replaced the variable `server` with `connection` everywhere that they were switched over.
* Read the spec changes.

Remaining questions:
* Do I have to change anything relating to server selection? The features and other server description attributes are all read off of the server in that part of the code base.
* Can I have some help fixing `spec/integration/server_spec.rb:41`?
* I'd love some guidance on what to do with the commented out test in `spec/integration/retryable_writes_spec.rb`.


spec/integration/server_spec.rb
```diff
@@ -26,23 +28,29 @@
...
+    #
+    #   it 'can be used for reads' do
+    #     result = server.with_connection do |connection|
+    #       view.send(:send_initial_query, connection)
+    #     end
+    #
+    #     # Driver falls back to the oldest MongoDB protocol
+    #     expect(result).to be_a(Mongo::Operation::Find::Legacy::Result)
+    #   end
+    # end
```
egiurleo, line 54: I could use some help figuring out why this test is failing.

spec/integration/retryable_writes_spec.rb
```diff
@@ -217,11 +219,11 @@
...
-            it 'indicates server used for operation' do
-              expect do
-                operation
-              end.to raise_error(Mongo::Error, /on #{ClusterConfig.instance.primary_address_str}/)
-            end
+            # it 'indicates server used for operation' do
+            #   expect do
+            #     operation
+            #   end.to raise_error(r, /on #{ClusterConfig.instance.primary_address_str}/)
+            # end
```
egiurleo, line 226: This test no longer passes because `server.with_connection` is no longer called from within a `add_server_diagnostics` block (from [response_handling.rb](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/operation/shared/response_handling.rb#L89)). I'm not sure if that's bad, or if there is a way to re-write this test so it accomplishes its original goal.

lib/mongo/session.rb
```diff
@@ -676,6 +676,7 @@ def pin(server)
         raise ArgumentError, 'Cannot pin to a nil server'
       end
       if Lint.enabled?
+        # TODO: does any work need to be done here?
         unless server.mongos?
```
egiurleo, line 680: This relates to my question in the PR description -- what should I do about server selection?

lib/mongo/server.rb
```diff
@@ -424,6 +444,25 @@ def with_connection(&block)
...
+
+    # Checks the connection back into the server's connection pool.
+    #
+    # @param [ Mongo::Server::Connection ] connection The connection to check
+    #   into the pool.
+    #
+    # @api private
+    def check_in_connection(connection)
+      pool.check_in(connection)
+    end
```
egiurleo, line 464: See retryable.rb for more information on why I added this and how it's used.

lib/mongo/server.rb
```diff
@@ -161,9 +156,34 @@ def compressor
...
+    #
+    # @!method max_write_batch_size
+    #   @deprecated
+    def_delegators :description,
+     :features,
+     :logical_session_timeout,
+     :max_bson_object_size,
+     :max_message_size,
+     :max_wire_version,
+     :max_write_batch_size
```
egiurleo, line 185: These are all the delegated methods that are no longer used by this class. The ones I didn't move over here are still being used in server selection. See my question in the PR description -- should I change the server selection layer at all?

lib/mongo/retryable.rb
```diff
@@ -206,8 +208,10 @@ def write_with_retry(session, write_concern, ending_transaction = false, &block)
       # failed retry_write_allowed? check.
 
       server = select_server(cluster, ServerSelector.primary, session)
+      connection = server.check_out_connection
 
-      unless ending_transaction || server.retry_writes?
+      unless ending_transaction || connection.retry_writes?
+        server.check_in_connection(connection)
         return legacy_write_with_retry(server, session, &block)
```
egiurleo, line 215: I spent a lot of time thinking about the best way to do this. I wanted to avoid a scenario where `legacy_write_with_retry` is called from within a `server.with_connection` block, causing two connections to be checked out of the same server's pool. I also thought of using the block API but then using `break` or raising an error to get out of the block and then call `legacy_write_with_retry`, but when I wrote that code it was very hard to read. I'd be open to any guidance here.

PR #1864, opened 2020-04-14
RUBY-2218 Clarify how a driver must handle wrong set name in single topology


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2218

PR #1865, opened 2020-04-14
RUBY-2217 Add tests for allowed cases when authSource is given but username is not


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2217

PR #1867, opened 2020-04-15
RUBY-2110 Examine features on connection rather than server in operations layer

egiurleo:


lib/mongo/operation/collections_info/command.rb
```diff
@@ -0,0 +1,42 @@
...
+
+module Mongo
+  module Operation
+    class CollectionsInfo
+
+      # A MongoDB collectionInfo operation sent as a command message.
+      #
+      # @api private
+      #
+      # @since 2.5.2
```
p-mongo, line 23: Can we drop this @since since it's not accurate?

lib/mongo/operation/shared/collections_info_or_list_collections.rb
```diff
@@ -0,0 +1,58 @@
...
+      #
+      # @param [ Mongo::Server ] server The server to send the operation to.
+      # @param [ Mongo::Client ] client The client that will be used to
+      #   perform auto-encryption if it is necessary to encrypt the command
+      #   being executed (optional).
+      #
+      # @return [ Mongo::Operation::CollectionsInfo::Result,
+      #           Mongo::Operation::ListCollections::Result ] The operation result.
+      #
+      # @since 2.0.0
```
p-mongo, line 33: This @since is also not accurate.

spec/integration/server_spec.rb
```diff
@@ -41,7 +41,7 @@
       it 'can be used for reads' do
         result = view.send(:send_initial_query, server)
         # Driver falls back to the oldest MongoDB protocol
-        expect(result).to be_a(Mongo::Operation::Find::Legacy::Result)
+        expect(result).to be_a(Mongo::Operation::Find::Result)
```
p-mongo, line 44: This needs to be two tests, each conditioned on server version.

p-mongo:
evergreen retry

spec/integration/server_spec.rb
egiurleo: This spec only runs on server fcvs >= 3.4, so I don't believe this is a concern for this particular test.

p-mongo: Is something preventing it from running on older servers though or was that omitted because the test used to test new servers using the old protocol?

egiurleo: The latter. I can update this test to run on all servers and see that it uses the legacy response on old server versions if you think that would improve coverage.

PR #1868, opened 2020-04-15
RUBY-2219 Force read preference primaryPreferred in Single topology

p-mongo:
The tests have been completely reorganized and now test legacy and modern behaviors in all possible configurations.

Docstrings have been added to implementation methods.

User-visible changes: when on 3.6+ servers in direct connection (Single topology), we now always send read preference of primaryPreferred as mandated by the specs to allow for reads from secondaries. User-specified primary read preference in  this situation is overwritten to become primaryPreferred to allow for secondary reads.


p-mongo:
https://jira.mongodb.org/browse/RUBY-2219

PR #1871, opened 2020-04-16
RUBY-1999 user documentation for AWS auth mechanism


p-mongo:

### DIFF
--- a/docs/tutorials/ruby-driver-authentication.txt
+++ b/docs/tutorials/ruby-driver-authentication.txt
@@ -31,10 +31,11 @@ client:
 
   client = Mongo::Client.new([ '127.0.0.1:27017' ],
                              user: 'test',
-                             password: '123' )
+                             password: '123',
+                             database: 'mydb' )
 
   # If using a URI:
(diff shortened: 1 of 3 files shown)


p-mongo:
Rendered: https://p-mongo.github.io/rendered-docs/driver/tutorial/tutorials/ruby-driver-authentication.html#aws

docs/tutorials/ruby-driver-authentication.txt
```diff
@@ -179,6 +180,166 @@ see the :manual:`X.509 tutorial in the MongoDB Manual
                              ssl_ca_cert: '/path/to/ca.pem' )
 
 
+AWS
+```
+
+*Requires MongoDB Enterprise Edition and server version 4.4 or later.*
```
ShaneHarvey, line 186: Will this be available in Atlas? If so, I'd suggest adding "Requires MongoDB Atlas or ...".

p-mongo: We do mention Atlas in a couple of places in documentation but we do not call it out as supported enterprise features. Do you find that users perceive a lack of clarity as to what functionality can be used with Atlas?

PR #1874, opened 2020-04-18
Fix RUBY-2041 Ensure server proof has been validated during SCRAM conversation


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2041

PR #1876, opened 2020-04-18
RUBY-2002 Improve testing around default writeConcern


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2002

PR #1879, opened 2020-04-19
RUBY-1688 Update Transactions Retry Example 3 to include read preference


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-1688

PR #1880, opened 2020-04-19
RUBY-1908 Provide transaction example using new withTransaction API


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-1908

PR #1881, opened 2020-04-19
Fix RUBY-2032 Tutorial references nonexistent tls Ruby option


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2032

PR #1883, opened 2020-04-21
RUBY-2231 Rename handle_errors to map_exceptions


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2231

PR #1889, opened 2020-04-23
RUBY-2232 fix some note adding issues pointed out by linter that were missed in the previous commit


p-mongo:
https://jira.mongodb.com/browse/RUBY-2232


p-mongo:
In order to know *when* a particular exception needs to have diagnostics added to it, this PR changes the scope of which exceptions are handled by the retryable reads/writes code. Previously it treated non-Mongo::Error errors as possibly being created by the driver; this shouldn't generally be happening but can (for example, driver raises ArgumentError in some situations). It is not possible to add notes to exceptions that are not derived from Mongo::Error or AuthError, therefore error handling has been reduced in scope to only apply to errors that the driver is supposed to be surfacing.

PR #1900, opened 2020-04-29
RUBY-1737 Describe default auth mechanism selection and necessity to specify mongodb_cr explicitly


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-1737

PR #1904, opened 2020-05-05
RUBY-2044 Part II: Update legacy retries to use RetryableWriteError label

egiurleo:
This PR should not change user-facing behavior. I have added an integration test for legacy retryable writes because I could not find any integration tests for that behavior.

### DIFF
--- a/lib/mongo/operation/shared/response_handling.rb
+++ b/lib/mongo/operation/shared/response_handling.rb
@@ -134,7 +134,7 @@ module Mongo
         aborting_transaction = in_transaction && session.aborting_transaction?
         modern_retry_writes = client && client.options[:retry_writes]
         legacy_retry_writes = client && !client.options[:retry_writes] &&
-          client.options[:max_write_retries] != 0
+          client.max_write_retries > 0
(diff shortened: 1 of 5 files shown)


lib/mongo/operation/shared/response_handling.rb
```diff
@@ -132,7 +132,10 @@ def maybe_add_retryable_write_error_label!(error, client, session)
         in_transaction = session && session.in_transaction?
         committing_transaction = in_transaction && session.committing_transaction?
         aborting_transaction = in_transaction && session.aborting_transaction?
-        retry_writes = client && client.options[:retry_writes]
+        modern_retry_writes = client && client.options[:retry_writes]
+        legacy_retry_writes = client && !client.options[:retry_writes] &&
+          client.options[:max_write_retries] != 0
```
p-mongo, line 137: Can this be `> 0` please?

egiurleo: I think it should actually be `client.max_write_retries > 0` because the method assigns a default. Good call.

PR #1905, opened 2020-05-08
RUBY-2044 Part III: Create new integration tests for retryable writes

egiurleo:
There are a few problems with the existing retryable writes integration tests:
1. **They are coupled to implementation**: They use knowledge of the inner workings of sockets and server selection to raise errors at certain times. This makes them dependent on the implementation of retryable writes.
2. **They don't always reflect real-life scenarios**: Because of the previous point, they test scenarios that would never happen in real life, such as an OperationFailure being raised during a socket write or a socket error occurring during an unacknowledged write. This makes the tests hard to reason about.
3. **They are hard to read**: They test a lot of cases, which are all kept in one file.

Before adding new behavior to retryable writes, I wanted to attempt to create some new tests that cover the same functionality as our existing retryable write integration tests and solve some of the problems.

**My approach:**
* Create separate files for shared examples. I have one file with all of the commands that we want to test. This central file uses shared examples for supporting modern retries, supporting legacy retries, and supporting no retries, all in separate files. Each of those files also uses some shared examples for performing modern retries, performing legacy retries, and performing no retries.
* Use failPoint to simulate real-world scenarios. These tests require no knowledge of the retryable writes implementation. They use failPoints to mock server behavior and then ensure that the driver behaves properly in response.

**How this addresses the problems:**
1. Tests are no longer coupled to implementation.
2. Tests do not override driver behavior, which means we're not testing any impossible scenarios.
3. I believe these tests are easier to read and reason about.

**Limitations of these tests:**
* The failPoint command was only introduced in mongod 4.0 and mongos 4.2, so these can only be run on newer server versions. Perhaps we can keep the old tests and only run them on server versions up to 3.6, and make a note not to add any more functionality to them.


spec/integration/retryable_writes/shared/performs_legacy_retries.rb
```diff
@@ -0,0 +1,142 @@
...
+        sleep 1
+      end
+    end
+
+    context 'on server versions >= 4.4' do
+      min_server_fcv '4.3'
+      # These tests will be implemented in a follow-up PR
+    end
+
+    context 'on server versions <= 4.4' do
```
p-mongo, line 66: `<` ?

spec/integration/retryable_writes/shared/supports_modern_retries.rb
```diff
@@ -0,0 +1,31 @@
...
+          'by default. Standalone should NOT retry when modern retries are enabled.'
+      end
+
+      it_behaves_like 'it performs no retries'
+    end
+
+    context 'against a replica set or sharded cluster' do
+      require_topology :replica_set, :sharded
+
+      before do
```
p-mongo, line 22: All of the befores in this file should be `before(:all)` I think.

spec/integration/retryable_writes/shared/supports_modern_retries.rb
```diff
@@ -0,0 +1,31 @@
...
+
+module SupportsModernRetries
+  shared_examples 'it supports modern retries' do
+    let(:retry_writes) { true }
+
+    context 'against a standalone server' do
+      require_topology :single
+
+      before do
+        skip 'RUBY-2171: standalone topology currently uses legacy write retries ' \
```
p-mongo, line 12: And if `pending` works instead of `skip` here that would be a better option I think.

spec/integration/retryable_writes/shared/performs_legacy_retries.rb
egiurleo: Good catch, thanks.

spec/integration/retryable_writes/shared/supports_modern_retries.rb
egiurleo: It looks like pending can only be used inside `it` blocks, so I believe skip is the appropriate option here. I've left a comment on RUBY-2171 to make sure whoever works on that ticket goes and un-skips these tests.

PR #1906, opened 2020-05-15
RUBY-1995 Test collection and index creation in multi-doc txns

p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-1995

PR #1914, opened 2020-05-19
RUBY-2044: Part IV, Modify Retryable Write functionality to match 4.4 spec

egiurleo:
Starting in server version 4.4, the server will add the RetryableWriteError label to most errors that it deems retryable. If it does not add the label to an error (unless it's a connection error), the driver should not add the label itself. This PR also supports receiving the error label in writeConcernError.


lib/mongo/error.rb
```diff
@@ -120,6 +121,25 @@ def label?(label)
...
+      @write_concern_error_labels.include?(label)
+    end
+
+    # Whether the error has a particular label as a standard error label or an
+    # error label associated with the write concern error.
+    #
+    # @param [ String ] label The label to check for the presence of.
+    #
+    # @return [ Boolean ] Whether the error has the given label.
+    def label_or_write_concern_label?(label)
```
p-mongo, line 139: Can this method be removed and the `or` done inline in the  3 call sites please?

PR #1923, opened 2020-05-21
RUBY-2000 Support the allowDiskUseOption on find commands

egiurleo:
To add to a follow-up PR: support this option in GridFS.


lib/mongo/error/unsupported_allow_disk_use.rb
```diff
@@ -0,0 +1,38 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+module Mongo
+  class Error
+    # Raised if the allow_disk_use option is specified for an operation but the server
+    # selected does not support this option.
+    class UnsupportedAllowDiskUse < Error
```
p-mongo, line 18: I think I would prefer a single exception class for all of the unsupported options, so that our public API is more flexible with respect to future changes (i.e. in case there is a way of generically validating options in the future).

egiurleo: Okay! I'm going to merge this and refactor.

PR #1927, opened 2020-05-26
RUBY-2000 Make more general error class for unsupported options

egiurleo:


spec/mongo/error/unsupported_option_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'spec_helper'
+
+describe Mongo::Error::UnsupportedOption do
+  describe '#self.hint_error' do
```
p-mongo, line 4: We would spell this `describe '.hint_error'`

PR #1933, opened 2020-05-28
RUBY-2137 Include JRuby version in app metadata


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2137

PR #1934, opened 2020-05-29
RUBY-2184 Supported hedged reads

egiurleo:


spec/mongo/server_selector/secondary_preferred_spec.rb
```diff
@@ -12,6 +12,8 @@
   it_behaves_like 'a server selector with sensitive data in its options'
 
   it_behaves_like 'a server selector accepting tag sets'
+  it_behaves_like 'a server selector accepting hedge'
+  it_behaves_like 'a server selector accepting hedge'
```
p-mongo, line 16: Duplicated?

lib/mongo/server_selector/selectable.rb
```diff
@@ -49,6 +53,8 @@ def initialize(options = nil)
         @options = options.freeze
         @tag_sets = (options[:tag_sets] || []).freeze
         @max_staleness = options[:max_staleness]
+        @hedge = options[:hedge]
```
p-mongo, line 56: Can either all of these assignments be frozen or none? If all are frozen, the input arguments must be duplicated first since freeze mutates the object. E.g. `options.dup.freeze`, `(options[:tag_sets] || []).dup.freeze` .

spec/mongo/server_selector/secondary_preferred_spec.rb
egiurleo: Yup, thanks.

PR #1935, opened 2020-06-02
RUBY-2221 Add commitQuorum option to createIndexes

egiurleo:


lib/mongo/index/view.rb
```diff
@@ -149,21 +154,59 @@ def create_one(keys, options = {})
...
+      #
+      # @option options [ String | Integer ] :commit_quorum Specify how many
+      #   data-bearing members of a replica set, including the primary, must
+      #   complete the index builds successfully before the primary marks
+      #   the indexes as ready. Supported values are integers from 0 to the
+      #   number of data-bearing members of the replica set, "majority,"
+      #   replica set tag names, and "votingMembers." Default is "votingMembers."
+      #
+      # @return [ Result ] The result of the command.
+      def create_many_with_options(models, options={})
```
egiurleo, line 178: The spec calls for an options argument to be added to the `create_many` method. However, because our existing `create_many` method takes an array with a splat operator, it's impossible to just add another argument without breaking the API.

Adding another public method to the API doesn't seem like the right thing to do, so I'd appreciate some guidance here.

lib/mongo/index/view.rb
```diff
@@ -149,21 +154,59 @@ def create_one(keys, options = {})
       #
       # @since 2.0.0
       def create_many(*models)
```
p-mongo, line 156: Wouldn't the following work?

`def create_many(*models, **options)`

egiurleo: I believe that creates breaking changes in the API. Consider the following example:

`create_many({ key: { random: 1 } })`

When the method definition is `def create_many(*models)`, the `models` argument is `[{key: { random: 1 } }]`.

However, when the method definition is `def create_many(*models, **options)`, the `models` argument is an empty array, and `options` contains the specified Hash.

(Also happy to chat about this tomorrow.)

egiurleo: A simplified example for communication purposes:

```
def func1(*models)
  puts "MODELS: #{models}"
end

def func2(*models, **options)
   puts "MODELS: #{models}"
   puts "OPTIONS: #{options}"
 end

func1({ key: 'hello'})
# MODELS: [{:key=>"hello"}]

func2({ key: 'hello' })
# MODELS: []
# OPTIONS: {:key=>"hello"}
```

lib/mongo/index/view.rb
```diff
@@ -139,31 +144,65 @@ def create_one(keys, options = {})
...
-                  session: session
-                 }
+            indexes: indexes,
+            db_name: database.name,
+            coll_name: collection.name,
+            session: session,
+            commit_quorum: options[:commit_quorum]
+           }
+
           spec[:write_concern] = write_concern if server.with_connection { |connection| connection.features }.collation_enabled?
```
p-mongo, line 204: Please also use the description retrieved earlier here.

PR #1937, opened 2020-06-03
RUBY-2132 stop connecting connections in ensure_connected


p-mongo:
https://jira.mongodb.com/browse/RUBY-2132


p-mongo:
There is a lot of red presently in CI, I might need to take another pass later to fix some mock leaks that might arise from this PR.

PR #1947, opened 2020-06-07
Fix RUBY-2272 Exceptions raised during address resolution are not mapped to Mongo::Error


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2272

PR #1948, opened 2020-06-08
RUBY-1972 RUBY-2001 RUBY-2148 Hint documentation 

egiurleo:


docs/tutorials/ruby-driver-crud-operations.txt
```diff
@@ -458,20 +458,67 @@ the modification occurs.
   doc = artists.find_one_and_update( { :name => 'José James' }, { '$set' => { :name => 'José' } } )
   doc # Return the document before the update.
 
-``find_one_and_replace``
+Update Options
+~~~~~~~~~~~~~~
+
+To add options to an update command, specify them as key value pairs in the options
```
p-mongo, line 464: "key-value" pairs?

docs/tutorials/ruby-driver-crud-operations.txt
```diff
@@ -458,20 +458,67 @@ the modification occurs.
...
 
-  doc = artists.find_one_and_replace(
-    { :name => 'José James' },
-    { '$set' => { :name => 'José' } },
-    :return_document => :after
+  # Use the _id index to perform this operation
+  result = artists.update_one(
+    { :name => 'Goldie' },
+    { "$inc" => { :plays =>  1 } },
+    { hint: { _id: 1 } }
```
p-mongo, line 476: Is _id index used for the update rather than the find, i.e. this call searches for the document twice?

docs/tutorials/ruby-driver-crud-operations.txt
```diff
@@ -458,20 +458,67 @@ the modification occurs.
...
+     - How to sort the results of a findAndModify command. Specified as a Hash
+       key value pair, where the key is the name of the field to sort by, and
+       the value is either 1 or -1, specifying a sort in ascending or descending
+       order (only available on ``find_one_and_delete``, ``find_one_and_replace``,
+       and ``find_one_and_update`` commands).
+   * - ``session``
+     - The session to use for this operation.
+   * - ``upsert``
+     - Whether to upsert if the document doesn't exist. Cannot be used on
+       ``find_one_and_delete`` operation.
```
p-mongo, line 515: What about the `:return_document` that was in the example previously, does it still work?

egiurleo: Yup! Thanks for pointing that out.

docs/tutorials/ruby-driver-crud-operations.txt
egiurleo: This call doesn't search for documents twice! My understanding is that `find` just creates a `View` on the collection, and then the update command performs the search (with the specified index) and the update.

p-mongo: In this case I am confused: how would the _id index be used with a query on name?

egiurleo: I'll think of a clearer example.

PR #1953, opened 2020-06-12
RUBY-2281 Use local event subscribers in test suite


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2281

PR #1954, opened 2020-06-12
RUBY-2282 Unify test event subscriber classes


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2282

PR #1958, opened 2020-06-16
Fix RUBY-2279 AWS request mutates passed in time

Neilshweky:
Fixed issue where a call to formatted_time mutates the @time instance. Changed the call to utc to getutc.

### DIFF
--- a/spec/mongo/auth/aws/request_spec.rb
+++ b/spec/mongo/auth/aws/request_spec.rb
@@ -1,24 +1,23 @@
 require 'spec_helper'
-require 'json'
 
 describe Mongo::Auth::Aws::Request do 
 
-    describe "#formatted_time" do 
-        let(:request) do 
-            described_class.new(access_key_id: 'access_key_id', 
-                secret_access_key: 'secret_access_key', 
-                session_token: 'session_token', 
-                host: 'host', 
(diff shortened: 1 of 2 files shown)


spec/mongo/auth/aws/request_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+        session_token: 'session_token', 
+        host: 'host', 
+        server_nonce: 'server_nonce',
+        time: original_time
+      )
+    end
+
+    it 'doesn\'t modify the time instance variable' do
+      expect { request.formatted_time }.to_not raise_error
+    end
```
egiurleo, line 19: I would love to see an example that tests the behavior of the method -- what should formatted_time return? Does it actually do that? I would also love to see the description of the test match what is being tested.

spec/mongo/auth/aws/request_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'spec_helper'
+
+describe Mongo::Auth::Aws::Request do 
+
+  describe "#formatted_time" do 
+    original_time = Time.now.freeze
```
p-mongo, line 6: This should be a let block also like `request` is.

PR #1959, opened 2020-06-16
RUBY-2280 make empty hash optional to count_documents

gjchong25:


egiurleo:
Looking good so far! 💪 Please re-request my review once you add tests.

spec/integration/crud_spec.rb
```diff
@@ -127,13 +127,13 @@
   describe 'upsert' do
     context 'with default write concern' do
       it 'upserts' do
-        collection.count_documents({}).should == 0
+        collection.count_documents().should == 0
```
egiurleo, line 130: For all of these instances where you've removed the argument, you can just remove the parentheses entirely (i.e. `collection.count_documents.should == 0`)

spec/mongo/collection_spec.rb
```diff
@@ -2000,6 +2000,51 @@ def generate
...
+          end 
+
+          before do
+            authorized_collection.delete_many
+          end
+
+          context 'empty collection' do
+            it 'returns true' do
+              expect(view.count_documents()).to eq(view.count_documents({}))
+            end
```
egiurleo, line 2016: I definitely see the reasoning behind this test, but I think there is a potential pitfall -- what if the behavior of `count_documents` breaks? i.e. what if `view.count_documents()` and `view.count_documents({})` both return the same wrong thing? This test would still pass but the driver would be broken.

I would test this method against what you expect it to return, and modify the test description (currently `it "returns true"`) accordingly.

spec/mongo/collection_spec.rb
```diff
@@ -2000,6 +2000,51 @@ def generate
...
+          end
+
+          context 'nonempty collection' do
+            it 'returns false' do
```
egiurleo, line 2020: I would modify this to describe the expected behavior of the method.

spec/mongo/collection_spec.rb
```diff
@@ -2000,6 +2000,51 @@ def generate
...
+              documents = []
+              1.upto(10) do |index|
+                documents << { key: 'a', _id: "in#{index}" }
+              end
+
+              authorized_collection.insert_many(documents)
```
egiurleo, line 2026: This can be done using lets and a before block:

```
let(:documents) do
   documents = []
   1.upto(10) do |index|
      documents << { key: 'a', _id: "in#{index}" }
   end
end

before do
   authorized_collection.insert_many(documents)
end

it "description here" do
   ...
end
```

egiurleo: Which is not to say that this doesn't work, just that `let` and `before` are more idiomatic for respec and also help you separate out the logic of the test itself vs setup logic.

spec/mongo/collection_spec.rb
```diff
@@ -2000,6 +2000,51 @@ def generate
...
+              expect(authorized_collection.count_documents()).to eq(10)
```
egiurleo, line 2027: You can remove the parentheses on the method call :)

spec/mongo/collection_spec.rb
```diff
@@ -2000,6 +2000,51 @@ def generate
...
+              documents = []
+              documents << { key: 'a', _id: "1" }
+              documents << { key: 'b', _id: "2" }
+              documents << { key: 'c', _id: "3" }
+
+              authorized_collection.insert_many(documents)
+              puts authorized_collection.count_documents()
+              expect(authorized_collection.count_documents()).to eq(3)
+            end
+          end
```
egiurleo, line 2041: I don't think you need both this test and the previous test, unless there's some fundamental difference between the case where the collection has 10 elements in it and the one where it only has 3.

spec/mongo/collection_spec.rb
```diff
@@ -2000,6 +2000,47 @@ def generate
...
+            described_class.new(authorized_collection, {}, {})
+          end
+
+          before do
+            authorized_collection.delete_many
+          end
+
+          context 'empty collection' do
+            it 'returns true' do
+              expect(view.count_documents()).to eq(view.count_documents({}))
```
egiurleo, line 2015: I don't think you need `view` here, since you're just testing the `count_documents` method on the collection (you used it correctly in the other two tests). You can also remove the `view` helper (the `let(:view)` block).

spec/mongo/collection_spec.rb
```diff
@@ -2000,6 +2000,41 @@ def generate
...
+    context 'no argument provided' do
+
+    # populate documents to add to collection
+      let(:documents) do
+        documents = []
+        1.upto(10) do |index|
+          documents << { key: 'a', _id: "in#{index}" }
+        end
+        documents
+      end
```
egiurleo, line 2013: Nice! This is way cleaner. Now I would go back to the organization you had before:

```
context 'no argument provided' do
   context 'when collection is empty' do
      ...
   end

   context 'when collection is not empty' do
      ...
   end
end
```

That simplifies your setup (only need to keep track of one collection) and makes the test output easier to read.

spec/mongo/collection_spec.rb
```diff
@@ -2000,6 +2000,39 @@ def generate
...
   describe '#count_documents' do
+
+    # populate documents to add to collection
+    let(:documents) do
+      documents = []
+      1.upto(10) do |index|
+        documents << { key: 'a', _id: "in#{index}" }
+      end
+      documents
+    end
```
egiurleo, line 2011: I would move this into the `when collection is not empty` context since you only need it there.

spec/mongo/collection_spec.rb
```diff
@@ -2000,6 +2000,39 @@ def generate
...
+
+    let(:other_collection) do
+      authorized_client[:other]
+    end
+
+    before do
+      authorized_collection.insert_many(documents)
+      other_collection.delete_many
```
egiurleo, line 2019: `authorized_collection.delete_many`

PR #1964, opened 2020-06-18
RUBY-2165 Yield client to block from constructor

Neilshweky:
The fix itself was omly a 2-3 line fix... but the testing was the real kicker. The tests work, but I modified a lot more code than maybe necessary... Let me know if I need to fix that.

### DIFF
--- a/lib/mongo/client.rb
+++ b/lib/mongo/client.rb
@@ -489,11 +489,6 @@ module Mongo
           end
         end
 
-        if block_given? then 
-          yield(self)
-          self.close
-        end
-
       rescue
         begin
           @cluster.disconnect!
@@ -502,6 +497,14 @@ module Mongo
           # Drop this exception so that the original exception is raised
         end
         raise
+      end
+
+      if block_given? 
+        begin 
+          yield(self)
+        ensure
+          close
+        end
       end
     end
 
--- a/lib/mongo/client.rb
+++ b/lib/mongo/client.rb
@@ -489,7 +489,11 @@ module Mongo
           end
         end
 
-        yield(self) if block_given?
+        if block_given? then 
+          yield(self)
+          self.close
+        end
+
       rescue
         begin
           @cluster.disconnect!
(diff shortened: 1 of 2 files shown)


Neilshweky:
I tested that the client closed for a base case of the client... I didn't think the complicated client configurations affect this ticket...

spec/mongo/client_construction_spec.rb
```diff
@@ -1243,6 +1243,22 @@
...
+      let(:block_client) do
+          c = nil
+          Mongo::Client.new(
+            SpecConfig.instance.addresses,
+            SpecConfig.instance.test_options.merge(database: SpecConfig.instance.test_db),
+          ) do |client|
+            c = client
+          end
+          c
+        end
```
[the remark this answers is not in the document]
p-mongo: Seems like the content inside `let` is indented with 4 instead of 2 spaces.

lib/mongo/client.rb
```diff
@@ -489,7 +489,11 @@ def initialize(addresses_or_uri, options = nil)
           end
         end
 
-        yield(self) if block_given?
+        if block_given? then 
+          yield(self)
+          self.close
```
p-mongo, line 494: The `close` must be called also when the block raises an exception, thus an `ensure` statement is needed here + a test added which raises an exception (any exception will do) in the block after assigning the client.

Neilshweky: Oleg, I think this is actually unnecessary because the next few lines is actually a rescue block which disconnects the cluster. I can put the code there put it seems repeated... I'll still put a test there just to make sure this works.

spec/mongo/client_construction_spec.rb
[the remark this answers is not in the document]
Neilshweky: got it.

lib/mongo/client.rb
Neilshweky: On that note, should I change this self.close to @cluster.disconnect ?

PR #1967, opened 2020-06-20
Add options to disable built-in monitoring subscribers

leoarnold:
Previously the user could only select between enabling and disabling
all monitoring for an instance of `Mongo::Client`.

That meant that if a custom subscriber was added through the global API

```ruby
Mongo::Monitoring::Global.subscribe(Mongo::Monitoring::COMMAND, MyCommandSubscriber.new)
```

and client monitoring was turned off, then the built-in subscribers
were not subscribed to the client's montoring, but neither was
the custom subscriber.

If client monitoring was turned on, then the custom subscriber was
subscribed to the client's monitoring, but so were most of the
buit-in subscribers with no option to turn them off.

Users who only want to see the debug output of their custom subscriber
may be annoyed by this and feel that it "clutters" their logs.

This commit adds the possibility to disable the built-in subscribers
one by one without breaking the gem's public API.


egiurleo:
Thanks for the patch @leoarnold, and sorry for my review yesterday, I meant to review a different PR and opened this one instead.

Just to make sure I understand the motivation behind this PR -- is the issue just that default subscribers clutter the logs? Would creating a different logger for your custom subscriber solve the same problem?

leoarnold:
@egiurleo My personal use case is this: In a Rails project with Mongoid I added an initializer

```ruby
Mongo::Monitoring::Global.subscribe(Mongo::Monitoring::COMMAND, MyCommandSubscriber.new)
```

where `MyCommandSubscriber` only logs to STDERR (it may also log to the Rails log in later iterations, that's tbd).

To my surprise, I noticed that `MyCommandSubscriber` did not log at all and wasn't even subscribed to the default Mongo client. This was because `development.clients.default.options.monitoring` was set to `false` in `mongoid.yml`.

I then activated that option and `MyCommandSubscriber` started logging - but so did all builtin subscribers as well. Suddenly a lot of `MongoDB` lines made it hard to read along `rails s` whilst running requests against the rails server.

I agree that setting `Mongo::Logger.logger` to a different logger which maybe logs to `log/mongo.log` has perks in its own right. But then again we still have to direct all or none of the builtin subscriber to that logger - which kind of takes away from the benefit of having the different categories of events already in place.

_tl;dr_ Kinda question of taste, i.e. maintainers' decision.

p-mongo:
@leoarnold What is your log level (https://docs.mongodb.com/mongoid/master/tutorials/mongoid-configuration/#logging) ?

leoarnold:
@p-mongo `Rails.logger == Mongo::Logger.logger == Mongoid.logger` and log level is `:debug` in the development environment (as it should be, I think). `MyCommandSubscriber` also uses debug, as its primary purpose is debugging (hence the logging to STDERR)

p-mongo:
Thanks. Can you clarify whether you would enable some of the driver logging facilities as proposed in this PR, as opposed to always disabling everything? 

If you configure a custom logger for Mongoid and/or driver as described in the docs, would this also fulfill your requirements?

If Mongoid provided a way to specify logger level for Mongoid and/or driver that is different from Rails' logger level, would this fulfill your requirements?

p-mongo:
@leoarnold ping ^

agolin95:
Tracked in https://github.com/mongodb/mongo-ruby-driver/pull/1967

PR #1968, opened 2020-06-22
RUBY-1262 added tests for closing session after block

Neilshweky:
It seems like the with_session function on the client already closes the session on completion of the block. All I did was add some tests to make sure it works...

Some notes:
- I did this in `session_spec.rb` but perhaps it should have been in `client_spec.rb`
- When there is an error inside the `with_session` block, it is rescued and never re-raised.

### DIFF
--- a/docs/tutorials/ruby-driver-sessions.txt
+++ b/docs/tutorials/ruby-driver-sessions.txt
@@ -92,7 +92,7 @@ A session can be created by calling the ``start_session`` method on a client:
 
   session = client.start_session
 
-When ``start_session`` is used without passing a block to it, the driver does not automatically clean up the session which can result in an accumulation of sessions on the server. Use `end_session<#end-a-session>` to manually end the session created. The server will automatically clean up old sessions after a timeout but the application should end sessions when the sessions are no longer needed.
+When ``start_session`` is used without passing a block to it, the driver does not automatically clean up the session which can result in an accumulation of sessions on the server. Use `end_session <#end-a-session>`_ to manually end the session created. The server will automatically clean up old sessions after a timeout but the application should end sessions when the sessions are no longer needed.
 
 Unacknowledged Writes
 =====================
@@ -141,5 +141,5 @@ The Ruby driver will then add the id for the corresponding server session to a p
 When a client is closed, the driver will send a command to the server to end all sessions it has cached
 in its server session pool. You may see this command in your logs when a client is closed.
 
-Note that when using the `block syntax<#creating-a-session-from-a-mongo-client>` for ``start_session`` the session is automatically ended after
+Note that when using the `block syntax <#creating-a-session-from-a-mongo-client>`_ for ``start_session`` the session is automatically ended after
 the block finishes executing.
--- a/docs/tutorials/ruby-driver-sessions.txt
+++ b/docs/tutorials/ruby-driver-sessions.txt
@@ -90,12 +90,6 @@ A session can be created by calling the ``start_session`` method on a client:
 
   session = client.start_session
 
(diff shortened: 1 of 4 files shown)


p-mongo:
There was a fail in `egos` tool causing all tests to fail, but the tests added in this PR do not appear to work.

It also appears that `with_session` is private, hence I would suggest changing `start_session` to take the block as it is already  a public method.

Lastly the documentation in https://docs.mongodb.com/ruby-driver/master/tutorials/ruby-driver-sessions/ needs to be updated.

I revised https://jira.mongodb.org/browse/RUBY-1262 with the current requirements.

p-mongo:
Everything in this PR is good. Can you also add a test that the (return) value of the block is returned from the `start_session` method when the block form is used?

Neilshweky:
I just updated the documentation to use the block syntax for `start_session`... The only section I didn't do this for was Causal Consistency, but this was after I introduced the alternative form of starting a session (i.e. `session = client.start_session`). It wouldn't be difficult to change I just think it looks pretty readable as it... Let me know if you think I should... 

I also added a note to the end of the End Session section specifying how the block syntax automatically closes the session.

docs/tutorials/ruby-driver-sessions.txt
```diff
@@ -23,12 +23,13 @@ Please note that session objects are not thread safe. They must only be used by
 Creating a session from a ``Mongo::Client``
 ===========================================
 
-A session can be created by calling the ``start_session`` method on a client:
+A session can be created by calling the ``start_session`` method on a client and passing it a block:
```
p-mongo, line 26: Can you please add a sentence like "When using the block form, the session will be automatically ended by the driver after the block executes" here.

docs/tutorials/ruby-driver-sessions.txt
```diff
@@ -23,12 +23,13 @@ Please note that session objects are not thread safe. They must only be used by
...
 
 .. code-block:: ruby
 
-  session = client.start_session
-
+  client.start_session do |session| 
+    # work with session
```
p-mongo, line 31: "work with the session"

docs/tutorials/ruby-driver-sessions.txt
```diff
@@ -52,29 +54,41 @@ Create a session and execute an insert, then a find using that session:
...
+  end
 
+Alternative way to create a session
+===================================
+
+A session can be created by calling the ``start_session`` method on a client:
+
+.. code-block:: ruby
+
+  session = client.start_session
```
p-mongo, line 91: Can you please add the following here after the code example:

> When `start_session` is used without passing a block to it, the driver does not automatically clean up the session which can result in an accumulation of sessions on the server. Use `end_session` (link to the section covering session ending) to manually end the session created. The server will automatically clean up old sessions after a timeout but the application should end sessions when the sessions are no longer needed.

docs/tutorials/ruby-driver-sessions.txt
```diff
@@ -122,3 +136,6 @@ To end a session, call the ``end_session`` method:
 The Ruby driver will then add the id for the corresponding server session to a pool for reuse.
 When a client is closed, the driver will send a command to the server to end all sessions it has cached
 in its server session pool. You may see this command in your logs when a client is closed.
+
+Note that when using the block syntax for ``start_session`` the session is automatically ended after
```
p-mongo, line 140: This sentence is good and should stay here. Can the phrase "block syntax" be hyperlinked to the section that covers starting sessions, and there should be a comma after `start_session`.

Neilshweky:
Okay I just added the two hyperlinks and finished all of the other things you asked me to add.

PR #1971, opened 2020-06-22
RUBY-1657 added error message to timeouts

Neilshweky:
I couldn't figure out how to test for this reliably. I tried to set the timeout duration from the test but ultimately it didn't work...

### DIFF
--- a/lib/mongo/socket/ssl.rb
+++ b/lib/mongo/socket/ssl.rb
@@ -137,7 +137,7 @@ module Mongo
       #
       # @since 2.0.0
       def connect!
-        Timeout.timeout(options[:connect_timeout], Error::SocketTimeoutError, 'The socket took over #{options[:connect_timeout]} seconds to connect') do
+        Timeout.timeout(options[:connect_timeout], Error::SocketTimeoutError, "The socket took over #{options[:connect_timeout]} seconds to connect") do
           map_exceptions do
             @tcp_socket.connect(::Socket.pack_sockaddr_in(port, host))
           end
--- a/lib/mongo/socket/ssl.rb
+++ b/lib/mongo/socket/ssl.rb
@@ -137,7 +137,7 @@ module Mongo
       #
       # @since 2.0.0
       def connect!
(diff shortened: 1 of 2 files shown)


lib/mongo/socket/ssl.rb
```diff
@@ -137,7 +137,7 @@ def initialize(host, port, host_name, timeout, family, options = {})
       #
       # @since 2.0.0
       def connect!
-        Timeout.timeout(options[:connect_timeout], Error::SocketTimeoutError) do
+        Timeout.timeout(options[:connect_timeout], Error::SocketTimeoutError, 'the socket took too long to connect') do
```
p-mongo, line 140: Can you please add the timeout into the message, e.g. `The socket took over #{options[:connect_timeout]} seconds to connect` (and capitalize the first letter).

p-mongo:
I don't think we need test coverage for this change.

Neilshweky:
Great! Is there anything else you need me to add then?

p-mongo:
@Neilshweky Can you please add the timeout values to the messages as I mentioned in the comment?

Neilshweky:
@p-mongo Sorry I didn't see your comment, I just changed that and pushed it

lib/mongo/socket/ssl.rb
```diff
@@ -137,7 +137,7 @@ def initialize(host, port, host_name, timeout, family, options = {})
       #
       # @since 2.0.0
       def connect!
-        Timeout.timeout(options[:connect_timeout], Error::SocketTimeoutError) do
+        Timeout.timeout(options[:connect_timeout], Error::SocketTimeoutError, 'The socket took over #{options[:connect_timeout]} seconds to connect') do
```
p-mongo, line 140: The string needs double quotes to perform interpolation.

You can test the message that is generated by adding e.g. a `sleep 10` call in the block here, and instantiating a client with a lower connect timeout: `Mongo::Client.new(['localhost'], connect_timeout: 1)`. This should raise the timeout error with the message you are adding here.

Neilshweky: fixed and tested.

PR #1972, opened 2020-06-23
RUBY-2031 fails when protocol is included in host array; added tests

Neilshweky:
Accomplished using a Regex to check for protocol. 

I wasn't sure what kind of error to raise here so I just raised a StandardError and included an error message that describes and tries to help diagnose the problem.

### DIFF
--- a/lib/mongo/client.rb
+++ b/lib/mongo/client.rb
@@ -443,7 +443,12 @@ module Mongo
         @srv_records = uri.srv_records
       else
         addresses = addresses_or_uri
-        @srv_records = nil
+        if addresses.any? { |addr| addr =~ /mongodb(\+srv)?:\/\// }
+          raise "Protocol should not be included in host list. Did you mean to not use an array?"
+        else
+          @srv_records = nil
+        end
+   
       end
 
       # Special handling for sdam_proc as it is only used during client
--- a/spec/mongo/client_construction_spec.rb
+++ b/spec/mongo/client_construction_spec.rb
@@ -1277,6 +1277,26 @@ describe Mongo::Client do
           expect(block_client_raise.cluster.connected?).to eq(false)
         end
       end
+
+      context 'when the hosts given include the protocol' do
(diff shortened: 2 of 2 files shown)


lib/mongo/client.rb
```diff
@@ -443,7 +443,12 @@ def initialize(addresses_or_uri, options = nil)
         @srv_records = uri.srv_records
       else
         addresses = addresses_or_uri
-        @srv_records = nil
+        if addresses.any? { |addr| addr =~ /mongodb(\+srv)?:\/\// }
+          raise "Protocol should not be included in host list. Did you mean to not use an array?"
```
p-mongo, line 447: I think this line can raise `ArgumentError` going by the validation  methods toward the bottom of the file.

The offending address should be included in the exception message, which means probably the raising should be done inside the loop that iterates the addresses.

lib/mongo/client.rb
```diff
@@ -443,7 +443,12 @@ def initialize(addresses_or_uri, options = nil)
...
+        else
+          @srv_records = nil
```
p-mongo, line 449: Can you put this line back on the original indentation level please. We generally do not nest code when nesting is not needed.

spec/mongo/client_construction_spec.rb
```diff
@@ -1277,6 +1277,26 @@
...
+            Mongo::Client.new(['mongodb+srv://127.0.0.1:27017/test'])
+          end.to raise_error(StandardError, "Protocol should not be included in host list. Did you mean to not use an array?")
+        end
+
+        it 'raises an error on multiple items' do
+          expect do 
+            Mongo::Client.new(['127.0.0.1:27017', 'mongodb+srv://127.0.0.1:27017/test'])
+          end.to raise_error(StandardError, "Protocol should not be included in host list. Did you mean to not use an array?")
+        end
+      end
```
p-mongo, line 1299: :+1:

PR #1973, opened 2020-06-23
RUBY-2143 Pass through all options to listCollections command

gjchong25:


lib/mongo/database/view.rb
```diff
@@ -55,7 +62,8 @@ def collection_names(options = {})
         @batch_size = options[:batch_size]
         session = client.send(:get_session, options)
         cursor = read_with_retry_cursor(session, ServerSelector.primary, self) do |server|
-          send_initial_query(server, session, name_only: true)
+          options[:name_only] = true
```
p-mongo, line 65: This mutates options. We generally do not mutate arguments to methods. A possible way of spelling this is `options = options.merge(name_only: true)`.

lib/mongo/database/view.rb
```diff
@@ -55,7 +62,8 @@ def collection_names(options = {})
...
+          send_initial_query(server, session, options)
```
p-mongo, line 66: Alternatively on this line you can use `send_initial_query(server, session, options.merge(name_only: true))`.

lib/mongo/database/view.rb
```diff
@@ -47,6 +47,13 @@ class View
       #
       # @option options [ Integer ] :batch_size  The batch size for results
       #   returned from the listCollections command.
+      # @option options [ Hash ] :filter A filter on the collections returned.
+      # @option options [ true, false ] :authorized_collections A flag, when
+      #   set to true and used with nameOnly: true, that allows a user without the
```
p-mongo, line 52: Since this method always adds `name_only`, the note about `nameOnly` does not apply to this method.

lib/mongo/database.rb
```diff
@@ -129,27 +139,44 @@ def collection_names(options = {})
...
     # Get all the non-system collections that belong to this database.
     #
     # @note The set of returned collections depends on the version of
     #   MongoDB server that fulfills the request.
     #
+    # @param [ Hash ] options
+    #
+    # @option options [ Hash ] :filter A filter on the collections returned. 
+    # @option options [ true, false ] :authorized_collections A flag, when
+    #   set to true and used with nameOnly: true, that allows a user without the
```
p-mongo, line 169: `name_only` please, since users will be specifying the option using Ruby syntax.

lib/mongo/database/view.rb
```diff
@@ -164,8 +180,12 @@ def collections_info_spec(session, options = {})
           db_name: @database.name,
           session: session
         }.tap do |spec|
-          spec[:selector][:nameOnly] = true if options[:name_only]
+          # nameOnly and authorizedCollections default to false
+          if options[:name_only] ? (spec[:selector][:nameOnly] = true) : (spec[:selector][:nameOnly] = false)
```
p-mongo, line 184: We do not need to specify false values. Was there some case that the existing code didn't handle? It seems sufficient to me.

lib/mongo/database/view.rb
```diff
@@ -164,8 +180,12 @@ def collections_info_spec(session, options = {})
...
+          end
           spec[:selector][:filter] = options[:filter] if options[:filter]
+          if options[:authorized_collections] ? (spec[:selector][:authorizedCollections] = true) : (spec[:selector][:authorizedCollections] = false)
```
p-mongo, line 187: Similarly here you should be able to just do `          spec[:selector][:authorizedCollections] = true if options[:authorized_collections]`.

spec/mongo/database_spec.rb
```diff
@@ -153,7 +153,80 @@
         expect(collection_names).to include('0_dalmatians')
         expect(collection_names).to include('1_dalmatians')
       end
+    end
+
+    context 'when provided a filter' do
+
+      before do
+        database[:users2].drop
```
p-mongo, line 161: We should generally be using strings for database and collection names, so I think this should be spelled `        database['users2'].drop`.

spec/mongo/database_spec.rb
```diff
@@ -153,7 +153,80 @@
...
+        expect(result.length).to eq(1)
+        expect(result.first).to eq('users2')
+      end
+    end
+
+    context 'when provided authorized_collections or not' do
+
+      require_topology :replica_set, :sharded
+      context 'on server versions >= 4.4' do
+        min_server_fcv '4.2'
```
p-mongo, line 179: The requirement here (4.2) does not match the context claim (4.4).

spec/mongo/database_spec.rb
```diff
@@ -153,7 +153,80 @@
...
 
+      it 'returns users2 collection' do
+        expect(result.length).to eq(1)
+        expect(result.first).to eq('users2')
+      end
+    end
+
+    context 'when provided authorized_collections or not' do
+
+      require_topology :replica_set, :sharded
```
p-mongo, line 177: Does listing collections behave differently in single topology?

lib/mongo/database/view.rb
gjchong25: If a user were to pass in name_only as false into list_collections (which does not force name_only to be true), then nameOnly would be set to nil because spec[:selector][:nameOnly] = true only when options[name_only] is true.

lib/mongo/database/view.rb
gjchong25: What is the difference functionality-wise between my method and yours?

PR #1975, opened 2020-06-24
RUBY-1413 modified Change Stream Example 4

Neilshweky:
I couldn't get the code to run on my local machine but I used the same syntax as what was there before and just modified the variable names/pipeline object, so it should be fine.

### DIFF
--- a/spec/integration/change_stream_examples_spec.rb
+++ b/spec/integration/change_stream_examples_spec.rb
@@ -174,10 +174,13 @@ describe 'change streams examples in Ruby' do
 
         # Start Changestream Example 4
 
-        pipeline = [ {'$match' => { '$or' => [{ 'fullDocument.username' => 'alice' },
-                                              { 'operationType' => 'delete' }] } }]
-        cursor = inventory.watch(pipeline).to_enum
-        cursor.next
+        pipeline = [
+          { $match => { 'fullDocument.username' => 'alice' } },
+          { $addFields => { 'newField' => 'this is an added field!' } }
+        ];
+        inventory = db[:inventory]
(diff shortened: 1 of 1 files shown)


spec/integration/change_stream_examples_spec.rb
```diff
@@ -174,10 +174,12 @@
...
 
-        pipeline = [ {'$match' => { '$or' => [{ 'fullDocument.username' => 'alice' },
-                                              { 'operationType' => 'delete' }] } }]
-        cursor = inventory.watch(pipeline).to_enum
-        cursor.next
+        pipeline = [
+          { "$match" => { 'fullDocument.username' => 'alice' } },
+          { "$addFields" => { 'newField' => 'this is an added field!' } }
+        ];
+        changeStream = inventory.watch(pipeline).to_enum
```
p-mongo, line 181: Let's use `cursor` for consistency with other examples.

spec/integration/change_stream_examples_spec.rb
```diff
@@ -174,10 +174,12 @@
...
+        changeStream.next
 
         # End Changestream Example 4
```
p-mongo, line 184: After this line please add an assertion that the document retrieved from the change stream has the expected fields (i.e. the `newField` is present). You can look at other examples to see how they assert after the example itself.

Neilshweky: Done!

spec/integration/change_stream_examples_spec.rb
Neilshweky: Done!

PR #1978, opened 2020-06-25
Switch to C version of egos for even more accurate timestamps


p-mongo:


p-mongo:
I think the 16K buffer/read size I am using in egos currently is causing massive batching, I'll play with the buffer size but it won't need code changes in the driver.

PR #1979, opened 2020-06-25
RUBY-2209 Support for 'authorizedDatabases' option

gjchong25:


lib/mongo/client.rb
```diff
@@ -815,6 +815,13 @@ def reconnect
     # @param [ Hash ] filter The filter criteria for getting a list of databases.
     # @param [ Hash ] opts The command options.
     #
+    # @option opts [ true, false ] :authorized_databases A flag that determines
+    #   which databases are returned based on user privileges when access control
+    #   enabled
```
egiurleo, line 820: Small typo -- maybe access control *is* enabled?

lib/mongo/client.rb
```diff
@@ -831,13 +838,21 @@ def database_names(filter = {}, opts = {})
     # @param [ true, false ] name_only Whether to only return each database name without full metadata.
     # @param [ Hash ] opts The command options.
     #
+    # @option opts [ true, false ] :authorized_databases A flag that determines
+    #   which databases are returned based on user privileges when access control
+    #   enabled
```
egiurleo, line 843: See above comment re: typo.

spec/mongo/client_spec.rb
```diff
@@ -595,6 +595,7 @@
     end
 
     context 'when name_only is true' do
+      min_server_fcv '3.6'
```
egiurleo, line 598: Good find!

PR #1981, opened 2020-06-25
Unbuffer standard i/o with stdbuf and on the Ruby side

p-mongo:


p-mongo:
With this PR the timestamps reported by driver's internal logger seem to 90% match the timestamps reported by egos.

PR #1983, opened 2020-06-29
Ruby 1865 - Support per-CRUD operation write options

gjchong25:


lib/mongo/collection.rb
```diff
@@ -537,8 +537,15 @@ def inspect
     # @since 2.0.0
     def insert_one(document, opts = {})
       client.send(:with_session, opts) do |session|
-        write_concern = write_concern_with_session(session)
+
+        if opts.has_key?(:write_concern) ? (write_concern = WriteConcern.get(opts[:write_concern])) : (write_concern = write_concern_with_session(session))
```
p-mongo, line 541: We prefer `key?` over `has_key?` and spelling out the if-then-else over writing ternaries.

spec/mongo/collection_spec.rb
```diff
@@ -1683,6 +1684,43 @@ def generate
...
+          client.subscribe(Mongo::Monitoring::COMMAND, subscriber)
+        end
+      end
+
+      let(:events) do
+        subscriber.command_started_events('insert')
+      end
+
+      let(:collection) do
+        authorized_collection.with(write_concern: {w:3})
```
p-mongo, line 1708: Space after the second colon please.

spec/mongo/collection_spec.rb
```diff
@@ -1683,6 +1684,43 @@ def generate
...
+      end
+
+      let!(:operation) do
+        collection.insert_one({name: 'grace'}, {session:session, write_concern: {w:1}})
```
p-mongo, line 1712: The hash braces are unnecessary on the last method argument.

lib/mongo/collection.rb
```diff
@@ -537,19 +537,26 @@ def inspect
...
-              :id_generator => client.options[:id_generator],
-              :session => session,
-              :txn_num => txn_num
-           ).execute(server, client: client)
+
+      write_concern = if opts[:write_concern]
+        WriteConcern.get(opts[:write_concern])
+      else
+        write_concern_with_session(session)
+      end
```
p-mongo, line 545: https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#behavior-of-the-writeconcern-field says the server will error when write concern is given in a transaction on an operation, therefore  I think the proposed code here is suitable.

lib/mongo/collection.rb
```diff
@@ -537,19 +537,26 @@ def inspect
...
+
+      write_with_retry(session, write_concern) do |server, txn_num|
+      max_time_ms = opts[:max_time_ms] if opts[:max_time_ms]
```
p-mongo, line 548: This and subsequent lines in the block appear to be not indented.

p-mongo: The `max_time_ms` assigned here is not used as far as I can tell?

lib/mongo/collection/view/writable.rb
```diff
@@ -182,6 +191,7 @@ def delete_many(opts = {})
                   :db_name => collection.database.name,
                   :coll_name => collection.name,
                   :write_concern => write_concern,
+                  :bypass_document_validation => !!opts[:bypass_document_validation],
```
p-mongo, line 194: I think this is fine for now as long but we may need to change this later to not send the false value if this fails any tests.

On which note, I don't see tests being run for this PR, so perhaps there are actually some failures due to the false value being passed.

lib/mongo/cursor/builder/get_more_command.rb
```diff
@@ -65,7 +65,7 @@ def get_more_command
           # If the max_await_time_ms option is set, then we set maxTimeMS on
           # the get more command.
           if view.respond_to?(:max_await_time_ms)
-            if view.max_await_time_ms && view.options[:await_data]
+            if view.max_await_time_ms && view.options[:await_data]
```
p-mongo, line 68: Remove the trailing space please.

spec/mongo/collection_spec.rb
```diff
@@ -1601,6 +1601,35 @@ def generate
...
+      let(:events) do
+        subscriber.command_started_events('insert')
+      end
+
+      let(:collection) do
+        authorized_collection.with(write_concern: {w: 2})
+      end
+
+      let!(:command) do
+       Utils.get_command_event(authorized_client, 'insert') do |client|
```
p-mongo, line 1620: This line appears to be misindented.

lib/mongo/collection/view/writable.rb
egiurleo: The CRUD spec tests are pretty sparse in terms of which options they test, so I wouldn't be shocked if they don't really test bypassDocumentValidation.

egiurleo: So agreed that this might need to be cleaned up one day, but it doesn't seem to be failing any tests right now.

p-mongo: https://jira.mongodb.org/browse/DRIVERS-1316 filed to address this.

PR #1984, opened 2020-06-29
RUBY-2295 include collation option in count_documents

Neilshweky:
Worked on this with Emily. Checked for collation option in opts and if it doesn't exist get it from @options.

### DIFF
--- a/lib/mongo/collection/view/readable.rb
+++ b/lib/mongo/collection/view/readable.rb
@@ -193,7 +193,9 @@ module Mongo
           pipeline << { :'$limit' => opts[:limit] } if opts[:limit]
           pipeline << { :'$group' => { _id: 1, n: { :'$sum' => 1 } } }
 
-          opts.select! { |k, _| [:hint, :max_time_ms, :read, :collation, :session].include?(k) }
+          opts = opts.select { |k, _| [:hint, :max_time_ms, :read, :collation, :session].include?(k) }
+          opts[:collation] = @options[:collation] unless opts[:collation]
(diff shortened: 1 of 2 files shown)


lib/mongo/collection/view/readable.rb
```diff
@@ -193,7 +193,9 @@ def count_documents(opts = {})
           pipeline << { :'$limit' => opts[:limit] } if opts[:limit]
           pipeline << { :'$group' => { _id: 1, n: { :'$sum' => 1 } } }
 
-          opts.select! { |k, _| [:hint, :max_time_ms, :read, :collation, :session].include?(k) }
+          opts = opts.select { |k, _| [:hint, :max_time_ms, :read, :collation, :session].include?(k) }
+          opts[:collation] = @options[:collation] unless opts[:collation]
```
egiurleo, line 197: I actually think this could be written in an even nicer way:
```
opts[:collation] ||= collation
```
With the `collation` method that's already defined on `view.rb`. Happy to chat more about `||=` syntax if you haven't seen it before.

spec/mongo/collection_spec.rb
```diff
@@ -941,6 +941,28 @@
...
       end
     end
+
+    context 'when collation has a strength' do 
+      let(:band_collection) do 
+        described_class.new(database, :bands)
+      end
+
+      before do
+        band_collection.delete_many({})
```
egiurleo, line 951: You can omit the parentheses and the empty hash :)

PR #1988, opened 2020-07-07
fix the test suite


p-mongo:


p-mongo:
This does not fix everything but it fixes most things and the suite is much greener now.

PR #1992, opened 2020-07-08
RUBY-2313 Manually batch tests to avoid intermittent failures in CI


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2313

p-mongo:
This still does not fix the failures but it proves that even a single test file in a fresh process can fail in evergreen, hopefully assisting with future diagnostics.

PR #2006, opened 2020-07-12
RUBY-2138 background thread leak checker


p-mongo:
https://jira.mongodb.com/browse/RUBY-2138

### DIFF
--- a/spec/integration/check_clean_slate_spec.rb
+++ b/spec/integration/check_clean_slate_spec.rb
@@ -2,7 +2,10 @@ require 'lite_spec_helper'
 
 # This test can be used to manually verify that there are no leaked
 # background threads - execute it after executing another test (in the same
-# rspec run) that is suspected to leak background threads.
+# rspec run) that is suspected to leak background threads, such as by
+# running:
+#
+# rspec your_spec.rb spec/integration/check_clean_slate_spec.rb
 
 describe 'Check clean slate' do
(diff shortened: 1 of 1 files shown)


spec/integration/check_clean_slate_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'lite_spec_helper'
+
+# This test can be used to manually verify that there are no leaked
+# background threads - execute it after executing another test (in the same
+# rspec run) that is suspected to leak background threads.
```
egiurleo, line 5: Could you provide some guidance for how to run this test in the same rspec run as another test?

p-mongo: It's literally just `rspec foo_spec.rb spec/integration/check_clean_slate_spec.rb`.  How would you like me to phrase that?

egiurleo: Oh that's great. I didn't know you could do that. Could you just include that in the comment?
```
execute it after executing another test in the same rspec run with the following command:

rspec your_spec.rb spec/integration/check_clean_slate_spec.rb
```

Does that sound good?

p-mongo: Yup, added to the comment thanks.

egiurleo: Thanks!

PR #2014, opened 2020-07-13
RUBY-2286 Use infinite timeout in push monitor when connectTimeoutMS=0


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2286

PR #2020, opened 2020-07-15
RUBY-2018 Add client metadata support for wrapping libraries


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2018

PR #2026, opened 2020-07-20
RUBY-2234 Fix error on large bulk writes with zlib

egiurleo:


lib/mongo/server/connection_base.rb
```diff
@@ -213,12 +212,21 @@ def serialize(message, client, buffer = BSON::ByteBuffer.new)
           max_bson_size += MAX_BSON_COMMAND_OVERHEAD
         end
 
-        final_message.serialize(buffer, max_bson_size)
+        temp_buffer = BSON::ByteBuffer.new
+        temp_buffer.put_string(buffer.to_s)
```
p-mongo, line 216: I believe this should be `put_bytes` and could you add a comment that this copies the existing passed in buffer contents, if any.

egiurleo:
I broke jruby, so please disregard for now :)

lib/mongo/server/connection_base.rb
```diff
@@ -213,12 +210,39 @@ def serialize(message, client, buffer = BSON::ByteBuffer.new)
...
+        #
+        # This is to avoid the case where the user performs a bulk write
+        # larger than 16MiB which, when compressed, becomes smaller than 16MiB.
+        # If the driver does not split the bulk writes prior to compression,
+        # the entire operation will be sent to the server, which will raise an
+        # error because the uncompressed operation exceeds the maximum bson size.
+        #
+        # To address this problem, we serialize the message prior to compression
+        # and raise an exception if the serialized message exceeds the maximum
+        # bson size.
```
p-mongo, line 225: :+1:  Thank you, very well written.

lib/mongo/server/connection_base.rb
```diff
@@ -213,12 +210,39 @@ def serialize(message, client, buffer = BSON::ByteBuffer.new)
...
+
+        # Create a separate buffer that contains the un-compressed message
+        # for the purpose of checking its size. Write any pre-existing contents
+        # from the original buffer into the temporary one.
+        temp_buffer = BSON::ByteBuffer.new
+        temp_buffer.put_bytes(buffer.get_bytes(buffer.length))
```
p-mongo, line 231: This actually mutates the buffer on the reading side, which we could look into addressing in the future.

lib/mongo/server/connection_base.rb
```diff
@@ -213,12 +210,39 @@ def serialize(message, client, buffer = BSON::ByteBuffer.new)
...
+
+        message.serialize(temp_buffer, max_bson_size)
+
+        if max_message_size && temp_buffer.length > max_message_size
```
p-mongo, line 235: Could you move the serialize call under this condition? It seems this should work correctly.

lib/mongo/server/connection_base.rb
p-mongo: Could you add a TODO comment noting that the buffer is mutated on this line?

lib/mongo/server/connection_base.rb
egiurleo: Sorry what serialize call should I move under this condition? I have to serialize the message into the `temp_buffer` before performing this check.

p-mongo: if max_message size; serialize; if temp_buffer.length > max_message_size ...

egiurleo: 👍 thanks for the clarification

lib/mongo/server/connection_base.rb
```diff
@@ -213,12 +210,41 @@ def serialize(message, client, buffer = BSON::ByteBuffer.new)
...
+          temp_buffer = BSON::ByteBuffer.new
+
+          # TODO: address the fact that this line mutates the buffer.
+          temp_buffer.put_bytes(buffer.get_bytes(buffer.length))
+
+          message.serialize(temp_buffer, max_bson_size)
+          if temp_buffer.length > max_message_size
+            raise Error::MaxMessageSize.new(max_message_size)
+          end
         end
```
egiurleo, line 239: @p-mongo is this what you had in mind?

p-mongo: :+1:  Yep looks good.

PR #2027, opened 2020-07-20
RUBY-1849 Implement Caching Cursor and Query Cache in driver

gjchong25:


lib/mongo/cached_cursor.rb
```diff
@@ -0,0 +1,60 @@
+# A Cursor that attempts to load documents from memory first before hitting
+# the database if the same query has already been executed.
+#
+# @since 5.0.0
```
egiurleo, line 4: Can you move this between lines 6 and 7? YARD won't understand that this comment is related to the `CachedCursor` class unless it's directly above the class definition.

lib/mongo/cached_cursor.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    # We iterate over the cached documents if they exist already in the
+    # cursor otherwise proceed as normal.
+    #
+    # @example Iterate over the documents.
+    #   cursor.each do |doc|
+    #     # ...
+    #   end
+    #
+    # @since 5.0.0
```
p-mongo, line 17: All of the since tags should be removed since they reference Mongoid versions (and we aren't using them anymore in new code).

lib/mongo/cached_cursor.rb
```diff
@@ -0,0 +1,60 @@
...
+    # Get a human-readable string representation of +Cursor+.
+    #
+    # @example Inspect the cursor.
+    #   cursor.inspect
+    #
+    # @return [ String ] A string representation of a +Cursor+ instance.
+    #
+    # @since 2.0.0
+    def inspect
+      "#<Mongo::QueryCache::CachedCursor:0x#{object_id} @view=#{@view.inspect}>"
```
p-mongo, line 42: `Mongo::CachedCursor`

lib/mongo/collection/view/iterable.rb
```diff
@@ -77,6 +102,25 @@ def close_query
...
+          end
+          cursor || QueryCache.cache_table[cache_key]
+        end
+
+        def cache_key
+          [ collection.namespace, selector, limit, skip, sort, projection, collation ]
+        end
+
+        def system_collection?
+          collection.namespace =~ /\Asystem./
```
p-mongo, line 121: The dot  needs to be escaped.

lib/mongo/collection/view/iterable.rb
```diff
@@ -34,24 +34,49 @@ module Iterable
...
-          @cursor = nil
-          session = client.send(:get_session, @options)
-          @cursor = if respond_to?(:write?, true) && write?
-            server = server_selector.select_server(cluster, nil, session)
-            result = send_initial_query(server, session)
-            Cursor.new(view, result, server, session: session)
-          else
-            read_with_retry_cursor(session, server_selector, view) do |server|
-              send_initial_query(server, session)
+          if system_collection? || !QueryCache.enabled?
```
p-mongo, line 39: Can this condition be inverted such that the "if" branch is the one with the cache enabled and the "else" branch is the one without the cache enabled please.

spec/mongo/cached_cursor_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  let(:reply) do
+    view.send(:send_initial_query, server)
+  end
+
+  let(:cursor) do
+    described_class.new(view, reply, server)
+  end
+
+  context 'when query cache enabled documents are cached' do
```
egiurleo, line 30: I think the description in these context blocks can be limited just to a description of the context (e.g. `context 'when the query cache is enabled' do`.

spec/mongo/cached_cursor_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    let(:documents) do
+      (1..3).map{ |i| { field: "test#{i}" }}
+    end
+
+    before do
+      authorized_collection.insert_many(documents)
+    end
+
+    context 'when query cache enabled' do
+
```
egiurleo, line 45: I believe this context block is unnecessary.

spec/mongo/cached_cursor_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      it 'docs are cached when query cache enabled and query is repeated' do
```
egiurleo, line 46: The descriptions for `it` blocks can just explain the behavior of the test (e.g. `it 'caches the result documents'`).

spec/mongo/cached_cursor_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+        expect(cursor.to_a.count).to eq(3)
+        expect(cursor.get_cached_docs.count).to eq(3)
+
+        expect(cursor.to_a.count).to eq(3)
+        expect(cursor.get_cached_docs.count).to eq(3)
+      end
+    end
+  end
+
+  context 'when query has results' do
```
egiurleo, line 57: I believe this context is unnecessary.

spec/mongo/cached_cursor_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    before do
+      3.times do |i|
+        authorized_collection.insert_one(test: i)
+      end
+    end
+
+    let(:view) do
+      Mongo::Collection::View.new(authorized_collection, {},
+        sort: {test: 1}, projection: {_id: 0}, batch_size: 2)
+    end
```
egiurleo, line 68: You can move these into the context below since they're only used for those tests :)

spec/mongo/query_cache_spec.rb
```diff
@@ -0,0 +1,313 @@
...
+
+  describe '#enabled' do
+
+    context 'when query cache is disabled' do
+
+      before do
+        Mongo::QueryCache.enabled = false
+      end
+
+      it 'query cache is disabled' do
```
egiurleo, line 32: I'd like you to make sure that all your `it` statements in the tests read like a sentence -- e.g. `it 'disables the query cache'`

spec/mongo/query_cache_spec.rb
```diff
@@ -0,0 +1,313 @@
...
+    authorized_client.tap do |client|
+      client.subscribe(Mongo::Monitoring::COMMAND, subscriber)
+    end
+  end
+
+  let(:authorized_collection) { client['collection_spec'] }
+
+  describe '#enabled' do
+
+    context 'when query cache is disabled' do
```
egiurleo, line 26: Could you add another test for enabling the query cache?

spec/mongo/query_cache_spec.rb
```diff
@@ -0,0 +1,313 @@
...
+
+      it 'query cache is disabled' do
+        expect(Mongo::QueryCache.enabled?).to be(false)
+      end
+    end
+  end
+
+  describe '#cache' do
+
+    context 'when a block is uncached' do
```
egiurleo, line 40: This context is unnecessary.

spec/mongo/query_cache_spec.rb
```diff
@@ -0,0 +1,313 @@
...
+
+      before do
+        authorized_collection.insert_one({ name: 'testing' })
+      end
+
+      it 'block is executed with disabled query cache' do
+        Mongo::QueryCache.cache {
+          authorized_collection.find(name: 'testing')
+          expect(Mongo::QueryCache.enabled?).to be(true)
+        }
```
egiurleo, line 50: It might also be good to set `Mongo::QueryCache.enabled = false` before the start of this test and then check that enabled is false again after the block is done executing.

spec/mongo/query_cache_spec.rb
```diff
@@ -0,0 +1,313 @@
...
+  describe '#cache' do
+
+    context 'when a block is uncached' do
+
+      before do
+        authorized_collection.insert_one({ name: 'testing' })
+      end
+
+      it 'block is executed with disabled query cache' do
+        Mongo::QueryCache.cache {
```
egiurleo, line 47: Please use `do/end` for multi-line blocks.

spec/mongo/query_cache_spec.rb
```diff
@@ -0,0 +1,313 @@
...
+          authorized_collection.find(name: 'testing')
+          expect(Mongo::QueryCache.enabled?).to be(true)
+        }
+      end
+    end
+  end
+
+  describe '#uncached' do
+
+    context 'when a block is uncached' do
```
egiurleo, line 57: All my comments from the `#cached` tests apply here :)

spec/mongo/query_cache_spec.rb
```diff
@@ -0,0 +1,313 @@
...
+
+    context 'when first query has a limit' do
+
+      before do
+        authorized_collection.find(limit: 2).to_a
+      end
+
+      let(:events) do
+        subscriber.command_started_events('find')
+      end
```
egiurleo, line 170: You can de-duplicate some code here by putting this code around line 72, which makes it available to all the tests in this context.

lib/mongo/cached_cursor.rb
```diff
@@ -0,0 +1,66 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module Mongo
+
+  # A Cursor that attempts to load documents from memory first before hitting
+  # the database if the same query has already been executed.
+  #
+  # @api semiprivate
+  class CachedCursor < Cursor
```
p-mongo, line 21: Can the class be renamed to CachingCursor please?

lib/mongo/collection/view/iterable.rb
```diff
@@ -35,15 +35,32 @@ module Iterable
...
-            result = send_initial_query(server, session)
-            Cursor.new(view, result, server, session: session)
+          if QueryCache.enabled?
+            unless @cursor = cached_cursor
+              @cursor = nil
+              session = client.send(:get_session, @options)
+              @cursor = if respond_to?(:write?, true) && write?
+                server = server_selector.select_server(cluster, nil, session)
+                result = send_initial_query(server, session)
+                CachedCursor.new(view, result, server, session: session)
```
p-mongo, line 45: Can this be DRYed with a helper method to eliminate duplication please.

spec/mongo/query_cache_spec.rb
```diff
@@ -0,0 +1,329 @@
...
+  describe '#uncached' do
+
+    before do
+      authorized_collection.insert_one({ name: 'testing' })
+    end
+
+    it 'executes the block with the query cache disabled' do
+      Mongo::QueryCache.uncached do
+        authorized_collection.find(name: 'testing').to_a
+        expect(Mongo::QueryCache.enabled?).to be(false)
```
p-mongo, line 82: This should also have an assertion on the query being issued, right?

spec/mongo/query_cache_spec.rb
```diff
@@ -0,0 +1,341 @@
...
+      Mongo::QueryCache.enabled = false
+      authorized_collection.insert_one({ name: 'testing' })
+      authorized_collection.find(name: 'testing').to_a
+    end
+
+    let(:events) do
+      subscriber.command_started_events('find')
+    end
+
+    it 'executes the block with the query cache enabled' do
```
egiurleo, line 61: These tests are looking good! I think this might be a good time to talk about integration vs. unit tests. 

Unit tests generally test just one method (aka a unit lol) to check that you get the expected output if you give a certain input. Integration tests test larger features or groups of components for behavior.

For the `#cache` method, a solid unit test might be something like:

```
it 'turns the query cache on inside the block' do
  Mongo::QueryCache.cache do
    expect(Mongo::QueryCache.enabled?).to be true
  end

  expect(Mongo::QueryCache.enabled?).to be false
end
```

I think the tests you have for `#cache` and `#uncached` (as well as the querying tests in the rest of this file) are really good integration tests, and I was wondering what you'd think about moving them over to a file at `spec/integration/query_cache_spec.rb`?

spec/mongo/query_cache_spec.rb
```diff
@@ -0,0 +1,341 @@
...
+    context 'when first query has no limit' do
+
+      before do
+        authorized_collection.find.to_a.count
+      end
+
+      context 'when next query has a limit' do
+
+        it 'uses the cache' do
+          authorized_collection.find({}, limit: 5).to_a.count
```
egiurleo, line 181: Could you add a check that this only returns 5 documents?

lib/mongo/caching_cursor.rb
```diff
@@ -0,0 +1,75 @@
...
+  #
+  # @api semiprivate
+  class CachingCursor < Cursor
+
+    # @return [ Array <BSON::Document> ] The cursor's cached documents.
+    # @api private
+    attr_reader :cached_docs
+
+    # @return [ Integer ] The max number of docs to return from the query.
+    attr_accessor :limit_docs
```
egiurleo, line 28: Correct me if I'm wrong here -- it seems like this is the limit on the view? If so, you can eliminate this method because the `Cursor` class already has a `limit` method. Let me know if I'm missing something and there's some reason you can't use that method.

lib/mongo/collection/view/iterable.rb
```diff
@@ -37,14 +37,18 @@ module Iterable
...
-              send_initial_query(server, session)
+          if QueryCache.enabled?
+            unless @cursor = cached_cursor
+              @cursor = select_cursor(session)
+              QueryCache.cache_table[cache_key] = @cursor
+            end
+            if limit
+              @cursor.limit_docs = limit
+            else
+              @cursor.limit_docs = nil
```
egiurleo, line 48: If my other comment about the `limit_docs` method is correct, you should also be able to remove this conditional statement.

lib/mongo/caching_cursor.rb
```diff
@@ -0,0 +1,75 @@
...
+    #   cursor.inspect
+    #
+    # @return [ String ] A string representation of a +Cursor+ instance.
+    def inspect
+      "#<Mongo::CachingCursor:0x#{object_id} @view=#{@view.inspect}>"
+    end
+
+    private
+
+    def process(result)
```
p-mongo, line 65: Can this method please get a comment explaining what it does, including the current implementation of query cache with respect to result sets exceeding batch sizes.

lib/mongo/collection/view/iterable.rb
```diff
@@ -37,18 +37,26 @@ module Iterable
...
-            read_with_retry_cursor(session, server_selector, view) do |server|
-              send_initial_query(server, session)
+          if QueryCache.enabled?
+            unless @cursor = cached_cursor
+              @cursor = select_cursor(session)
+              QueryCache.cache_table[cache_key] = @cursor
+            end
+            if limit
+              range = limit
             end
```
egiurleo, line 47: The final nitpick -- this can be written as `range = limit || nil`

PR #2032, opened 2020-07-29
RUBY-2342 Implement proper AWS auth region calculation


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2342

PR #2038, opened 2020-08-04
Fix RUBY-2346 Tag names are downcased when given in URI options


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2346

PR #2042, opened 2020-08-06
Fix RUBY-2347 GSSAPI service name is not defaulted when any other auth mechanism properties are set


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2347

PR #2043, opened 2020-08-06
RUBY-2293 Document retrieving auth source from TXT records with SRV URIs


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2293

PR #2045, opened 2020-08-06
RUBY-2299 support per-operation write concern for db and collection w…

gjchong25:
…rite methods

### DIFF
--- a/lib/mongo/collection.rb
+++ b/lib/mongo/collection.rb
@@ -239,6 +239,11 @@ module Mongo
       operation.delete(:write)
       operation.delete(:write_concern)
       client.send(:with_session, opts) do |session|
+        write_concern = if opts[:write_concern]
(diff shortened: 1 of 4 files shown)


lib/mongo/collection.rb
```diff
@@ -239,6 +239,12 @@ def create(opts = {})
       operation.delete(:write)
       operation.delete(:write_concern)
       client.send(:with_session, opts) do |session|
+        temp = write_concern
+        write_concern = if opts[:write_concern]
+          WriteConcern.get(opts[:write_concern])
+        else
+          temp
```
p-mongo, line 246: You should be able to replace this with `self.write_concern` and get rid of line 242.

p-mongo: For this and the other methods, can you please add write_concern as an accepted option to the docstring.

PR #2046, opened 2020-08-06
RUBY-1672 Document passing read concern to find operations in tutorial


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-1672

PR #2047, opened 2020-08-07
RUBY-1632 Handle bogus/invalid read concern sub-options

gjchong25:


lib/mongo/client.rb
```diff
@@ -1103,6 +1103,18 @@ def cluster_modifying?(new_options)
     # but does not check for interactions between combinations of options.
     def validate_new_options!(opts = Options::Redacted.new)
       return Options::Redacted.new unless opts
+      if opts[:read_concern]
+        given_keys = opts[:read_concern].keys
+        allowed_keys = ['level']
+        # raise an error for non user-settable options
+        if opts[:read_concern][:after_cluster_time]
+          raise Mongo::Error::OperationFailure.new('after_cluster_time is not a user-settable option')
```
p-mongo, line 1111: This needs to be a different exception class since OperationFailure is for errors produced by the server. I am not seeing one that we have that would fit, perhaps something like InvalidOption could be appropriate.

lib/mongo/client.rb
```diff
@@ -1103,6 +1103,18 @@ def cluster_modifying?(new_options)
...
+        end
+        # warn that options are invalid but keep it and forward to the server
+        if given_keys != allowed_keys
+          log_warn("Read concern has invalid key.")
```
p-mongo, line 1115: For this to be more helpful/actionable I would say at least the actual options given should be included, as well as what operation is being attempted.

lib/mongo/client.rb
```diff
@@ -1112,8 +1112,26 @@ def cluster_modifying?(new_options)
...
       return Options::Redacted.new unless opts
+      if opts[:read_concern]
+        # Raise an error for non user-settable options
+        if opts[:read_concern][:after_cluster_time]
+          raise Mongo::Error::InvalidReadConcern.new(
+            'The after_cluster_time read_concern option cannot be specified by the user'
+          )
+        end
+
+        given_keys = opts[:read_concern].keys
```
p-mongo, line 1123: This is relying on the strings being keys which might change due to Ruby 3 work, can there be an additional `.map(&:to_s)` to stringify them?

lib/mongo/error/invalid_read_concern.rb
```diff
@@ -0,0 +1,28 @@
...
+
+module Mongo
+  class Error
+
+    # Raised when an invalid read concern is provided.
+    class InvalidReadConcern < Error
+      # Instantiate the new exception.
+      def initialize(msg = nil)
+        super(msg || 'Invalid read concern option provided.' \
+              'The only valid key is level, for which accepted values are' \
```
p-mongo, line 23: I would use `:level` here.

p-mongo:
Also some/all of the commits seem to be listed twice, rebase on master?

lib/mongo/client.rb
egiurleo: Good point!

PR #2048, opened 2020-08-08
Fix RUBY-2252 Bulk write documentation references nonexistent insert_many operation


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2252

PR #2051, opened 2020-08-11
RUBY-2356 Set default logger level to info

p-mongo:
https://jira.mongodb.com/browse/RUBY-2356


p-mongo:
https://jira.mongodb.org/browse/RUBY-2356

PR #2058, opened 2020-08-13
RUBY-2359 Implement OCSP URI options


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2359

PR #2059, opened 2020-08-14
RUBY-1131 Ensure that split bulk writes don't produce command failed event

egiurleo:
This behavior was documented in [this Google groups question](https://groups.google.com/g/mongodb-user/c/o1H5Q0PQJjs/m/jb2Vr4JDEQAJ?pli=1).

### DIFF
--- a/spec/integration/bulk_write_spec.rb
+++ b/spec/integration/bulk_write_spec.rb
@@ -16,4 +16,35 @@ describe 'Bulk writes' do
       end.not_to raise_error
     end
   end
+
+  context 'when bulk write needs to be split' do
+    let(:subscriber) { EventSubscriber.new }
+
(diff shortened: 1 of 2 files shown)


spec/integration/bulk_write_spec.rb
```diff
@@ -16,4 +16,35 @@
...
+    let(:insert_events) do
+      subscriber.command_started_events('insert')
+    end
+
+    let(:failed_events) do
+      subscriber.failed_events
+    end
+
+    let(:operations) do
+      [{ insert_one: { text: 'a' * (max_bson_size/2) } }] * 3
```
p-mongo, line 34: We appear to be splitting at server command size limit (48m rather than 16m) so bumping this to 6 would make the tests green.

PR #2060, opened 2020-08-15
MONGOID-4968: WIP: Thread-safe implicit sessions

johnnyshields:
Currently, sessions in the driver are not thread-safe.

Session thread-safety is currently handled in Mongoid. This thread safety belongs in the driver. Having moved the QueryCache to the driver, we now have a precedent for using `Thread.current[]` key-value storage in the driver code.

This PR is still a proof-of-concept.

**TODO:** The `Client#start_session` method is now problematic, as it calls the implicit session code with `{ implicit: false }`. This may cause strange behavior and at a minimum makes the code difficult to read. I recommend to deprecate the block form of `Client#start_session` in favor of the other existing method `Client#with_session`.

### DIFF
--- a/lib/mongo/client.rb
+++ b/lib/mongo/client.rb
@@ -1025,11 +1025,40 @@ module Mongo
       yield session
     ensure
       if session && session.implicit?
-        session.end_session
+        clear_implicit_session
       end
     end
 
     private
+
+    # Starts a new implicit session, persists it on the thread, and returns it.
+    # Raises an error if there is already an implicit session in use on the current thread.
+    def start_implicit_session(options = {})
+      if get_implicit_session
+        raise Mongo::Error::InvalidSession.new(:invalid_session_nesting)
+      end
+
+      session = Session.new(cluster.session_pool.checkout, self, { implicit: true }.merge(options))
+      set_implicit_session(session)
+      session
+    end
+
+    # Gets the implicit session for the current thread
+    def get_implicit_session
+      Thread.current["[mongo]:session"]
+    end
+
+    # Sets the implicit session for the current thread
+    def set_implicit_session(session)
+      Thread.current["[mongo]:session"] = session
+    end
+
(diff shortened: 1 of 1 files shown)


p-mongo:
Implicit session is a session that is used by the driver when performing an operation against a 3.6+ server and the application has not specified a session. The application-specified sessions are called explicit sessions.

One use case for implicit sessions is to kill all sub-operations that are involved in a single application-issued operation across a sharded cluster. For this reason all crud operations on 3.6+ servers are associated with sessions, either implicit or explicit.

Thread-local storage would only apply to explicit sessions, the ones that application explicitly creates.

johnnyshields:
@p-mongo are you sure about that? In Mongoid, the method `Document#with_session` creates an "implicit" session today--i.e. a session which only exists fort he scope of that block, and is not otherwise referenceable.

p-mongo:
The terms "implicit session" and "explicit session" are defined in https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#explicit-vs-implicit-sessions.

p-mongo:
@johnnyshields If you are interested in working on thread-local session storage, I suggest the following:

1. Edit the documentation at https://docs.mongodb.com/ruby-driver/master/tutorials/ruby-driver-sessions/ to show an example that would use thread-local sessions. This could be modeled after the current Mongoid API.
2. Add a test case to https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/mongo/session_spec.rb that mirrors the usage example. This would need to be an integration test eventually, you could either write it as an integration test (in a new file in https://github.com/mongodb/mongo-ruby-driver/tree/master/spec/integration) or edit the existing session spec and I'll move it later.
3. Implement enough of the functionality that the new test passes, and all existing tests also pass.

p-mongo:
Closing per the above comment.

PR #2064, opened 2020-08-16
RUBY-2355 Use server error codes in place of error messages when possible


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2355

PR #2072, opened 2020-08-24
RUBY-2338 Stop query cache from caching multi-batch queries

egiurleo:
This code will further be modified to support caching multi-batch queries, but for now, I have eliminated the `Cannot restart iteration of a cursor which issued a getMore` exception from occurring when a user tries to iterate through a multi-batch result multiple times.

### DIFF
--- a/lib/mongo/collection/view/iterable.rb
+++ b/lib/mongo/collection/view/iterable.rb
@@ -40,11 +40,12 @@ module Mongo
 
           if QueryCache.enabled? && @cursor.is_a?(Mongo::CachingCursor)
             QueryCache.cache_table[cache_key] = @cursor
+            range = limit || nil
           end
 
           if block_given?
-            if limit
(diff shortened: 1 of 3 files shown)


lib/mongo/collection/view/iterable.rb
```diff
@@ -35,24 +35,21 @@ module Iterable
...
             range = limit || nil
-          else
-            @cursor = select_cursor(session)
           end
+
           if block_given?
-            if !range
-              @cursor.each do |doc|
+            if range
+              @cursor.to_a[0...range].each do |doc|
```
p-mongo, line 48: I believe I had to figure out what this does exactly previously, if this code stays around can some comments be added for this line and maybe other lines in this method explaining what the logic is?

egiurleo: Agreed. I made a comment on RUBY-2367 to add an explanation here if this code still exists.

PR #2079, opened 2020-09-03
RUBY-2352 Add try_next example to resuming change stream user guide


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2352

egiurleo (approved):

PR #2081, opened 2020-09-10
RUBY-2340 Part 1: Refactor the query cache to implement read and write methods

egiurleo:
This PR does not change any behavior but implements methods in the query cache that abstract some of the behavior of creating a cache key and reading from/writing to the cache. These methods will be used to implement efficient limits in the query cache.

### DIFF
--- a/lib/mongo/collection/view/iterable.rb
+++ b/lib/mongo/collection/view/iterable.rb
@@ -39,7 +39,7 @@ module Mongo
           @cursor = select_cursor(session)
 
           if QueryCache.enabled? && @cursor.is_a?(Mongo::CachingCursor)
-            QueryCache.write(@cursor, cache_options)
+            QueryCache.set(@cursor, cache_options)
             range = limit || nil
           end
 
@@ -104,7 +104,7 @@ module Mongo
         end
 
         def cached_cursor
-          QueryCache.read(cache_options)
+          QueryCache.get(cache_options)
         end
 
         def cache_options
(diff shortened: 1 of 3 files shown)


lib/mongo/collection/view/iterable.rb
```diff
@@ -39,7 +39,7 @@ def each
           @cursor = select_cursor(session)
 
           if QueryCache.enabled? && @cursor.is_a?(Mongo::CachingCursor)
-            QueryCache.cache_table[cache_key] = @cursor
+            QueryCache.write(@cursor, cache_options)
```
p-mongo, line 42: What do you think about get/set instead of read/write?

lib/mongo/query_cache.rb
```diff
@@ -83,6 +83,41 @@ def cache_table
       def clear_cache
         Thread.current["[mongo]:query_cache"] = nil
       end
+
+      def write(cursor, options = {})
```
p-mongo, line 95: Are these public API?

lib/mongo/collection/view/iterable.rb
egiurleo: I'm totally happy with that!

lib/mongo/query_cache.rb
egiurleo: They are currently, but they don't have to be! Do you think it would be better to make them private?

p-mongo: I think the methods that take/return cursors should default to private.

p-mongo (approved):

PR #2083, opened 2020-09-10
RUBY-2340 Update the query cache to efficiently handle limits

egiurleo:
The goal of this PR is to have the query cache reuse the same query results when a query with the same or a lower limit as the stored results is executed.

Here is my general design:
* Remove the limit from the cache key
* When reading from the cache, check the limit on the cursor's view. If that limit is greater than or equal to the limit of the query currently being executed, then return that cached cursor.
* The query cache does not store more than one result for the same query with different limits. It stores the result thus far with the most permissive limit (either a large number or nil).

Let me know if you think there needs to be more comments in the code!

### DIFF
--- a/lib/mongo/collection/view/iterable.rb
+++ b/lib/mongo/collection/view/iterable.rb
@@ -39,7 +39,9 @@ module Mongo
           @cursor = select_cursor(session)
 
           if QueryCache.enabled? && @cursor.is_a?(Mongo::CachingCursor)
-            QueryCache.set(@cursor, cache_options)
+            # No need to store the cursor in the query cache if there is
+            # already a cached cursor stored at this key.
+            QueryCache.set(@cursor, cache_options) unless cached_cursor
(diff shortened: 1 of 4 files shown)


p-mongo:
With query cache enabled, when a query is made with limit 1 and subsequently two queries are made with limit 2, how many queries are issued to the database in total?

I also think as of this PR the assertions should check that the returned data is correct (via exact comparison to expected data) and not just how many documents were returned.

egiurleo:
@p That's a good idea! I'll add those assertions.

For the scenario you mentioned, 2 queries would be made in total -- one for limit 1, one for limit 2, the limit 2 results would overwrite the limit 1 results, and then the second limit 2 query would read from the cache.

p-mongo:
:+1:  That's great.

p-mongo (approved):

p-mongo (approved):

PR #2086, opened 2020-09-11
RUBY-2363 Clear query cache on transaction commit and abort

egiurleo:


spec/integration/query_cache_spec.rb
```diff
@@ -513,4 +513,113 @@
       expect(events.length).to eq(2)
     end
   end
+
+  describe 'in transactions' do
+    min_server_fcv '4.0'
+    require_topology :replica_set
```
p-mongo, line 519: We have a `require_transaction_support` for this.

lib/mongo/session.rb
```diff
@@ -602,6 +603,8 @@ def commit_transaction(options=nil)
     #
     # @since 2.6.0
     def abort_transaction
+      Mongo::QueryCache.clear_cache
```
p-mongo, line 606: `Mongo::` shouldn't be needed here.

p-mongo (commented):
If one issues a query outside a transaction, starts a transaction and repeats the query, the query isn't repeated right? But if tx has snapshot read concern it would be repeated?

There should also be a docs update stating we clear cache on commit & abort, is that covered by another ticket?

spec/integration/query_cache_spec.rb
egiurleo: So convenient! Thanks for letting me know.

egiurleo:
@p I added tests for both scenarios you mentioned and they work as you described. I also added a bullet point to [RUBY-2376](https://jira.mongodb.org/browse/RUBY-2376) about adding documentation for using the query cache in transactions.

p-mongo (approved):

spec/integration/query_cache_spec.rb
```diff
@@ -513,4 +513,155 @@
...
+    require_transaction_support
+
+    let(:collection) { authorized_client['test'] }
+
+    let(:events) do
+      subscriber.command_started_events('find')
+    end
+
+    before do
+      collection.delete_many
```
p-mongo, line 527: What do you think about making a helper like `create_collection` that would drop & create? Should take care of red CI then I expect.

PR #2092, opened 2020-09-15
Unify evergreen configs between standard configurations and atlas


p-mongo:


p-mongo:
This PR moves all evergreen config bits into template files in `.evergreen/config` and generates the two final configurations from those. The generated configuration files remain committed in version control.

The primary intended benefit of this change is programmatic generation of ocsp configurations.

A later benefit may be merging of atlas configurations into the main projects.

egiurleo (approved):

PR #2094, opened 2020-09-16
RUBY-2387 Mark operation result as a public API


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2387

egiurleo (commented):
I agree with this in theory -- do we have information in our documentation about what semiprivate means?

p-mongo:
I am unaware of a definition. What I mean by it is "private unless declared otherwise", although for all of the classes/modules in question I labeled everything with language public visibility as api public or private explicitly. What are your thoughts as to what is needed for semiprivate?

egiurleo (approved):

PR #2095, opened 2020-09-17
RUBY-2367 Allow CachingCursor to cache multi-batch queries

egiurleo:
I did this by modifying the CachingCursor so that it adds document to `@cached_docs` on `try_next` rather than `process`. This way, the cursor can iterate through the `@cached_docs` and then use `try_next` to iterate through the remaining documents, calling `get_more` if there are other batches of documents that haven't been fetched.

### DIFF
--- a/lib/mongo/caching_cursor.rb
+++ b/lib/mongo/caching_cursor.rb
@@ -63,11 +63,8 @@ module Mongo
     # get_more_operation.
     def process(result)
       documents = super
-      if @cursor_id.zero? && !@after_first_batch
-        @cached_docs ||= []
-        @cached_docs.concat(documents)
-      end
-      @after_first_batch = true
+      @cached_docs ||= []
+      @cached_docs.concat(documents)
       documents
     end
   end
--- a/lib/mongo/caching_cursor.rb
+++ b/lib/mongo/caching_cursor.rb
@@ -32,9 +32,17 @@ module Mongo
     #     # ...
     #   end
     def each
-      if @cached_docs && closed?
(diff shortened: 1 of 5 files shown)


lib/mongo/caching_cursor.rb
```diff
@@ -51,24 +59,16 @@ def inspect
...
-        @cached_docs ||= []
-        @cached_docs.concat(documents)
-      end
-      @after_first_batch = true
-      documents
+    # @api private
+    def try_next
+      @cached_docs ||= []
+      document = super
+      @cached_docs << document
```
p-mongo, line 69: super can return nil which I don't think should be inserted here?

egiurleo: That's a great call. I'll add a test for this case.

spec/integration/query_cache_spec.rb
```diff
@@ -177,6 +243,27 @@
...
+    end
+
+    # When the last batch runs out, try_next will return nil instead of a
+    # document. This test checks that nil is not added to the list of cached
+    # documents or returned as a result.
+    it 'returns the correct response' do
+      expect(authorized_client['test'].find({}, batch_size: 2).to_a).to eq(expected_result)
+      expect(authorized_client['test'].find({}, batch_size: 2).to_a).to eq(expected_result)
+    end
+  end
```
egiurleo, line 265: @p I added a test for the case you mentioned earlier!

p-mongo (approved):

PR #2096, opened 2020-09-17
RUBY-2377 Cache aggregations and clear the cache when $out or $merge stages provided

egiurleo:
This PR does two things:
1. It allows aggregations to be cached. Previously, trying to aggregate with the QueryCache turned on would have caused an error. Now, it caches the aggregation, taking into account read concern, read preference, and collation options. (All other options are specified in the pipeline).
2. It clears the query cache when $out or $merge stages are specified in an aggregation, since those stages can write to the collection.

Documentation on aggregation caching will be provided in a later ticket.

### DIFF
--- a/lib/mongo/collection/view/aggregation.rb
+++ b/lib/mongo/collection/view/aggregation.rb
@@ -89,6 +89,16 @@ module Mongo
           self.class.new(view, pipeline, options.merge(explain: true)).first
         end
 
+        # Whether this aggregation will write its result to a database collection.
+        #
+        # @return [ Boolean ] Whether the aggregation will write its result
+        #   to a collection.
+        #
+        # @api private
+        def write?
+          pipeline.any? { |op| op.key?('$out') || op.key?(:$out) || op.key?('$merge') || op.key?(:$merge) }
+        end
+
         private
 
         def server_selector
@@ -112,10 +122,6 @@ module Mongo
             connection.description
           end
           description.standalone? || description.mongos? || description.primary? || secondary_ok?
-        end
-
-        def write?
-          pipeline.any? { |op| op.key?('$out') || op.key?(:$out) || op.key?('$merge') || op.key?(:$merge) }
         end
 
         def secondary_ok?
--- a/lib/mongo/collection/view/aggregation.rb
+++ b/lib/mongo/collection/view/aggregation.rb
@@ -137,15 +137,16 @@ module Mongo
           end
         end
 
+        # Skip, sort, limit, projection are specified as pipeline stages
+        # rather than as options.
         def cache_options
           {
             namespace: collection.namespace,
             selector: pipeline,
+            read_concern: view.read_concern,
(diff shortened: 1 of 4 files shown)


lib/mongo/collection/view/aggregation.rb
```diff
@@ -136,6 +136,18 @@ def validate_collation!(server)
             raise Error::UnsupportedCollation.new
           end
         end
+
+        # Skip, sort, limit, projection are specified as pipeline stages
+        # rather than as options.
+        def cache_options
+          {
+            namespace: collection.namespace,
+            selector: pipeline,
```
p-mongo, line 145: Is selector then a hash for queries and array for agg pipelines? Is this documented somewhere?

lib/mongo/collection/view/iterable.rb
```diff
@@ -42,12 +42,18 @@ def each
...
+            # re-use results from an earlier query with the same or larger
+            # limit, and then impose the lower limit during iteration.
+            limit_for_cached_query = respond_to?(:limit) ? limit : nil
           end
 
           if block_given?
-            if range
-              @cursor.to_a[0...range].each do |doc|
+            # Ruby versions 2.5 and older do not support arr[0..nil] syntax, so
+            # this must be a separate conditional.
```
p-mongo, line 54: Would slice work? https://ruby-doc.org/core-2.7.1/Array.html#method-i-slice

lib/mongo/collection/view/readable.rb
```diff
@@ -45,6 +45,14 @@ module Readable
         #
         # @since 2.0.0
         def aggregate(pipeline, options = {})
+          # Because the $merge and $out pipeline stages write documents to the
+          # collection, it is necessary to clear the cache when they are performed.
+          write = pipeline.any? do |op| 
+            op.key?('$out') || op.key?(:$out) ||
+              op.key?('$merge') || op.key?(:$merge)
+          end
```
p-mongo, line 53: Aggregation has a `write?` method that does this, can it be used to decide whether to clear the cache?

p-mongo (approved):

lib/mongo/collection/view/readable.rb
egiurleo: Good point!

lib/mongo/collection/view/iterable.rb
egiurleo: It looks like `slice` only takes Integer arguments (get a `cannot convert nil to integer` exception).

lib/mongo/collection/view/aggregation.rb
egiurleo: Yes precisely! Not currently documented anywhere but I added a note to RUBY-2376.

lib/mongo/collection/view/iterable.rb
p-mongo: :+1:

PR #2098, opened 2020-09-22
RUBY-2375 More efficient query cache invalidation

egiurleo:
Currently, the query cache completely clears on every write operation. However, most write operations only impact one database and collection, so it doesn't make sense to clear every single cursor from the cache.

This PR introduces the following changes:
* The cache is now structured in two levels: the first-level key is the namespace (database.collection), and every value is a Hash containing CachingCursors.
* Instead of clearing the whole query cache, most write operations only clear the namespace that they're modifying.
* Aggregation results are cached at the `nil` namespace because aggregation pipelines can read data from multiple namespaces.
* Aggregation operations that write to a collection ($out and $merge stages) clear the entire cache.

### DIFF
--- a/lib/mongo/query_cache.rb
+++ b/lib/mongo/query_cache.rb
@@ -117,9 +117,13 @@ module Mongo
       # @api private
       def get(options = {})
         limit = options[:limit]
+        namespace = options[:namespace]
         key = cache_key(options)
 
-        caching_cursor = QueryCache.cache_table[key]
+        namespace_hash = QueryCache.cache_table[namespace]
+        return nil unless namespace_hash
+
+        caching_cursor = namespace_hash[key]
         return nil unless caching_cursor
 
         caching_cursor_limit = caching_cursor.view.limit
--- a/lib/mongo/query_cache.rb
+++ b/lib/mongo/query_cache.rb
@@ -161,8 +161,8 @@ module Mongo
       private
 
       def cache_key(options)
-        unless options[:namespace] && options[:selector]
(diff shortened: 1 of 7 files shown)


lib/mongo/query_cache.rb
```diff
@@ -84,6 +84,24 @@ def clear_cache
...
+      # from this namespace.
+      #
+      # @param [ String ] namespace The namespace to be cleared, in the format
+      #   "database.collection".
+      #
+      # @return [ nil ] Always nil.
+      #
+      # @api private
+      def clear_namespace(namespace)
+        cache_table[namespace] = nil if cache_table[namespace]
```
p-mongo, line 97: I think this and the next call can be simplified to `cache_table.delete(namespace)`.

lib/mongo/query_cache.rb
```diff
@@ -114,9 +135,13 @@ def set(cursor, options = {})
       # @api private
       def get(options = {})
         limit = options[:limit]
+        namespace = options[:namespace]
```
p-mongo, line 138: Can you document the options on all of the methods please?

p-mongo (approved):
Is the intention to not allow users to clear a particular namespace?

lib/mongo/query_cache.rb
egiurleo: Thank you!

egiurleo:
Oleg and I chatted on Slack -- for now, we'll leave the clear_namespace method as api private, unless a user expresses a need to manually clear a namespace.

PR #2100, opened 2020-09-23
RUBY-2408 Invalidate the cache on bulk writes

egiurleo:


p-mongo (approved):
Is it possible for (non-bulk?) CRUD operations to be interleaved between when a bulk write is constructed and when it is actually executed?

egiurleo:
@p-mongo I believe it is -- do you foresee that impacting the query cache in any way?

egiurleo:
Oleg and I spoke on Slack -- he pointed out that there is a possibility for a scenario where a query is interleaved between the creation of a bulk write and its execution. If this feature was not properly implemented, something could happen like:

bulk insert instantiated on A -> query on A (writes to QC) -> bulk insert executed -> query on A

If the bulk write instantiation clears the cache and the bulk write execution does not, then the final query could pull stale data from the query cache. I will add a test for this shortly.

PR #2101, opened 2020-09-23
RUBY-2376 Query cache documentation

egiurleo:
I have moved the query cache documentation to its own page and linked to it from the CRUD page. I have gone in depth on various topics, such as cache invalidation, using the query cache with transactions and aggregations, and queries with limits.

### DIFF
--- a/docs/tutorials/query-cache.txt
+++ b/docs/tutorials/query-cache.txt
@@ -1,0 +1,177 @@
+***********
+Query Cache
+***********
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+The MongoDB Ruby driver provides a built-in query cache. When enabled, the
+query cache saves the results of previously-executed find and aggregation
+queries and returns those cached results when the same queries are executed
+to prevent unnecessary round trips to the database.
+
+Usage
+=====
+
+The query cache is disabled by default. It cache can be enabled on a gobal
+scope as well as within the context of a specific block.
+
+To enable the query cache globally:
+
+.. code-block:: ruby
+
+  Mongo::QueryCache.enabled = true
+
+Similarly, to disable it globally:
+
+.. code-block:: ruby
+
+  Mongo::QueryCache.enabled = false
+
+To enable the query cache within the context of a block:
+
+.. code-block:: ruby
+
+  Mongo::QueryCache.cache do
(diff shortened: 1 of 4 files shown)


docs/tutorials/query-cache.txt
```diff
@@ -0,0 +1,180 @@
...
+
+    client['artists'].find(genre: 'Rock', limit: 20)
+    #=> Queries the database again and replaces the previously cached query results
+  end
+
+Cache Invalidation
+==================
+
+The query cache is cleared in part or in full on every write operation. Most
+write operations will clear any queries were performed on the same
```
p-mongo, line 94: Suggest "will clear results of any queries that were performed"

docs/tutorials/query-cache.txt
```diff
@@ -0,0 +1,180 @@
...
+collection that is being written to. Some operations will clear the entire
+query cache.
+
+The following operations will clear cached queries on the same database and
```
p-mongo, line 98: Suggest "cached query results"

p-mongo (approved):
This is excellent!

Could you also add a section that talks about the scope of each cached result, to specifically address the lack of cache clearing when transaction is starting (i.e. so that users understand that if they start a tx with snapshot read concern, the existing entries won't be used, and if continue with the previous write concern they may end up reading cache entries in a transaction).

docs/tutorials/query-cache.txt
```diff
@@ -0,0 +1,180 @@
...
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+.. _query-cache:
+
+The MongoDB Ruby driver provides a built-in query cache. When enabled, the
+query cache saves the results of previously-executed find and aggregation
+queries. When those same queries are performed again, the driver returns
+the cached reuslts to prevent unnecessary round trips to the database.
```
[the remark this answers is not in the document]
egiurleo: You're right, it's "roundtrip."

docs/tutorials/query-cache.txt
```diff
@@ -0,0 +1,213 @@
...
+* Skip
+* Sort
+* Projection
+* Collation
+* Read Concern
+* Read Preference
+
+For example, if you perform one query, and then perform a mostly identical query
+with a different sort order, those queries will not be considered matching,
+and the second query will not use the cached results of the first.
```
egiurleo, line 89: @p-mongo What do you think of this section?

p-mongo: :+1:

p-mongo (approved):

PR #2102, opened 2020-09-24
RUBY-2410 Fix issue where same aggregation occuring on different collection returns cached result

egiurleo:
I introduced a bug in the query cache by storing aggregations in the `nil` namespace -- if aggregations with the same pipeline were performed in two different namespaces, then the second one would return the cached results of the first. This PR adds namespace back to the cache key and introduces a new option (`multi_collection`) which indicates that results could come from multiple namespaces and should be stored under the `nil` namespace key.

### DIFF
--- a/lib/mongo/query_cache.rb
+++ b/lib/mongo/query_cache.rb
@@ -125,13 +125,17 @@ module Mongo
       #   (e.g. { level: :majority }).
       # @option options [ Hash | nil ] read_preference The read preference of
       #   the query (e.g. { mode: :secondary }).
+      # @option options [ Boolean | nil ] multi_collection Whether the query
+      #   results could potentially come from multiple collections. When true,
+      #   these results will bes tored under the nil namespace key and cleared
+      #   on every write command.
       #
       # @return [ true ] Always true.
       #
       # @api private
       def set(cursor, options = {})
         key = cache_key(options)
-        namespace = options[:namespace]
+        namespace = namespace_key(options)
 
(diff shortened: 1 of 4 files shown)


lib/mongo/query_cache.rb
```diff
@@ -162,14 +166,18 @@ def set(cursor, options = {})
       #   (e.g. { level: :majority }).
       # @option options [ Hash | nil ] read_preference The read preference of
       #   the query (e.g. { mode: :secondary }).
+      # @option options [ Boolean | nil ] multi_collection Whether the query
+      #   results could potentially come from multiple collections. When true,
+      #   these results will bes tored under the nil namespace key and cleared
```
p-mongo, line 171: be stored

lib/mongo/query_cache.rb
```diff
@@ -200,11 +208,12 @@ def get(options = {})
       private
 
       def cache_key(options)
-        unless options[:selector]
-          raise ArgumentError.new("Cannot generate cache key without selector")
+        unless options[:namespace] && options[:selector]
+          raise ArgumentError.new("Cannot generate cache key without namesapce or selector")
```
p-mongo, line 212: namespace or

p-mongo (approved):

lib/mongo/query_cache.rb
egiurleo: I can't type today 😅 Thank you!

PR #2105, opened 2020-09-30
RUBY-2413 Add a test and documentation for system collections with query cache

egiurleo:
I have opted here not to add cache invalidation to the helpers that modify those systems collections as that would be a fairly significant project if I tried to do it exhaustively. Instead, I added a documentation section explaining that system collections use the query cache and the data will not be invalidated, so users should disable the query cache if they need up to date systems collection information.

### DIFF
--- a/spec/integration/query_cache_spec.rb
+++ b/spec/integration/query_cache_spec.rb
@@ -1020,4 +1020,41 @@ describe 'QueryCache' do
       expect(events.length).to eq(2)
     end
   end
+
+  context 'with system collection' do
+    let(:client) do
+      ClientRegistry.instance.global_client('root_authorized').tap do |client|
+        client.subscribe(Mongo::Monitoring::COMMAND, subscriber)
+      end
+    end
+
+    before do
+      begin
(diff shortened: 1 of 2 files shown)


p-mongo (approved):

spec/integration/query_cache_spec.rb
```diff
@@ -1030,15 +1014,47 @@
...
       expect(events.length).to eq(2)
     end
   end
 
+  context 'with system collection' do
+    let(:client) do
+      ClientRegistry.instance.global_client('root_authorized').tap do |client|
+        client.subscribe(Mongo::Monitoring::COMMAND, subscriber)
+      end
+    end
```
egiurleo, line 1029: @p-mongo is it okay that I'm using the root authorized client here?

p-mongo: Yes this makes sense to me, our authorized user I believe is scoped to the default test database hence to operate on users the root authorized user appears appropriate.

p-mongo (approved):

PR #2109, opened 2020-10-06
RUBY-2419 Exclude connectionId field when checking for equality

knovoselic:
`connectionId` does not represent state of the topology and as such should be excluded when comparing two Description objects.


p-mongo (approved):

egiurleo (approved):

p-mongo:
Thank you for the patch!

PR #2110, opened 2020-10-06
Fix RUBY-2421 Auth mechanism properties not downcased when client is created with host and Ruby options with symbol key

p-mongo:
https://jira.mongodb.com/browse/RUBY-2421


p-mongo:
https://jira.mongodb.org/browse/RUBY-2421

egiurleo (approved):

PR #2115, opened 2020-10-14
RUBY-2399 Provide more descriptive error message when inserting nil documents

ksadoff:
I added more detailed error messages for when users insert nil or non-hash documents (as described [here](https://jira.mongodb.org/browse/RUBY-2399)) and wrote corresponding unit test cases.

### DIFF
--- a/lib/mongo/collection.rb
+++ b/lib/mongo/collection.rb
@@ -566,7 +566,7 @@ module Mongo
         end
 
         if document.nil?
-          raise ArgumentError, "Inserted document cannot be nil"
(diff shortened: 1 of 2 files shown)


p-mongo (approved):

p-mongo:
How about "Document to be inserted cannot be nil" for the language instead? I think this would be more accurate.

egiurleo (approved):
Agreed with Oleg's comment!

egiurleo:
Awesome! Could you change the title of the PR as well to reflect that you're just handling the nil case?

PR #2116, opened 2020-10-16
RUBY-2211 Use namespace returned from find command in getMore

egiurleo:
I have manually tested this code against an Atlas Data Lake instance, and in a PR for RUBY-2325, I will set up our test suite to run with Data Lake.

### DIFF
--- a/lib/mongo/cursor.rb
+++ b/lib/mongo/cursor.rb
@@ -298,7 +298,7 @@ module Mongo
         # However, sometimes the collection name will contain periods, which
         # is why this method joins all the namespace components after the first.
         ns_components = @namespace.split('.')
-        ns_components[1...].join('.')
+        ns_components[1..ns_components.length].join('.')
       else
         collection.name
       end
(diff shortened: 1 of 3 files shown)


lib/mongo/cursor.rb
```diff
@@ -288,7 +288,20 @@ def close
...
+      # object in the driver. However, in some cases (e.g. when connected
+      # to an Atlas Data Lake), the namespace returned by the find command
+      # may be different, which is why we want to use the collection name based
+      # on the namespace in the command result.
+      if @namespace
+        # Often, the namespace will be in the format "database.collection".
+        # However, sometimes the collection name will contain periods, which
+        # is why this method joins all the namespace components after the first.
+        ns_components = @namespace.split('.')
+        ns_components[1..ns_components.length].join('.')
```
p-mongo, line 301: `..` includes the end of the range, this needs to either use `...` or be `..ns_components.length-1`.

egiurleo: Good catch!

p-mongo (approved):

PR #2118, opened 2020-10-19
Fix FLE tests

egiurleo:
It looks like libmongocrypt has changed some of its KMS endpoint validation behavior -- now calling the Amazon KMS endpoint via https is valid, and the data key context object also performs some endpoint validation. This does not impact the user experience with FLE.


p-mongo (approved):
Can you please make a ticket with the error message we are getting on master and a quote of the parameters we are using that results in that message?

PR #2120, opened 2020-11-03
Remove pending status from BSON Symbol test

egiurleo:
A couple tests were broken/fixed as a result of the bson jruby 4.11.0 release.

### DIFF
--- a/spec/support/shared/protocol.rb
+++ b/spec/support/shared/protocol.rb
@@ -5,7 +5,8 @@ shared_examples 'message with a header' do
     describe 'length' do
       let(:field) { bytes.to_s[0..3] }
       it 'serializes the length' do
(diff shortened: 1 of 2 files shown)


spec/support/shared/protocol.rb
```diff
@@ -5,7 +5,8 @@
     describe 'length' do
       let(:field) { bytes.to_s[0..3] }
       it 'serializes the length' do
-        expect(field).to be_int32(bytes.length)
+        bytes.rewind!
```
p-mongo, line 8: If this is only needed for JRuby can it be conditioned on `BSON::Environment.jruby?` ?

egiurleo: Sure thing!

p-mongo (approved):

PR #2123, opened 2020-11-05
RUBY-1387 Add global TLS context hooks

egiurleo:
This is a first attempt at adding TLS context hooks. I'm wondering if there are any particular cases we want to test that I haven't covered here?

- [x] Add TLS context hooks to Ruby driver
- [x] Once we agree on API and usage, add documentation to Ruby driver
- [ ] Add documentation to Mongoid


lib/mongo/tls_context_hooks.rb
```diff
@@ -0,0 +1,48 @@
...
+      def hooks=(hooks)
+        unless hooks.is_a?(Array) && hooks.all? { |hook| hook.is_a?(Proc) }
+          raise ArgumentError, "TLS context hooks must be an array of Procs"
+        end
+
+        @hooks=hooks
+      end
+
+      # Clear the TLS context hooks.
+      def clear_hooks
```
p-mongo, line 43: I don't think this is needed since one can call Array#clear.

spec/mongo/socket/ssl_spec.rb
```diff
@@ -50,6 +50,19 @@
   end
 
   describe '#connect!' do
+    context 'when TLS context hooks are provided' do
+      let(:proc) { Proc.new { |context| context.ciphers = ["AES256-SHA"] } }
+
+      before do
+        Mongo::TLSContextHooks.hooks = [ proc ]
```
p-mongo, line 57: I suggest `Mongo.tls_context_hooks = [ proc ]` as the API, either delegated to Mongo::TlsContextHooks or defined on Mongo module itself.

lib/mongo.rb
```diff
@@ -78,4 +78,27 @@ module Mongo
...
+  #
+  # @param [ Array<Proc> ] hooks An Array of Procs, each of which should take
+  #   an OpenSSL::SSL::SSLContext object as an argument.
+  module_function def tls_context_hooks=(hooks)
+    unless hooks.is_a?(Array) && hooks.all? { |hook| hook.is_a?(Proc) }
+      raise ArgumentError, "TLS context hooks must be an array of Procs"
+    end
+
+    @tls_context_hooks=hooks
+  end
```
egiurleo, line 103: @p-mongo Let me know if this is more what you had in mind.

docs/tutorials/ruby-driver-create-client.txt
```diff
@@ -1090,6 +1090,34 @@ file and both must be stored in the same file. Example usage:
...
+To set the TLS context hooks, use the ``Mongo.tls_context_hooks`` method, and
+provide an Array of ``Procs`` that will be run prior to connecting any SSL socket:
+
+.. code-block:: ruby
+
+  Mongo.tls_context_hooks = [
+    Proc.new { |context|
+      context.ciphers = ["AES256-SHA"]
+    }
+  ]
```
egiurleo, line 1113: Let me know if there's a more realistic example than this -- I'm not sure how people would actually want to use this.

egiurleo: ALSO, let me know if we should recommend including this code in any specific part of a user's application?

docs/tutorials/ruby-driver-create-client.txt
```diff
@@ -1090,6 +1090,34 @@ file and both must be stored in the same file. Example usage:
...
+by enabling or disabling certain ciphers. Currently, the Ruby driver does not
+provide a way to do this when initializing a ``Mongo::Client``.
+
+However, the Ruby driver provides a way to set global "TLS context hooks" --
+these are user-provided methods that will be invoked before any SSL socket
+connection and can be used to modify the underlying ``OpenSSL::SSL::SSLContext``
+object used by the socket.
+
+To set the TLS context hooks, use the ``Mongo.tls_context_hooks`` method, and
+provide an Array of ``Procs`` that will be run prior to connecting any SSL socket:
```
p-mongo, line 1105: Suggest `Proc`s (un-format the "s")

p-mongo: and "TLS socket"

docs/tutorials/ruby-driver-create-client.txt
p-mongo: The use case requested in the ticket was removing a cipher, but I don't know how that works exactly. If you'd like to research that I would suggest giving that example, but the way this example is now is also fine with me.

p-mongo: For the location, I can suggest:

> Set the TLS context hooks before creating any `Mongo::Client` instances. For example, in a Rails application this code could be placed in an initializer.

docs/tutorials/ruby-driver-create-client.txt
```diff
@@ -1090,6 +1090,34 @@ file and both must be stored in the same file. Example usage:
...
+
+  Mongo.tls_context_hooks = [
+    Proc.new { |context|
+      context.ciphers = ["AES256-SHA"]
+    }
+  ]
+
+  # Only the AES256-SHA cipher will be enabled from this point forward
+
+Every ``Proc`` in the array passed to ``Mongo.tls_context_hooks`` should take
```
p-mongo, line 1117: Suggest s/should take/will be passed/

docs/tutorials/ruby-driver-create-client.txt
```diff
@@ -1090,6 +1090,34 @@ file and both must be stored in the same file. Example usage:
...
+an ``OpenSSL::SSL::SSLContext`` object as its sole argument. These ``Procs`` will
```
p-mongo, line 1118: `Proc`s

docs/tutorials/ruby-driver-create-client.txt
```diff
@@ -1090,6 +1090,34 @@ file and both must be stored in the same file. Example usage:
...
+be executed sequentially on the creation of every ``Mongo::Socket::SSL`` object.
```
p-mongo, line 1119: Suggest s/on the/during the/

spec/mongo/socket/ssl_spec.rb
```diff
@@ -50,6 +50,19 @@
   end
 
   describe '#connect!' do
+    context 'when TLS context hooks are provided' do
+      let(:proc) { Proc.new { |context| context.ciphers = ["AES256-SHA"] } }
+
+      before do
+        Mongo.tls_context_hooks = [ proc ]
```
p-mongo, line 57: The hooks need to be cleared afterwards, right?

lib/mongo.rb
p-mongo: Yes I think this is all that is needed.

lib/mongo.rb
```diff
@@ -78,4 +78,27 @@ module Mongo
...
+  # Set the TLS context hooks.
+  #
+  # @param [ Array<Proc> ] hooks An Array of Procs, each of which should take
+  #   an OpenSSL::SSL::SSLContext object as an argument.
+  module_function def tls_context_hooks=(hooks)
+    unless hooks.is_a?(Array) && hooks.all? { |hook| hook.is_a?(Proc) }
+      raise ArgumentError, "TLS context hooks must be an array of Procs"
+    end
+
+    @tls_context_hooks=hooks
```
p-mongo, line 102: Can you add spaces around `=` please?

p-mongo (commented):
This is great. Could you also add two notes to the tutorial:

- The TLS context hooks are global. While it is possible to assign the entire array of hooks, doing so removes any hooks that were previously set. If an application intends to add a hook, the recommended pattern is to use `push` or `unshift` methods rather than assigning to `tls_context_hooks`.
- Because the TLS context hooks are global, libraries should expose methods that would modify the hooks (which would be called by the application at an appropriate time) rather than automatically modifying hooks when a library is loaded.
- To remove a hook, use a method like `Array#delete_if` (a reference to hook being removed must have been stored somewhere by the application).

docs/tutorials/ruby-driver-create-client.txt
p-mongo: I think the array assignment can easily become problematic and it should not be recommended. I suggest instead

```
  Mongo.tls_context_hooks.push(
    Proc.new { |context|
      context.ciphers = ["AES256-SHA"]
    }
  )
```

p-mongo (approved):

PR #2126, opened 2020-11-14
Fix RUBY-2432 Cannot list collections on 2.6 server if multiple batches are employed


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2432

ksadoff (approved):

PR #2152, opened 2020-12-04
RUBY-2463 Enforce timeout on socket closes


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2463

lib/mongo/socket.rb
```diff
@@ -155,7 +155,14 @@ def alive?
     #
     # @since 2.0.0
     def close
-      @socket.close rescue nil
+      begin
+        # Sometimes it seems the close call can hang for a long time
+        ::Timeout.timeout(5) do
```
durran, line 160: Nitpick: I would generally create a constant for `5` here so its purpose is labeled instead of being a magic number. But I'm also not fussed if it stays like this.

durran (approved):

PR #2153, opened 2020-12-04
RUBY-2464 Remove bundler setup from rakefile


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2464

PR #2161, opened 2021-01-08
Fix RUBY-2453 Segfault at program exit due to SRV polling


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2453

durran (approved):

PR #2166, opened 2021-01-28
RUBY-1682: Add Zstandard (zstd) network compression support

alexbevi:
Add support for zstd network compression (per DRIVERS-600) to the Ruby driver.

Note that this PR depends on https://github.com/mongodb-labs/mongo-ruby-spec-shared/pull/6 being merged for the specs to run.

### DIFF
--- a/docs/tutorials/ruby-driver-create-client.txt
+++ b/docs/tutorials/ruby-driver-create-client.txt
@@ -319,9 +319,8 @@ Ruby Options
      - none
 
    * - ``:compressors``
-     - A list of potential compressors to use, in order of preference. The driver chooses the first
-       compressor that is also supported by the server. Currently the driver only supports 'zstd',
-       'snappy' and 'zlib'.
+     - A list of potential compressors to use, in order of preference.
+       Please see below for details on how the driver implements compression.
      - ``Array<String>``
      - none
 
@@ -740,23 +739,9 @@ URI options are explained in detail in the :manual:`Connection URI reference
    * - compressors=Strings
      - ``:compressors => Array<String>``
 
-       Specified as an ordered, comma-separated list. The Ruby driver currently supports zstd, snappy
-       and zlib compression algorithms; the first algorithm in the list that the server
-       also supports will be used. Compression must be explicitly enabled by
-       specifying the desired compression algorithm; use "zstd,snappy,zlib" for maximum server
-       compatibility.
-
-       Snappy compression requires separate installation of the
-       snappy Ruby library; if the snappy Ruby library is not available,
-       the driver raises an error during ``Mongo::Client`` creation.
-       When using MongoDB 4.0 and earlier, zlib compression is supported by the
-       server but must be manually enabled; if zlib is the only compression algorithm
-       requested (or available due to missing snappy Ruby library), 4.0 and
-       earlier servers must be configured to explicitly enable zlib compression.
-
-       Zstd compression was introduced with Server 4.2 and requires a separate installation of the
-       zstd-ruby Ruby library. If the zstd-ruby Ruby library is not available,
-       the driver raises an error during ``Mongo::Client`` creation.
+       A comma-separated list of potential compressors to use, in order of
+       preference. Please see below for details on how the driver implements
+       compression.
 
    * - connect=String
      - ``:connect => Symbol``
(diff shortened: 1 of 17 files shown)


docs/tutorials/ruby-driver-create-client.txt
```diff
@@ -740,18 +740,24 @@ URI options are explained in detail in the :manual:`Connection URI reference
    * - compressors=Strings
      - ``:compressors => Array<String>``
 
-       Specified as a comma-separated list. The Ruby driver currently supports snappy
+       Specified as a comma-separated list. The Ruby driver currently supports snappy, zstd
```
p-mongo, line 743: I suggest placing zstd ahead of snappy since my understanding is it is a better choice (better compression at the same speed).

p-mongo: It seems this section is missing the explicit note that the driver tries to use the compression algorithms  in the order they are specified. I suggest replacing the first sentence with

> Specified as an ordered, comma-separated list.

lib/mongo/client.rb
```diff
@@ -116,7 +116,8 @@ class Client
     # @since 2.5.0
     VALID_COMPRESSORS = [
       Mongo::Protocol::Compressed::SNAPPY,
-      Mongo::Protocol::Compressed::ZLIB
+      Mongo::Protocol::Compressed::ZLIB,
+      Mongo::Protocol::Compressed::ZSTD
```
p-mongo, line 120: Let's have zstd ahead of snappy here also and please add the trailing comma to the last entry.

lib/mongo/client.rb
```diff
@@ -226,7 +227,7 @@ def hash
     # @option options [ Array<String> ] :compressors A list of potential
     #   compressors to use, in order of preference. The driver chooses the
     #   first compressor that is also supported by the server. Currently the
-    #   driver only supports 'snappy' and 'zlib'.
+    #   driver only supports 'snappy', 'zstd' and 'zlib'.
```
p-mongo, line 230: And reorder here.

lib/mongo/protocol/compressed.rb
```diff
@@ -149,6 +157,10 @@ def compress(buffer)
           Zlib::Deflate.deflate(buffer.to_s, @zlib_compression_level).force_encoding(BSON::BINARY)
         elsif @compressor_id == SNAPPY_BYTE
           Snappy.deflate(buffer.to_s).force_encoding(BSON::BINARY)
+        elsif @compressor_id == ZSTD_BYTE
+          # Use compression level of 3 by default as this is what the Java driver does
+          # DRIVERS-600 will allow this to be configurable in the future
```
p-mongo, line 162: If there is a reason (e.g. published benchmarks) for why 3 should be used, let's link to that. Otherwise let's remove the level and use the default provided by the zstd library.

p-mongo: Related: https://github.com/SpringMT/zstd-ruby/issues/28

alexbevi: Regarding the default compression level have a look at https://github.com/facebook/zstd/issues/1148, which resulted in https://github.com/facebook/zstd/pull/1174 and https://github.com/facebook/zstd/pull/1174/commits/e34c000e44444b9f8bd62e5af0a355ee186eb21f

```
/*-*************************************
*  Tuning parameters
***************************************/
#ifndef ZSTD_CLEVEL_DEFAULT
#  define ZSTD_CLEVEL_DEFAULT 3
#endif
```

As level 3 is the default in `zstd.h` as well it seems like a sane default (though not the default chosen by `zstd-ruby`)

alexbevi: @p if you feel strongly one way or another let me know, otherwise I would suggest leaving the level at 3

alexbevi: I've updated the comment in the source regarding why level 3 was chosen

p-mongo: If the default level is 3 then I suggest not specifying the level and making use of the default.

p-mongo:
I revised the documentation to have a separate section describing the compression behavior that is referenced from both the Ruby option and the URI option for compressors.

alexbevi:
Message compression now uses the [`zstd-ruby` default of 0](https://github.com/SpringMT/zstd-ruby/blob/master/ext/zstdruby/zstdruby.c#L22)

p-mongo:
Looking at the test results, zstd on java is failing. Looks like the C extension is trying to get built which probably won't work. If there is a java implementation of zstd we could look into integrating that in, otherwise removing the jruby configuration would be OK until someone asks for it to be supported.

alexbevi:
> Looking at the test results, zstd on java is failing. Looks like the C extension is trying to get built which probably won't work. If there is a java implementation of zstd we could look into integrating that in, otherwise removing the jruby configuration would be OK until someone asks for it to be supported.

I've commented out the evergreen matrixes for zstd-jruby for the time being. I can look into integrating the zstandard compressor from https://github.com/airlift/aircompressor (Apache 2 licensed) in a future PR.

p-mongo:
Looks like there is another zstd failure in Evergreen, can you investigate it?

p-mongo:
I retested this PR with the recent fixes on master, I believe it's good to be merged.

p-mongo (approved):

PR #2183, opened 2021-03-04
RUBY-2538 Update index commit quorum test for changed server error message


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2538

PR #2185, opened 2021-03-05
RUBY-2429 Implement base Versioned MongoDB API


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2429

lib/mongo/protocol/msg.rb
```diff
@@ -275,6 +275,28 @@ def bulk_write?
...
+        if client.nil?
+          raise ArgumentError, 'Client must be provided'
+        end
+
+        if server_api = client.options[:server_api]
+          main_document = @main_document.dup
+          if version = server_api[:version]
+            main_document['apiVersion'] = version
+          end
+          if server_api[:strict]
```
alcaeus, line 288: Seeing this check made me realise that the spec may need some extra tests to cover this bit of the spec:
> Drivers MUST add the `apiStrict` and `apiDeprecationErrors` options if they were specified by the user, even when the specified value is equal to the server default.

If the `strict` option was not specified by the user it should be omitted, but if it was specified than it should indeed be sent to the server, even when it's `false`.

alcaeus (changes requested):

lib/mongo/auth/conversation_base.rb
```diff
@@ -47,6 +47,38 @@ def initialize(user, connection, **opts)
       def speculative_auth_document
         nil
       end
+
+      # @return [ Protocol::Message ] The message to send.
+      def build_message(connection, auth_source, selector)
```
durran, line 52: Nice refactoring introducing this.

lib/mongo/operation/shared/executable.rb
```diff
@@ -59,14 +59,25 @@ def get_result(connection, client, options = {})
...
       end
 
-      def build_message(connection)
-        message(connection)
+      def build_message(connection, client)
+        msg = message(connection)
+        # We do not supply the client to the operations layer from end_sessions
+        # method in session pool.
+        # TODO fix this
+        if client &&
```
durran, line 73: Nit: Not necessary but I generally like to extract complex checks like this into their own methods.

lib/mongo/protocol/msg.rb
p-mongo: Changed.

lib/mongo/auth/conversation_base.rb
p-mongo: This was actually part of https://github.com/mongodb/mongo-ruby-driver/pull/2187, not this PR.

lib/mongo/protocol/msg.rb
```diff
@@ -275,6 +275,28 @@ def bulk_write?
...
+          raise ArgumentError, 'Client must be provided'
+        end
+
+        if server_api = client.options[:server_api]
+          main_document = @main_document.dup
+          if version = server_api[:version]
+            main_document['apiVersion'] = version
+          end
+          unless server_api[:strict].nil?
+            main_document['apiStrict'] = server_api[:strict]
```
durran, line 289: Are we ok with the user being able to set this to any truthy value?

durran: @p-mongo Was for you. The spec expects a boolean value but the user could potentially provide any truthy value here. There's nothing in the spec that I saw about validation here in dynamic languages so I was just asking if we were ok with that potentially erroring if a user supplied, say a string. I'm not sure what the server would do in that case.

alcaeus: The server will reject any non-boolean value:
```
MongoDB Enterprise > db.adminCommand({isMaster: 1, apiVersion: "1", apiStrict: "asdf"})
{
	"ok" : 0,
	"errmsg" : "BSON field 'APIParametersFromClient.apiStrict' is the wrong type 'string', expected type 'bool'",
	"code" : 14,
	"codeName" : "TypeMismatch"
}
```

While the spec specifies that the `strict` and `deprecationErrors` options are booleans, it does not specify whether this should trigger a client-side or server-side error. I'm leaning toward client-side errors to better distinguish the wrong configuration from other command errors. Let me know if I should clarify this in the spec.

p-mongo: The documentation specifies true and false as the only allowed values. Supplying other values would be contrary to documented usage, hence a user error (unless the server happens to accept other values in the future for some reason in which case it'll be a-ok). If the spec mandates client-side rejection of other values I will be happy to implement that.

alcaeus: I'm fine leaving this as-is for the time being.

lib/mongo/protocol/msg.rb
```diff
@@ -275,6 +275,28 @@ def bulk_write?
...
+          end
+          if server_api[:deprecation_errors]
```
alcaeus, line 291: This is missing the change applied to `strict`.

alcaeus (approved):
Lgtm. I'll defer to @durran for a final approval.

lib/mongo/protocol/msg.rb
durran: Looks good to me then as well.

durran (approved):
LGTM

PR #2190, opened 2021-03-06
RUBY-2533 Add test to ensure drivers expect operations to succeed when neither expectError nor expectResult are specified


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2533

alcaeus (approved):

PR #2196, opened 2021-03-21
RUBY-2548 Verify that changing server api in #with works


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2548

alcaeus (approved):

PR #2200, opened 2021-03-26
Fix RUBY-2558 :secondary_preferred read preference in 3.6+ sharded clusters performs primary read


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2558

lib/mongo/operation/shared/read_preference_supported.rb
```diff
@@ -36,47 +36,44 @@ module ReadPreferenceSupported
...
+        elsif connection.server.cluster.single?
+          # In Single topology the driver forces primaryPreferred read
+          # preference mode (via the slave_ok flag, in case of old servers)
+          # so that the query is satisfied.
+          true
+        else
+          # In replica sets and sharded clusters, read preference is passed
+          # to the server if one is specified by the application, and there
+          # is no default.
+          read && read.slave_ok? || false
```
alcaeus, line 70: Question without knowing internal semantics, why does this require `|| false` to be included? Asking because `A || false` will always evaluate to `A` in boolean algebra, so I'm wondering about what Ruby does when `read` evaluates to a false-ish value.

lib/mongo/operation/shared/read_preference_supported.rb
```diff
@@ -95,14 +92,18 @@ def command(connection)
...
+      def add_read_preference_legacy(sel, connection)
         if read && connection.description.mongos? && read_pref = read.to_mongos
-          Mongo::Lint.validate_camel_case_read_preference(read_pref)
-          sel = sel[:$query] ? sel : {:$query => sel}
-          sel = sel.merge(:$readPreference => read_pref)
-        else
-          sel
+          # If the read preference contains only mode and mode is secondary
+          # preferred and we are sending to a pre-OP_MSG server, this read
+          # preference is indicated by the :slave_ok wire protocol flag.
```
alcaeus, line 99: Took me a moment to realise that the comment explains when the `$readPreference` option doesn't need to be added. Maybe add another sentence like "In all other cases, add the `$readPreference` option" might help clarify this.

alcaeus (approved):

lib/mongo/operation/shared/read_preference_supported.rb
p-mongo: The `|| false` converts nil to false (when `read` is nil, in this case).

lib/mongo/operation/shared/read_preference_supported.rb
p-mongo: Sure, I added a note to this effect.

PR #2203, opened 2021-04-07
RUBY-2561 Remove skipReason from estimatedDocumentCount test in versioned API


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2561

PR #2204, opened 2021-04-07
RUBY-2426: Check if indexes for GridFS exist prior to creating them

alexbevi:
https://jira.mongodb.org/browse/RUBY-2426


alexbevi:
At least one spec failure with this PR that needs cleaning up:
```
MONGODB_URI="mongodb://dev:dev@localhost:27017/" rake spec SPEC=spec/mongo/grid/fs_bucket_spec.rb
                                                                                                                                                                                    
  1) Mongo::Grid::FSBucket#insert_one when the index creation encounters an error raises the error to the user
     Failure/Error:
       expect {
         fs.insert_one(file)
       }.to raise_error(Mongo::Error::OperationFailure)
     
       expected Mongo::Error::OperationFailure, got #<RSpec::Mocks::MockExpectationError: (#<Mongo::Collection:0x8280 namespace=ruby-driver.fs.files>).indexes(no args)
           expected: 0 times with any arguments
           received: 1 time> with backtrace:
         # ./lib/mongo/grid/fs_bucket.rb:491:in `ensure_indexes!'
         # ./lib/mongo/grid/fs_bucket.rb:167:in `insert_one'
         # ./spec/mongo/grid/fs_bucket_spec.rb:397:in `block (5 levels) in <top (required)>'
         # ./spec/mongo/grid/fs_bucket_spec.rb:396:in `block (4 levels) in <top (required)>'
         # /home/alex/.rvm/gems/ruby-2.7.1/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:124:in `block in run'
         # /home/alex/.rvm/gems/ruby-2.7.1/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:110:in `loop'
         # /home/alex/.rvm/gems/ruby-2.7.1/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:110:in `run'
         # /home/alex/.rvm/gems/ruby-2.7.1/gems/rspec-retry-0.6.2/lib/rspec_ext/rspec_ext.rb:12:in `run_with_retry'
         # /home/alex/.rvm/gems/ruby-2.7.1/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:37:in `block (2 levels) in setup'
         # ./spec/support/background_thread_registry.rb:62:in `block (2 levels) in <top (required)>'
     # ./spec/mongo/grid/fs_bucket_spec.rb:396:in `block (4 levels) in <top (required)>'
     # /home/alex/.rvm/gems/ruby-2.7.1/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:124:in `block in run'
     # /home/alex/.rvm/gems/ruby-2.7.1/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:110:in `loop'
     # /home/alex/.rvm/gems/ruby-2.7.1/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:110:in `run'
     # /home/alex/.rvm/gems/ruby-2.7.1/gems/rspec-retry-0.6.2/lib/rspec_ext/rspec_ext.rb:12:in `run_with_retry'
     # /home/alex/.rvm/gems/ruby-2.7.1/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:37:in `block (2 levels) in setup'
     # ./spec/support/background_thread_registry.rb:62:in `block (2 levels) in <top (required)>'
```

lib/mongo/grid/fs_bucket.rb
```diff
@@ -487,8 +487,28 @@ def files_name
...
+          unless index_exists?(chunks_name, FSBucket::CHUNKS_INDEX)
+            chunks_collection.indexes.create_one(FSBucket::CHUNKS_INDEX, :unique => true)
+          end
+        end
+      end
+
+      # Helper method to test for index existence by first verifying the collection exists.
+      # This allows calling Mongo::Collection#indexes without raising
+      #   Mongo::Error::OperationFailure Exception: [26:NamespaceNotFound]
+      # if the collection doesn't exist
```
alcaeus, line 505: I'll note that the "Enumerating indexes" spec [contains a suggested implementation](https://github.com/mongodb/specifications/blob/master/source/enumerate-indexes.rst#enumeration-getting-index-information) on how to enumerate indexes. That implementation checks for the `NamespaceNotFound` error and returns an empty array in that case.

If the driver should continue raising an exception when listing indexes on non-existing collections, this method should catch the exception itself and return `false` in that case. This also removes the need for running two commands for an index check.

alcaeus (commented):
I'll let @p-mongo decide on how to best handle the `NamespaceNotFound` error (either in the `listIndexes` API or in the helper), but the logic can be optimised either way.

lib/mongo/grid/fs_bucket.rb
p-mongo: +1

p-mongo:
I believe there needs to be a test added that the driver is able to write to gridfs when the user lacks permission to create indexes. This would, I expect, involve defining another test user.

The spec language is somewhat confusing in this regard:

> Drivers MUST check whether the indexes already exist before attempting to create them. This supports the scenario where an application is running with read-only authorizations.

Surely, an application "running with read-only authorizations" won't be able to write files to gridfs either. Does the spec indeed mean an environment where inserts are allowed but index creation is not?

p-mongo:
The documentation should also be updated. https://docs.mongodb.com/ruby-driver/master/tutorials/ruby-driver-gridfs/ currently says:

> Note that the first time a file is inserted, it will create the required index for you on the chunks collection.

One might read this to mean "the first time a file is inserted *using the Ruby driver*" which isn't accurate; additionally, the files index isn't mentioned and the lack of permissions isn't addressed either. I suggest something like the following wording instead:

> When the driver is inserting the first file into a bucket, it will attempt to create the required indexes on `files` and `chunks` collections. The required indexes are as follows: (both index specifications) 
> If the indexes cannot be created, such as due to the current user lacking the permissions to do so, the file insert will be aborted.
> If the bucket already has files, the driver will not attempt to create indexes, even if they are missing and the current user has permissions to create them. In this case a database administrator should create the needed indexes as soon as possible to ensure data integrity.

lib/mongo/grid/fs_bucket.rb
alexbevi: Rewrote the logic here

lib/mongo/grid/fs_bucket.rb
```diff
@@ -487,8 +487,26 @@ def files_name
...
+          unless database[collection_name].indexes.get(index_spec).nil?
+            database[collection_name].indexes.create_one(index_spec, options)
+          end
+        rescue Mongo::Error::OperationFailure => e
+          # proceed with index creation if a NamespaceNotFound error is thrown
+          if e.code == 26
+            database[collection_name].indexes.create_one(index_spec, options)
+          else
+            raise
+          end
```
alexbevi, line 509: @alcaeus I think this implementation is more aligned with what you had in mind

alexbevi: > I believe there needs to be a test added that the driver is able to write to gridfs when the user lacks permission to create indexes. This would, I expect, involve defining another test user.

@p do you think this test should be added?

p-mongo:
Yes I think so.

lib/mongo/grid/fs_bucket.rb
```diff
@@ -487,8 +487,26 @@ def files_name
...
+
+        if chunks_collection.find({}, limit: 1, projection: { _id: 1 }).first.nil?
+          create_index_if_missing!(chunks_name, FSBucket::CHUNKS_INDEX, :unique => true)
+        end
+      end
+
+      def create_index_if_missing!(collection_name, index_spec, options = {})
+        begin
+          unless database[collection_name].indexes.get(index_spec).nil?
+            database[collection_name].indexes.create_one(index_spec, options)
```
p-mongo, line 501: This appears to create the index if it exists?

lib/mongo/grid/fs_bucket.rb
```diff
@@ -487,8 +487,26 @@ def files_name
 
       def ensure_indexes!
         if files_collection.find({}, limit: 1, projection: { _id: 1 }).first.nil?
-          chunks_collection.indexes.create_one(FSBucket::CHUNKS_INDEX, :unique => true)
-          files_collection.indexes.create_one(FSBucket::FILES_INDEX)
+          create_index_if_missing!(files_name, FSBucket::FILES_INDEX)
+        end
+
+        if chunks_collection.find({}, limit: 1, projection: { _id: 1 }).first.nil?
+          create_index_if_missing!(chunks_name, FSBucket::CHUNKS_INDEX, :unique => true)
```
p-mongo, line 494: This seems to create each index based on the presence of data in the respective backing collection. I had a different understanding of the mechanism so the docs should be updated to match the actual behavior.

lib/mongo/grid/fs_bucket.rb
alexbevi: 🤦🏻

lib/mongo/grid/fs_bucket.rb
alexbevi: I'm going based on the [spec's "Before write operations"](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#id19) guidance:

```
Immediately before the first write operation on an instance of a GridFSBucket class is attempted (and not earlier), drivers MUST:

determine if the files collection is empty using the primary read preference mode.
and if so, create the indexes described above if they do not already exist
To determine whether the files collection is empty drivers SHOULD execute the equivalent of the following shell command:

> db.fs.files.findOne({}, { _id : 1 })
If no document is returned the files collection is empty.
```

alexbevi:
@p / @comandeo finally got around to this. hopefully this is good now, though might need rebasing :/

lib/mongo/grid/fs_bucket.rb
alexbevi: I haven't addressed this yet, so may still have to revise the docs before this can be merged

lib/mongo/grid/fs_bucket.rb
alexbevi: Fixed my backwards logic (finally)

p-mongo (approved):

comandeo (approved):

PR #2207, opened 2021-04-14
RUBY-2371: Use monotonic clock for measuring time

comandeo:
This PR introduces usage of monotonic clock instead of wall clock for measuring time. This ensures correct calculation of elapsed times.

See https://blog.dnsimple.com/2018/03/elapsed-time-with-ruby-the-right-way/ for details

### DIFF
--- a/lib/mongo/cluster.rb
+++ b/lib/mongo/cluster.rb
@@ -213,8 +213,7 @@ module Mongo
       end
 
       # Need to record start time prior to starting monitoring
-      start_time = Time.now
-      start_time_monotonic = Utils.monotonic_time
+      start_monotime = Utils.monotonic_time
 
       servers.each do |server|
         server.start_monitoring
@@ -230,7 +229,7 @@ module Mongo
         if server_selection_timeout < 3
           server_selection_timeout = 3
         end
-        deadline = start_time_monotonic + server_selection_timeout
+        deadline = start_monotime + server_selection_timeout
         # Wait for the first scan of each server to complete, for
         # backwards compatibility.
         # If any servers are discovered during this SDAM round we are going to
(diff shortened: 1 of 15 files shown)


lib/mongo/cluster.rb
```diff
@@ -214,6 +214,7 @@ def initialize(seeds, monitoring, options = Options::Redacted.new)
 
       # Need to record start time prior to starting monitoring
       start_time = Time.now
```
p-mongo, line 216: Is this start_time used for something now?

lib/mongo/utils.rb
```diff
@@ -87,5 +87,20 @@ def initialize(**opts)
...
+    # This function should be used if you need to measure time.
+    # @example Calculate elapsed time.
+    #   starting = Utils.monotonic_time
+    #   # do something time consuming
+    #   ending = Utils.monotonic_time
+    #   puts "It took #{(ending - starting).to_i} seconds"
+    #
+    # @see https://blog.dnsimple.com/2018/03/elapsed-time-with-ruby-the-right-way/
+    #
+    # @return [Float] seconds according to monotonic clock
```
p-mongo, line 100: :+1:

lib/mongo/cluster.rb
comandeo: It is used on line 244: `server.last_scan >= start_time`

The `last_scan` attribute of `Server` is public and explicitly documented with type `Time`. So, I am hesitant to change it.

p-mongo: I think there need to be two last_scan fields, the new one being monotonic_last_scan or last_scan_monotonic (or, given the naming conventions I prefer, I would suggest last_scan_monotime as a contraction from last_scan_monotonic_time).

Last scan is used in two places: 

- The server selector uses it for staleness calculations. The value there is compared to the server-reported time and must therefore remain to be a wallclock time.
-Cluster uses it for waiting in the Client constructor until all servers have been scanned. The time there should be monotonic.

The monotonic last scan can be `@api private`.

p-mongo (approved):

PR #2208, opened 2021-04-15
RUBY-2296: Fix app metadata spec

comandeo:
This commit fixes an assumption that the spec is always executed on a
Linux machine.

### DIFF
--- a/spec/mongo/server/app_metadata_shared.rb
+++ b/spec/mongo/server/app_metadata_shared.rb
@@ -11,14 +11,40 @@ shared_examples 'app metadata document' do
     document[:client][:driver][:version].should == Mongo::VERSION
   end
 
-  it 'includes operating system information' do
-    %w(linux darwin).each do |os|
-      if !!(RUBY_PLATFORM =~ /\#{os}/)
-        document[:client][:os][:type].should == os
+  context 'linux' do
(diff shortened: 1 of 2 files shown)


p-mongo (commented):
Unfortunately this test now lost all of the explicit assertions on what data is actually ending up in the metadata document. For example, if cpu/platform are empty it would still pass.

Can the original test be left as is and skipped if OS isn't linux and a new test be added which asserts specific values for mac instead?

comandeo:
We can do that; however, in this case we will have more or less same problems with `architecture` attribute.  Ideally we need to create a test for every supported combination of OS/CPU type. I am not sure if it worth it, though.

My assumption was that we if `RbConfig::CONFIG` is not complete, there is not much we can do anyway, so we can rely on it. On the other hand, I do understand your reasons.

p-mongo:
For now I think two cases would be sufficient: the existing one which works in Evergreen and hopefully on all of the (at least mainstream) linux distributions, and one which works on your machine which would hopefully work on all apple products.

p-mongo:
I would also like to see the values that the driver will use for the metadata fields on apples, right now this isn't in the test.

p-mongo (approved):

PR #2220, opened 2021-04-30
RUBY-2566 Prohibit skip and limit in estimated_document_count

comandeo:


lib/mongo/collection/view/readable.rb
```diff
@@ -236,6 +236,12 @@ def estimated_document_count(opts = {})
             raise ArgumentError, "Cannot call estimated_document_count when querying with a filter"
           end
 
+          %w[limit skip].each do |opt|
```
p-mongo, line 239: `%w` produces a list of strings, this needs to be `%i` to produce a list of symbols since the options are documented as being symbols.

lib/mongo/collection/view/readable.rb
```diff
@@ -236,6 +236,12 @@ def estimated_document_count(opts = {})
...
+            unless @options[opt].nil?
```
p-mongo, line 240: Does `if @options.key?(opt}` work in this case? If yes  I think this would be a clearer statement as to the requirements.

PR #2227, opened 2021-05-12
RUBY-2611 Remove blacklist and whitelist terms

comandeo:


lib/mongo/protocol/msg.rb
```diff
@@ -199,7 +199,7 @@ def documents
       # auto-encryption options, the client has not been instructed to bypass
       # auto-encryption, and mongocryptd determines that this message is
       # eligible for encryption. A message is eligible for encryption if it
-      # represents one of the command types whitelisted by libmongocrypt and it
+      # represents one of the command types allowlisted by libmongocrypt and it
```
p-mongo, line 202: I think these need to be hyphenated, i.e. `allow-listed`. Looking through the rest of the diff, people use the phrase "allow list" as two words, plus "allowlist" doesn't flow very well.

spec/support/cluster_tools.rb
```diff
@@ -121,7 +121,7 @@ def step_up(address)
       rescue Mongo::Error::OperationFailure => e
         # Election failed. (125)
         if e.code == 125
-          # Possible reason is the node we are trying to elect has blacklisted
+          # Possible reason is the node we are trying to elect has denylisted
```
p-mongo, line 124: I suppose `deny-listed` here would be good enough. "Blacklisted" here means a general prohibition rather than referring to a particular blacklist.

p-mongo (approved):

PR #2232, opened 2021-05-14
Remove NotMasterNoSlaveOk

comandeo:


p-mongo (approved):
The changes in spec files were also made upstream in the specs repo, right?

comandeo:
> The changes in spec files were also made upstream in the specs repo, right?

These changes are already in specs repo. I decide not to copy complete new specs since there are other changes.

PR #2236, opened 2021-05-17
RUBY-2618 Rename slaveOk to secondaryOk

comandeo:


lib/mongo/server_selector/nearest.rb
```diff
@@ -41,13 +41,13 @@ def name
...
 
-      # Whether the slaveOk bit should be set on wire protocol messages.
+      # Whether the secondaryOk bit should be set on wire protocol messages.
       #   I.e. whether the operation can be performed on a secondary server.
       #
       # @return [ true ] true
       #
       # @since 2.0.0
-      def slave_ok?
+      def secondary_ok?
```
p-mongo, line 50: This looks like a public method, in which case I believe the `slave_ok?` method should be added back as an alias of `secondary_ok?` with a @deprecated note.

p-mongo: And a ticket to remove these under 3.0 epic.

p-mongo: And then the tests should test both methods.

comandeo: Good point, I'll check other public methods as well.

comandeo: Added aliases and tests.

lib/mongo/server_selector/nearest.rb
```diff
@@ -41,16 +41,25 @@ def name
...
 
-      # Whether the slaveOk bit should be set on wire protocol messages.
+      # Whether the secondaryOk bit should be set on wire protocol messages.
       #   I.e. whether the operation can be performed on a secondary server.
       #
       # @return [ true ] true
       #
-      # @since 2.0.0
-      def slave_ok?
+      # @since 2.15.0
```
p-mongo, line 49: We stopped using @since tags because they fail to account for signature expansions/added options, so whenever a method is modified the @since tag may be dropped if the signature changes and for new methods we don't need them.

spec/support/shared/server_selector.rb
```diff
@@ -231,3 +231,12 @@
     end
   end
 end
+
+shared_examples 'a server selector with legacy secondary_ok' do
+  describe '#slave_ok?' do
+    it 'behaves like #secondary_ok?' do
+      selector = Mongo::ServerSelector::Nearest.new
+      expect(selector.slave_ok?).to eq(selector.secondary_ok?)
```
p-mongo, line 239: I am confused by this test. If this simply asserts that slave_ok is the same as secondary_ok then why is selector specified to be nearest? Shouldn't this apply to all selector types?

lib/mongo/protocol/query.rb
```diff
@@ -174,7 +174,7 @@ def determine_limit
       FLAGS = [
         :reserved,
         :tailable_cursor,
-        :slave_ok,
+        :secondary_ok,
```
p-mongo, line 177: We generally consider the operation and protocol layer private. If so the flag list can be mutated at will but it also should be marked @api private given that the class itself is "semiprivate".

Have you considered whether users may legitimately use this flag list for anything?

spec/support/shared/server_selector.rb
```diff
@@ -231,3 +231,12 @@
     end
   end
 end
+
+shared_examples 'a server selector with legacy secondary_ok' do
```
p-mongo, line 235: Can this be "with slave_ok" for clarity please.

lib/mongo/protocol/query.rb
```diff
@@ -174,7 +174,7 @@ def determine_limit
       FLAGS = [
```
p-mongo, line 174: Let's add @api private to FLAGS and leave this as is.

p-mongo:
Upon review of usage of ServerSelector I believe the pattern that applications should use for determining whether a server selector would allow requests to be fulfilled by secondaries is to compare the selector to Primary, e.g.

```
server_selector.is_a?(Mongo::ServerSelector::Primary)
```

... instead of consulting 

```
server_selector.slave_ok?
server_selector.secondary_ok?
```

Therefore I suggest removing the slave_ok? aliases and renaming slave_ok? to secondary_ok? as was done previously, and marking all secondary_ok? methods @api private.

spec/support/shared/server_selector.rb
p-mongo: This will now be removed altogether.

lib/mongo/server_selector/nearest.rb
comandeo: Understood, thanks.

lib/mongo/protocol/query.rb
comandeo: Marked with `@api private`

PR #2241, opened 2021-05-19
RUBY-2571 Use hello with versioned API

comandeo:


lib/mongo/server/app_metadata.rb
```diff
@@ -51,6 +51,12 @@ class AppMetadata
       # @api private
       PURPOSES = %i(application monitor push_monitor).freeze
 
+      HELLO = {
+        hello: 1,
+        helloOk: true,
```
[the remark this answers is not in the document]
comandeo: We should update specs then, I guess. I was looking at this one - https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.rst#id7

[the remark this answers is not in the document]
alcaeus: Oh, thanks for pointing that out!

lib/mongo/server/app_metadata.rb
```diff
@@ -109,25 +109,31 @@ def initialize(options = {})
       # @api private
       attr_reader :purpose
 
+      # @return [ Hash | nil ] Information server api.
```
p-mongo, line 112: What do you mean by "information server api"? I think this hash is referred to elsewhere as "server API parameters".

lib/mongo/server/app_metadata.rb
```diff
@@ -109,25 +109,31 @@ def initialize(options = {})
...
+      # @api private
+      attr_reader :server_api
+
       # @return [ Array<Hash> | nil ] Information about libraries wrapping
       #   the driver.
       attr_reader :wrapping_libraries
 
-      # Get the bytes of the ismaster message including this metadata.
+      # Get the metadata as BSON::Document to be sent to
+      # a server for handshake purposes. The document is incomplete, and should
```
p-mongo, line 127: s/for handshake purposes/as part of the handshake/.

I also wouldn't say that the document is incomplete because that is confusing - *all* of the metadata is included, yes? I think just leaving "this metadata document must be incorporated into a suitable handshake command" would be sufficient.

lib/mongo/server/connection_common.rb
```diff
@@ -52,6 +52,38 @@ def connected?
...
+      #
+      # @param [Server::AppMetadata] app_metadata Application metadata
+      #
+      # @return [BSON::Document] Document that should be sent to a server
+      #     for handshake purposes.
+      #
+      # @api private
+      def handshake_document(app_metadata)
+        document = if app_metadata.server_api && app_metadata.server_api[:version]
+                     HELLO_OP_QUERY
```
p-mongo, line 65: This is going to get replaced by hello op message later, right? If yes can you add a brief comment saying so please.

lib/mongo/server/connection_common.rb
```diff
@@ -52,6 +52,38 @@ def connected?
...
+                   else
+                     LEGACY_HELLO_OP_QUERY
```
p-mongo, line 67: Is this the ismaster payload to be used with op_query (and I gather that the payload differs between op_query and op_msg)? I am wondering if it's better to have a single payload and modify it for the protocol at the point where the protocol message is created.

lib/mongo/server/monitor.rb
```diff
@@ -263,10 +257,10 @@ def pre_stop
       def do_scan
         begin
           monitoring.publish_heartbeat(server) do
-            ismaster
+            check
```
p-mongo, line 260: :+1:  I think this reads well.

spec/mongo/client_construction_spec.rb
```diff
@@ -1682,14 +1682,16 @@
             client.should be_a(Mongo::Client)
           end
 
-          it 'discovers servers' do
-            client.summary.should_not =~ /UNKNOWN/
+          it 'does not discovers servers' do
```
p-mongo, line 1685: does not discover servers

spec/mongo/server/app_metadata_spec.rb
```diff
@@ -66,7 +66,9 @@
         end
 
         it 'truncates the document' do
-          expect(app_metadata.send(:ismaster_bytes)).to be_a(String)
+          expect(
+            app_metadata.validated_document.to_bson.to_s.size
+          ).to be < described_class::MAX_DOCUMENT_SIZE
```
p-mongo, line 71: Can this constant be replaced by the literal value (512?) OR the assertion made additionally that the constant is no greater than 512, to maintain the meaningfulness of this test if document size increases.

spec/mongo/server/app_metadata_spec.rb
```diff
@@ -110,21 +118,10 @@
...
-            sasl_supported_mechs_size += 'saslSupportedMechs'.length + 1        # length of BSON key
-            sasl_supported_mechs_size += 4                               # length of BSON string length
-            sasl_supported_mechs_size += sasl_supported_mechs.length + 1 # length of BSON string
-          end
-
-          expect(app_metadata.ismaster_bytes.length).to eq(Mongo::Server::Monitor::Connection::ISMASTER_BYTES.length + sasl_supported_mechs_size + 26)
+        it 'truncates the document' do
+          expect(
+            app_metadata.validated_document.to_bson.to_s.size
+          ).to be < described_class::MAX_DOCUMENT_SIZE
```
p-mongo, line 124: I think the content of the truncation tests got largely lost with these changes. Is the validated document here static? If server versions are restricted to, say, 3.6-4.2 or 4.4+or 4.9+, is the contents then static? If I think an assertion on the contents would be helpful (that the driver_doc is omitted).

lib/mongo/server/connection_common.rb
p-mongo: After reading the rest of the diff I think there is abstraction leakage going on where the details of wire protocol ($db field) is leaking into the content documents (i.e. the actual ismaster/hello payloads).

I would prefer to have $db and other fields needed for a particular message type to be added in the same place where the corresponding message class is instantiated (either with or without adding a helper to do so).

lib/mongo/server/connection_common.rb
```diff
@@ -52,6 +52,28 @@ def connected?
...
+      def handshake_document(app_metadata)
+        document = if app_metadata.server_api && app_metadata.server_api[:version]
+                     HELLO_DOC
+                   else
+                     LEGACY_HELLO_DOC
+                   end
+        document.merge(app_metadata.validated_document)
+      end
+
+      HELLO_DOC = BSON::Document.new({ hello: 1 }).freeze
```
p-mongo, line 72: These constants either need to be moved under the `private` declaration, marked `@api private` or documented with docstrings if they are public (which I don't think they should be ).

lib/mongo/server/app_metadata.rb
```diff
@@ -151,10 +157,6 @@ def full_client_document
         end
       end
 
-      def serialize
-        Protocol::Query.new(Database::ADMIN, Database::COMMAND, document, :limit => -1).serialize
-      end
-
       def document
```
p-mongo, line 160: Can this method please gain either a docstring or a comment providing a summary of what it returns?

spec/mongo/server/push_monitor/connection_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+require 'spec_helper'
+
+describe Mongo::Server::PushMonitor::Connection do
+  describe '#check_document' do
+    it 'returns hello document with API version' do
+      meta = Mongo::Server::AppMetadata.new({
+        server_api: { version: '1' }
+      })
+      subject = described_class.new(dup(), {app_metadata: meta})
```
p-mongo, line 12: What does `dup` do here? You don't need the empty parens for argument-less invocations in Ruby.

spec/mongo/server/push_monitor/connection_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+# encoding: utf-8
+
+require 'spec_helper'
+
+describe Mongo::Server::PushMonitor::Connection do
+  describe '#check_document' do
+    it 'returns hello document with API version' do
+      meta = Mongo::Server::AppMetadata.new({
```
p-mongo, line 9: `{}` may be omitted for the last (or only) argument of a method which is a hash.

spec/mongo/server/push_monitor/connection_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+        server_api: { version: '1' }
+      })
+      subject = described_class.new(dup(), {app_metadata: meta})
+      document = subject.check_document
+      expect(document['hello']).to eq(1)
+    end
+
+    it 'returns legacy hello document without API version' do
+      meta = Mongo::Server::AppMetadata.new({})
+      subject = described_class.new(dup(), {app_metadata: meta})
```
p-mongo, line 19: `{` and `}`  may be omitted here also.

lib/mongo/server/app_metadata.rb
comandeo: Corrected.

lib/mongo/server/app_metadata.rb
comandeo: Good point, corrected accordingly.

lib/mongo/server/connection_common.rb
comandeo: After a discussion, handshake and check documents building is now properly distributed between layers.

lib/mongo/server/connection_common.rb
comandeo: Comment added close to places where we build messages.

lib/mongo/server/connection_common.rb
comandeo: Very good point, moved under `private`.
I suppose there is no need to add `@api private` for constants/methods that are in `private` section. Please let me know if it is still needed.

lib/mongo/server/app_metadata.rb
comandeo: Documenting comment added.

spec/mongo/server/push_monitor/connection_spec.rb
comandeo: Good catch, thanks! `double` was meant, corrected.

p-mongo (approved):

lib/mongo/server/connection_common.rb
p-mongo: No, if a method has private visibility, it isn't part of the public api and @api annotation is not needed.

p-mongo: ... which holds for most things in driver. For Mongoid I suppose there could be private methods that are part of public API. BSON would be probably in between since it does muck with stdlib classes.

PR #2246, opened 2021-05-19
RUBY-2630 Implement load balanced topology type


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2630

lib/mongo/cluster/topology/load_balanced.rb
```diff
@@ -0,0 +1,126 @@
...
+
+      # Defines behavior for when a cluster is in load-balanced topology.
+      class LoadBalanced < Base
+
+        # The display name for the topology.
+        NAME = 'LoadBalanced'.freeze
+
+        # Get the display name.
+        #
+        # @example Get the display name.
```
comandeo, line 30: I would rather remove this example while usage of this method is rather clear. But if we want to keep it, let's change to `LoadBalanced.display_name`.

lib/mongo/server/description.rb
```diff
@@ -237,6 +243,10 @@ def initialize(address, config = {}, average_round_trip_time = nil)
       # @return [ Hash ] The actual result from the ismaster command.
       attr_reader :config
 
+      def load_balancer?
```
comandeo, line 246: Maybe a short docstring here? Since all other methods in this class have one.

comandeo (approved):

lib/mongo/cluster/topology/load_balanced.rb
p-mongo: I removed all of the examples.

lib/mongo/server/description.rb
p-mongo: Added, there should've been one here.

PR #2248, opened 2021-05-21
RUBY-2636 Assert auth source in both client options and as calculated by Auth::User


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2636

comandeo (approved):

PR #2255, opened 2021-05-28
RUBY-2268 RUBY-2490 Ruby 3.0 support

johnnyshields:
Summary:
- Add Ruby 3.0 to CI test definitions
- Upgrade Rspec to 3.10+ for Ruby 3.0 support (tests fail otherwise)
- Remove rspec-mocks-diag gem and switch back to standard rspec-mocks


johnnyshields:
@alcaeus can you give me a dump of the failing specs for this one?

alcaeus:
Failures for 4.4 build with ruby-3.0: https://gist.github.com/alcaeus/ad0431dfc373083e5b1c8b1160a2983e. Most other 3.0 builds fail as well, but we can check those out once this first build succeeds.

johnnyshields:
@alcaeus the failures do not look like Ruby 3.0-specific issues. Would be good if MongoDB team can own this PR as I don't think I'll be able to fix it.

johnnyshields:
@comandeo why are we using `rspec-mocks-diag` ? Can we get rid of it? In my testing, I got errors using RSpec < 3.10

comandeo:
> @comandeo why are we using `rspec-mocks-diag` ? Can we get rid of it? In my testing, I got errors using RSpec < 3.10

This gem provides better stack traces for exceptions happens outside of main thread. Since we have multiple threads in the driver, we should keep the gem, I suppose.

Actually, I can see that tests passes on CI (and on my computer) with current set of gems. So, I am wondering why it fails for you now. Could you double check?

johnnyshields:
@comandeo the failures happened on mongoid repo actually, not mongo-ruby-driver. So fine to keep it I guess. However, we should consider contributing the changes from rspec-mocks-diag back to rspec-mocks so that mongoDB doesn't need to maintain their own version of RSpec....

.evergreen/config.yml
```diff
@@ -994,9 +998,9 @@ buildvariants:
   - matrix_name: "lint"
     matrix_spec:
       lint: on
-      ruby: "ruby-2.6"
+      ruby: "ruby-2.7"
       mongodb-version: ["4.4"]
-      topology: ["standalone", "replica-set", "sharded-cluster"]
+      topology: '*'
       os: ubuntu1604
```
p-mongo, line 1004: Load balancer config is failing on ubuntu1604 because there is no haproxy there, let's try ubuntu1804 instead.

p-mongo (approved):
If the lint configs on ubuntu1804 are green feel free to merge.

PR #2258, opened 2021-06-01
RUBY-2612 Use hello for monitoring if supported

comandeo:


lib/mongo/server/monitor/connection.rb
```diff
@@ -240,6 +244,14 @@ def add_server_connection_id
           end
           raise e
         end
+
+        # Detect whether server supports hello command based on server reply
```
p-mongo, line 248: This method updates `@hello_ok` based on server support of hello. I would expect a method that only detects something to not mutate the receiver.

spec/mongo/server/monitor/connection_spec.rb
```diff
@@ -137,20 +137,35 @@
...
+
+      it 'returns legacy hello document' do
+        subject = described_class.new(double("address"), app_metadata: meta)
+        document = subject.check_document
+        expect(document['isMaster']).to eq(1)
+      end
+
+      it 'returns hello document when server responded with helloOk' do
+        subject = described_class.new(double("address"), app_metadata: meta)
+        subject.instance_variable_set(:@hello_ok, true)
```
p-mongo, line 165: This should use a mock probably. I suggest adding `hello_ok?` and installing a mock version of it here which would return true rather than overwriting state in the object.

p-mongo (approved):

spec/mongo/server/monitor/connection_spec.rb
comandeo: Good idea, updated code and tests accordingly.

lib/mongo/server/monitor/connection.rb
comandeo: Updated the docsctring.

p-mongo (approved):

PR #2259, opened 2021-06-03
Fix RUBY-2662 app_metadata_shared is being overwritten


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2662

comandeo (approved):

PR #2261, opened 2021-06-07
RUBY-2652 Send read preference to load balancers


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2652

PR #2270, opened 2021-06-14
RUBY-2689 Run specs against serverless

comandeo:


.evergreen/serverless/create-instance.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
```
p-mongo, line 1: Are these files copied from somewhere?

spec/support/spec_config.rb
```diff
@@ -410,17 +410,21 @@ def auth_options
...
         ssl_verify: true,
-        ssl_cert:  client_cert_path,
-        ssl_key:  client_key_path,
-        ssl_ca_cert: ca_cert_path,
-      }.merge(Utils.underscore_hash(@uri_tls_options))
-    else
-      {}
-    end
+    }.tap do |options|
+      if @mongodb_uri.servers.grep(/mongodb.*\.net/).empty?
```
p-mongo, line 418: The domain is `mongodb.net` right? So that should be matched verbatim?

spec/support/spec_config.rb
```diff
@@ -526,6 +530,9 @@ def root_user
 
   # Get the default test user for the suite on versions 2.6 and higher.
   def test_user
+    # When testing against a serverless instance, we are not allowed to create
+    # new users, we just have one user for everyhing.
+    return root_user if Utils.serverless?
```
p-mongo, line 535: Can you please add a blank line after this one to separate this piece from the rest of the method.

spec/support/utils.rb
```diff
@@ -595,4 +595,14 @@ def match_with_type?(expected, actual)
       end
     end
   end
+
+  # Detect whether specs are running against Mongodb Atlas serverless instance.
+  # This method does not do any magic, it just checks whether environment
+  # variable SERVERLES is set. This is a recommended way to inform spec runners
```
p-mongo, line 601: `SERVERLESS`

.evergreen/serverless/create-instance.sh
comandeo: Yes, from here - https://github.com/mongodb-labs/drivers-evergreen-tools/tree/master/.evergreen/serverless

spec/support/spec_config.rb
comandeo: Apparently, test serverless instances are in the domain `mongodb-dev.net`, hence the regex.

.evergreen/serverless/create-instance.sh
p-mongo: We have d-e-t as a submodule in https://github.com/mongodb/mongo-ruby-driver/tree/master/.mod. You can then symlink into it like https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/support/ocsp does.

spec/support/spec_config.rb
p-mongo: Can you add that as a comment please.

.evergreen/serverless/create-instance.sh
comandeo: Oh, missed that, thanks. Fixed.

spec/support/spec_config.rb
comandeo: Done.

spec/support/spec_config.rb
comandeo: Done

spec/support/utils.rb
comandeo: Done

comandeo:
https://jira.mongodb.org/browse/RUBY-2689

spec/runners/transactions.rb
```diff
@@ -40,7 +40,7 @@ def define_transactions_spec_tests(test_paths)
                 elsif !test.multiple_mongoses? && SpecConfig.instance.addresses.length > 1
                   # Many transaction spec tests that do not specifically deal with
                   # sharded transactions fail when run against a multi-mongos cluster
-                  skip "Test does not specify multiple mongoses"
+                  skip "Test does not specify multiple mongoses" unless Utils.serverless?
```
p-mongo, line 43: I am having trouble parsing this. Can the new condition be moved to the if-elsif level and a note added reflecting what the behavior should be under serverless and why?

p-mongo (approved):

spec/runners/transactions.rb
comandeo: Now I think that this additional `unless` is not necessary. We should treat serverless instances as multiple mongoses environment without a special case here.
So, the check is removed.

spec/spec_tests/retryable_writes_spec.rb
```diff
@@ -6,7 +6,12 @@
 require 'runners/crud'
 
 describe 'Retryable writes spec tests' do
-  require_wired_tiger
+  unless Utils.serverless?
+    # Storage detection fails for serverless instances. However, it is save to
```
p-mongo, line 10: safe

spec/runners/transactions.rb
p-mongo: In theory any tests that succeed against  2 mongoses should also succeed against 3 mongoses, and if they don't that is probably a bug.

p-mongo:
To troubleshoot the pinning failures I suggest either retrieving server logs from atlas (this is likely awkward if serverless logs are treated the same as atlas logs, possibly untenable) or enabling command logging for all operations on the affected tests and comparing the logged queries against spec requirements, i.e. that the queries that are timing out (likely this means they are hanging) are being sent to the correct mongoses.

spec/spec_tests/retryable_writes_spec.rb
comandeo: Fixed

PR #2275, opened 2021-06-18
RUBY-2220 Move find command construction code into operations layer


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2220

spec/mongo/operation/find/legacy_spec.rb
```diff
@@ -78,7 +78,7 @@
         double('description').tap do |description|
           expect(description).to receive(:mongos?) { false }
           expect(description).to receive(:standalone?) { false }
-          expect(description).to receive(:load_balancer?) { false }
+          #expect(description).to receive(:load_balancer?) { false }
```
comandeo, line 81: Maybe delete it or add a couple of words why it is commented out?

comandeo (approved):

spec/mongo/operation/find/legacy_spec.rb
p-mongo: I might need this later, I haven't actually investigated in detail why this diff makes this invocation not happen. I'll review this line before completing the epic.

PR #2279, opened 2021-06-23
RUBY-2663 Formalize behavior of Server#unknown! when the server is a load balancer


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2663

comandeo (approved):

PR #2285, opened 2021-06-25
RUBY-2690 Copy fixed spec test

comandeo:


comandeo:
Spec test is copied from https://github.com/mongodb/specifications/pull/1027
So, we should not merge until the one is specifications is merged.

p-mongo (approved):

PR #2286, opened 2021-06-28
RUBY-2619 Add test for security-sensitive commands

comandeo:


spec/runners/unified/assertions.rb
```diff
@@ -99,7 +99,7 @@ def assert_events
         client = entities.get(:client, client_id)
         subscriber = @subscribers.fetch(client)
         expected_events = spec.use!('events')
-        actual_events = subscriber.wanted_events
+        actual_events = subscriber.wanted_events(@observe_secnsitive)
```
p-mongo, line 102: `sensitive` is misspelled

spec/runners/unified/test.rb
```diff
@@ -156,6 +156,15 @@ def create_entities
                 end
               end
             end
+            if @observe_secnsitive = spec.use('observeSensitiveCommands')
+              %w(authenticate saslStart saslContinue getnonce createUser updateUser copydbgetnonce copydbsaslstart copydb hello ismaster isMaster).each do |event|
+                  subscriber = (@subscribers[client] ||= EventSubscriber.new)
+                  unless client.send(:monitoring).subscribers[Mongo::Monitoring::COMMAND].include?(subscriber)
+                    client.subscribe(Mongo::Monitoring::COMMAND, subscriber)
+                  end
+                  subscriber.add_wanted_events(event)
```
p-mongo, line 165: This method is used to indicate which events are wanted (e.g. CommandStarted), not which commands should be recorded. I am confused what this call is actually doing/is meant to do?

lib/mongo/monitoring/event/secure.rb
```diff
@@ -53,8 +54,14 @@ module Secure
...
-          if REDACTED_COMMANDS.include?(command_name.to_s) &&
-            !%w(1 true yes).include?(ENV['MONGO_RUBY_DRIVER_UNREDACT_EVENTS']&.downcase)
+          if %w(1 true yes).include?(ENV['MONGO_RUBY_DRIVER_UNREDACT_EVENTS']&.downcase)
+            return document
+          end
+
+          if REDACTED_COMMANDS.include?(command_name.to_s)
+            BSON::Document.new
+          elsif %w(hello ismaster isMaster).include?(command_name.to_s) &&
+            !!document['speculativeAuthenticate']
```
p-mongo, line 64: If the spec requires omitting the entire event payload rather than just the `speculativeAuthenticate` subkey, as I think it does, can that please be added as a note here for future readers?

spec/runners/unified/assertions.rb
comandeo: Fixed

spec/runners/unified/test.rb
comandeo: Thanks, good point. Removed.

lib/mongo/monitoring/event/secure.rb
comandeo: As I understood the spec, we must omit the entire payload. A comment added.

comandeo:
Waiting for https://github.com/mongodb/specifications/pull/1026

spec/runners/crud/requirement.rb
```diff
@@ -106,6 +107,9 @@ def satisfied?
             ok = ok && [:allow, :forbid].include?(serverless)
           end
         end
+        if @auth
+          ok &&= (@auth == cc.auth_enabled?)
```
p-mongo, line 111: This should simply be `ok &&= cc.auth_enabled?`, the extra equality condition may fail if auth_enabled? returns a truthy but not true value.

p-mongo (approved):

spec/runners/crud/requirement.rb
comandeo: 👍

PR #2293, opened 2021-07-05
RUBY-2529 Time-series collection

comandeo:


spec/runners/unified/ddl_operations.rb
```diff
@@ -17,7 +17,14 @@ def create_collection(op)
         if session = args.use('session')
           opts[:session] = entities.get(:session, session)
         end
-        database[args.use!('collection')].create(**opts)
+        collection_opts = {}
+        if timeseries = args.use('timeseries')
+          collection_opts['timeseries'] = timeseries
+        end
+        if expire_after_seconds = args.use('expireAfterSeconds')
+          collection_opts['expireAfterSeconds'] = expire_after_seconds
```
p-mongo, line 25: The `collection_opts` are the options documented in https://docs.mongodb.com/ruby-driver/master/api/Mongo/Collection.html#initialize-instance_method, which are Ruby options. If this method accepts server-named camel cased options (e.g. `writeConcern`), that should be documented. Otherwise the options here need to be converted to Ruby spelling, I suggest `:time_series` and `:expire_after`. Note that in Ruby options we give times in seconds where feasible and we do not use unit suffixes.

spec/runners/unified/ddl_operations.rb
```diff
@@ -32,7 +39,7 @@ def drop_collection(op)
     def assert_collection_exists(op, state = true)
       consume_test_runner(op)
       use_arguments(op) do |args|
-        client = ClientRegistry.instance.global_client('authorized')
+        client = ClientRegistry.instance.global_client('root_authorized')
```
comandeo, line 42: I am not sure if this is the right way to fix the issue, but without this change specs fail with
```
[13:Unauthorized]: not authorized on ts-tests to execute command { listCollections: 1, cursor: {}, nameOnly: true, $db: "ts-tests", lsid: { id: UUID("4e1557e6-2023-4eba-98e4-ce33bd6e99e4") } } (on localhost:27017, modern retry, attempt 1)
```

p-mongo: This is probably a-ok, the biggest issue should be that  root_authorized doesn't share cluster with authorized hence would create more overhead.

The databases/collections used in spec tests are enumerated in https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/support/spec_config.rb#L554 for access control purposes.

lib/mongo/collection.rb
```diff
@@ -87,6 +87,11 @@ def ==(other)
     #   option.
     # @option options [ Hash ] :write_concern The write concern options.
     #   Can be :w => Integer|String, :fsync => Boolean, :j => Boolean.
+    # @option options [ Hash ] :time_series Create a time-series collection.
+    #   See https://docs.mongodb.com/manual/core/timeseries-collections/ for more
+    #   information about time-series collection.
+    # @option options [ Integer ] :expire_after (optional) Number indicating
```
p-mongo, line 93: Aren't all options optional?

lib/mongo/collection.rb
```diff
@@ -239,6 +244,12 @@ def create(opts = {})
       options = Hash[self.options.reject do |key, value|
         %w(read read_preference read_concern).include?(key.to_s)
       end]
+      # Converting Ruby spelled time series options to server style.
+      [[:time_series,  :timeseries], [:expire_after, :expireAfterSeconds]].each do |(ruby_key, server_key)|
```
p-mongo, line 248: Can you please use a map for this. A map is still iterable via `map.each do |key, value|`

lib/mongo/collection.rb
```diff
@@ -239,6 +244,12 @@ def create(opts = {})
...
+        if options.has_key?(ruby_key)
+          options[server_key] = options.delete(ruby_key)
```
p-mongo, line 250: This is mutating the passed in argument which is not good. If you want to write to `options` you need to `dup` it first.

p-mongo:
Just to clarify, you are not allowing specifying the time series and expriration options in the actual create collection call (which takes the `opts` argument), and the only place where these options can be specified are on collection level where options generally apply to all operations?

This PR should also have a doc section added to https://docs.mongodb.com/ruby-driver/master/tutorials/ruby-driver-collection-tasks/, and if the creation options cannot be passed to the create method this also needs to be mentioned in the user documentation so that people don't try to do that.

spec/runners/unified/ddl_operations.rb
comandeo: Actually, I added `ts-tests` to spec_config - https://github.com/mongodb/mongo-ruby-driver/pull/2293/files#diff-88804426c739314d026c14eaa170a02831dd50c15481e2125d93fe74818ccb8bR596

Still failing without root_authorized.

lib/mongo/collection.rb
comandeo: Yes, copy paste leftover. Removed.

comandeo:
I am not sure if I am ready to add a documentation section. Server docs are not ready yet, afaik. Should I add some stub, or should we wait for server docs?

lib/mongo/collection.rb
```diff
@@ -239,6 +253,13 @@ def create(opts = {})
       options = Hash[self.options.reject do |key, value|
         %w(read read_preference read_concern).include?(key.to_s)
       end]
+      options.update(opts.slice(*TIME_SERIES_OPTIONS.keys))
+      # Converting Ruby spelled time series options to server style.
+      TIME_SERIES_OPTIONS.each do |ruby_key, server_key|
+        if options.has_key?(ruby_key)
```
p-mongo, line 259: We generally use `key?` instead of `has_key?` but you don't need to change this now.

p-mongo (approved):

p-mongo:
Hash#slice doesn't exist on Ruby 2.3, I suggest adding an implementation of it to Mongo::Utils a-la `compact`.

p-mongo:
For docs if you don't want to add them now either leaving the ticket open with a note or making a new ticket would be OK I'd say.

alcaeus:
> For docs if you don't want to add them now either leaving the ticket open with a note or making a new ticket would be OK I'd say.

We have [DRIVERS-1812](https://jira.mongodb.org/browse/DRIVERS-1812) to track this effort, so if you do create a ticket you can add a "split to" link to the drivers ticket to associate it. This will avoid creating an additional ticket when the drivers ticket is advanced through the workflow.

PR #2298, opened 2021-07-08
RUBY-2351 Increase minimum Ruby version to 2.4

comandeo:


.evergreen/config.yml
```diff
@@ -1412,7 +1408,7 @@ buildvariants:
       ocsp-algorithm: ecdsa
       ocsp-must-staple: on
       ocsp-delegate: on
-      ruby: [ruby-3.0, ruby-2.7, ruby-2.3]
```
p-mongo, line 1415: I'm pretty sure 2.3 is here as the lower bound of the supported versions, so I think it needs to be changed to 2.4 rather than removed unless ocsp+2.4 is tested elsewhere.

docs/reference/driver-compatibility.txt
```diff
@@ -243,7 +243,7 @@ for that Ruby version is deprecated.
      - |checkmark|
      - |checkmark|
      - D
-     - D
+     -
```
p-mongo, line 246: Can this PR go into 2.16 rather than 2.15?  Did we specifically intend to remove this support in 2.15?

lib/mongo/utils.rb
```diff
@@ -103,6 +103,7 @@ def initialize(**opts)
     end
 
     # Hash#compact implementation for Ruby 2.3
+    # @deprecated Use stdlib version since min Ruby version required is 2.4.
```
p-mongo, line 106: This entire module is private, so we can remove this method at any time. If you don't want to change calls throughout the tree to use Hash#compact now would you make a new ticket to do that please?

.evergreen/config.yml
p-mongo: I see that 2.3 had special code, so argument can be made that 2.4 configuration should've already been in place in addition to 2.3. I think we should test 2.4 as the lower bound.

docs/reference/driver-compatibility.txt
comandeo: I think the idea is to remove 2.3 support in the next release. I assumed it would be 2.15.

lib/mongo/utils.rb
comandeo: I think it's better to remove such code in this PR. Will do.

docs/reference/driver-compatibility.txt
p-mongo: We've already had 2.15 alpha and 2.15 has many other changes (i.e. all of the 5.0). Plus it's late in the dev cycle. I would suggest leaving Ruby version bump to 2.16, either by holding off on merging this PR or by branching 2.15-stable prior to the merge (holding off on merging this PR should be less work overall).

p-mongo:
I would say this can go into master now, one additional change needed is adding 2.16 to all of the compat tables.

docs/release-notes.txt
```diff
@@ -65,6 +65,8 @@ The following smaller improvements have been made:
   ``writeErrors`` field in the server response.
 - Server selection timeout for ``mongocryptd`` has been increased to 10 seconds.
 
+This release of the Ruby driver increases the minimum required Ruby version
+to 2.4.
```
p-mongo, line 69: This needs to be moved right?

p-mongo (approved):

PR #2300, opened 2021-07-09
RUBY-2649 Implement cursor pinning to services


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2649

lib/mongo/cursor.rb
```diff
@@ -448,6 +453,15 @@ def register
     def unregister
       cluster.unregister_cursor(@cursor_id)
     end
+
+    def execute_operation(op)
```
comandeo, line 457: 👍

comandeo (approved):

lib/mongo/server/connection_pool.rb
```diff
@@ -729,6 +747,16 @@ def self.finalize(available_connections, pending_connections, populator)
 
       private
 
+      def next_available_connection(service_id: nil)
+        if service_id
+          @available_connections.detect do |conn|
```
p-mongo, line 752: This does not remove the connection from available connections, I'll need to repair that.

PR #2301, opened 2021-07-11
RUBY-2722 Fix server monitoring with api version

comandeo:


lib/mongo/server/connection_common.rb
```diff
@@ -59,14 +59,16 @@ def connected?
...
       # @return [BSON::Document] Document that should be sent to a server
       #     for handshake purposes.
       #
       # @api private
-      def handshake_document(app_metadata, speculative_auth_doc: nil, load_balancer: false)
-        document = if app_metadata.server_api && app_metadata.server_api[:version]
-                     HELLO_DOC
+      def handshake_document(app_metadata, speculative_auth_doc: nil, load_balancer: false, server_api: {})
+        serv_api = app_metadata.server_api || server_api || {}
+        document = if serv_api[:version]
```
p-mongo, line 70: I don't think this is equivalent to what was in place before. An api version with, for example, only `:strict` set was previously sent to the server as `apiStrict` parameter but now wouldn't be sent, if I read this correctly.

comandeo: You're right, I didn't notice that.
However, I think that the new behavior is actually correct. According to versioned-api spec server api client options must include `version`, and may include `strict` and `deprecationErrors`. So, we should not send two latter to the server without a version.

alcaeus: That is correct - a serverApi object must always contain a version. Only passing a strict or deprecation option is invalid.

p-mongo (approved):

PR #2306, opened 2021-07-13
2678 rename sdam specs

comandeo:


comandeo:
Should be reviewed after https://github.com/mongodb/mongo-ruby-driver/pull/2305 is merged.

PR #2308, opened 2021-07-13
RUBY-2678 Rename SDAM specs

comandeo:


p-mongo:
I think the files you removed are Ruby driver-specific, can they be left as they are? Are they failing in any circumstances?

p-mongo (approved):

PR #2309, opened 2021-07-13
RUBY-2651 Implement transaction pinning to services


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2651

PR #2311, opened 2021-07-13
RUBY-2610 Add versioned API examples for docs

comandeo:


p-mongo:
I added a find call to ensure each client works and a close call to not leave background threads running.

PR #2316, opened 2021-07-16
RUBY-2738 Fix test failures against a serverless instance

comandeo:


p-mongo (approved):

p-mongo:
Though if some tests (e.g. crud) run against all mongoses, I suggest adding a configuration that runs those tests against the multiple mongoses.

p-mongo (approved):

PR #2320, opened 2021-08-02
RUBY-2578 Disable Ubuntu 14.04 based configs

comandeo:
https://jira.mongodb.org/browse/RUBY-2758

PR #2321, opened 2021-08-03
RUBY-2578 Repair Ubuntu 14.04 based configs

p-mongo:
Reverts mongodb/mongo-ruby-driver#2320

### DIFF
--- a/.evergreen/config-atlas.yml
+++ b/.evergreen/config-atlas.yml
@@ -606,10 +606,9 @@ axes:
       - id: ubuntu1604
         display_name: "Ubuntu 16.04"
         run_on: ubuntu1604-small
-      # https://jira.mongodb.org/browse/RUBY-2758
-      # - id: ubuntu1404
-      #   display_name: "Ubuntu 14.04"
-      #   run_on: ubuntu1404-small
+      - id: ubuntu1404
+        display_name: "Ubuntu 14.04"
+        run_on: ubuntu1404-small
       - id: rhel70
         display_name: "RHEL 7"
         run_on: rhel70-small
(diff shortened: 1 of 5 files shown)


p-mongo:
Looks like OCSP verifier tests need a new certificate.

.evergreen/config-atlas.yml
```diff
@@ -606,10 +606,9 @@ axes:
       - id: ubuntu1604
         display_name: "Ubuntu 16.04"
         run_on: ubuntu1604-small
-      # https://jira.mongodb.org/browse/RUBY-2758
-      # - id: ubuntu1404
-      #   display_name: "Ubuntu 14.04"
-      #   run_on: ubuntu1404-small
+      - id: ubuntu1404
```
comandeo, line 609: I am sure I will be confused when id ubuntu1404 while it is actually ubuntu1604. Can we change the id? Or just use existing 1604?

comandeo (approved):

.evergreen/config-atlas.yml
p-mongo: So just switch everything to ubuntu1604?

comandeo: As I understand the changes, now both ubuntu1404 and and ubuntu1604 are based on `ubuntu1604-small`. So, they are basically the same, aren't they? If they are the same, I would rather stick with just one. If they are different, I would rename `ubuntu1404` to something that better reflects what it is.

comandeo (approved):

p-mongo:
https://jira.mongodb.org/browse/RUBY-2758

PR #2324, opened 2021-08-06
RUBY-2668 Implement LB handshake error handling

comandeo:


lib/mongo/server.rb
```diff
@@ -443,11 +443,13 @@ def with_connection(service_id: nil, &block)
     def handle_handshake_failure!
       yield
     rescue Mongo::Error::SocketError, Mongo::Error::SocketTimeoutError => e
-      unknown!(
-        generation: e.generation,
-        service_id: e.service_id,
-        stop_push_monitor: true,
-      )
+      unless load_balancer?
```
p-mongo, line 446: I don't think this is needed. The `unknown!` method already conditions on `load_balancer?` and it only clears the pool in that case, and only if the service id is given. So as long as the service id is nil when an error happens prior to handshake completion (which would be quite natural) I think no changes perhaps are needed for this ticket.

PR #2326, opened 2021-08-12
RUBY-2760 Prohibit connect: :load_balanced and :replica_set option combination


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2760

lib/mongo/client.rb
```diff
@@ -1278,6 +1278,20 @@ def validate_options!(addresses = nil)
         end
       end
 
+      if connect == :load_balanced
+        if addresses && addresses.length > 1
+          raise ArgumentError, "connect=load_balanced cannot be used with multiple seeds"
+        end
+
+        if options[:direct_connection]
+          raise ArgumentError, "connect=load_balanced cannot be used with load_balanced=true"
```
comandeo, line 1287: I'm not sure if I understand this error. Looking at `if` above, I think it should be like `connect=load_balanced cannot be used with direct_connection=true`. But this is already handled - https://github.com/mongodb/mongo-ruby-driver/blob/d9c18487e84682eec52d835df5d479e3dcd41780/lib/mongo/client.rb#L1251

You can also see it in your spec where you are expecting an error from existing code.

p-mongo: I think I botched the error message here.

p-mongo: I believe you are correct, this check is redundant. This combination is verified earlier in the `:direct_connection` block on line 1251.

comandeo (approved):

PR #2328, opened 2021-08-16
RUBY-2771 Allow awaited hello to fail with EOF when client is shutting down


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2771

comandeo (approved):

PR #2329, opened 2021-08-17
Fix RUBY-2675 QueryCache returns wrong results after partial iteration of result set, e.g. after calling `none?`


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2675

comandeo (approved):

PR #2330, opened 2021-08-20
RUBY-2966 Add ActiveJob middleware for query cache.

johnnyshields:
Fixes MONGOID-4990

I'm not sure the best way to test this. We could use mocks? I don't think we want to require Rails into the test suite...

### DIFF
--- a/docs/reference/query-cache.txt
+++ b/docs/reference/query-cache.txt
@@ -275,6 +275,9 @@ You may read more about system collections in the
 Query Cache Middleware
 ======================
 
+Rack Middleware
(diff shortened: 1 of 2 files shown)


johnnyshields:
Oleg –> the issue is with your tests. See comment here https://github.com/p-mongo/tests/commit/45184897bd39656508827c3f50e69479a9ff250b

p-mongo (approved):

Neilshweky (approved):

comandeo (approved):

PR #2331, opened 2021-08-21
Fix .evergreen/config.yml line which should be commented

johnnyshields:
In addition, the buildvariant `kerberos-integration-ubuntu` is missing fields `mongodb-version` and `topology`. All other nodes have those fields, so you might want to check it.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1230,7 +1230,7 @@ buildvariants:
 #      os: ubuntu1604
 #    display_name: "${compressor} ${mongodb-version} ${topology} ${auth-and-ssl} ${ruby}"
 #    tasks:
-      - name: "test-mlaunch"
+#      - name: "test-mlaunch"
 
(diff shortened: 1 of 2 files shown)


comandeo:
Good catch, thanks! I added the fix to the corresponding template, so that it is not gone next time we generate `config.yml`
Regarding `kerberos-integration-ubuntu` - I added omitted parameters so that all configurations look similar.

comandeo (approved):

p-mongo:
Kerberos integration tests use an existing server deployment and do not create their own servers. This is why the server version and topology fields do not apply to them.

comandeo:
> Kerberos integration tests use an existing server deployment and do not create their own servers. This is why the server version and topology fields do not apply to them.

Sorry, I thought we're just using defaults there. Reverted corresponding changes.

p-mongo (approved):

PR #2332, opened 2021-08-21
RUBY-2655 Setup build pipeline for outside contributors

johnnyshields:
Refer to comments in https://github.com/mongodb/mongoid/pull/5043

This is essentially the same script adapted for `mongo-ruby-driver`. (If you diff the two you can see the changed lines.)

This does not yet support:
* topology = `sharded-cluster`
* auth / ssl tests

Both of these would be possible to support if we can get a better Github Actions config for running MongoDB itself. Would suggest that MongoDB Drivers team make one and maintain it, it would be a big benefit to the open source community.


.github/workflows/test.yml
```diff
@@ -16,420 +16,37 @@ jobs:
...
-          gemfile: Gemfile
-          experimental: false
+          - os: macos
+            ruby: 3.0
+            mongodb: 5.0
+            topology: server
+          - os: ubuntu-latest
+            ruby: 3.0
+            mongodb: 5.0
+            topology: server
```
johnnyshields, line 31: Let's add Ruby 2.6 and 2.7 just for completeness

comandeo:
https://jira.mongodb.org/browse/RUBY-2655

comandeo:
@johnnyshields Thanks for pushing this topic! I adjusted `test.yml` config a bit. Now it executes on all topologies (except load-balanced).

However, I do not think that we need two identical set of jobs for evergreen and GH actions. Especially since Evergreen will stay with us for a while. I suggest that we have a smaller (but an extensive) subset of jobs executing on GH automatically without a need of a constant approval by maintainers. When GH jobs are green, we will manually trigger Evergreen.

Therefore, I do not think we actually need `make-github-actions` script anymore. The `test.yml` is quite compact, and it is easy enough to edit it manually.

What do you think?

johnnyshields:
@comandeo It's up to you. I  wrote that script as a proof-of-concept, to demonstrate that such a script was possible. If you want to scrap it and maintain the action yml by hand, that's your call. I'm fine either way.

comandeo (approved):

p-mongo (approved):

PR #2335, opened 2021-08-26
RUBY-2743 Versioned API Strict Migration Examples

comandeo:


p-mongo (approved):
It is actually not clear where the query is supposed to go. I noted that in https://jira.mongodb.org/browse/DRIVERS-1846?focusedCommentId=4028010&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4028010 and given the lack of clarity I'm OK with the PR as proposed since it technically meets the stated requirements.

PR #2350, opened 2021-09-14
correct links to csfle guide

terakilobyte:
This corrects the broken links on the client-side-encryption page. It needs to be backported to all affected versions. The `:drivers:` role is added to the docs companion repo.


p-mongo:
Thank you Nathan!

p-mongo:
I backported through 2.12 which I believe is the earliest affected version when we implemented FLE.

terakilobyte:
Great, thanks! I'll get the docs rebuilt.

PR #2353, opened 2021-09-21
RUBY-2769 Test Serverless behind a load balancer

comandeo:


comandeo:
https://jira.mongodb.org/browse/RUBY-2769

p-mongo (approved):

PR #2354, opened 2021-10-04
RUBY-2800 Document deprecation of server versions

comandeo:


docs/reference/driver-compatibility.txt
```diff
@@ -28,7 +28,9 @@ particular version of the driver will generally work with newer versions of
 the server but may not take advantage of the functionality released in the
 newer version of the server.
 
-The first column lists the driver versions.
+The first column lists the driver versions.“D” in other columns means support
+for that MongoDB version is deprecated and may be removed upcoming driver
```
p-mongo, line 32: I would say "is deprecated and will be removed in a future driver version".

p-mongo (approved):

PR #2356, opened 2021-10-08
Fix RUBY-2805 Push monitor thread can exit when address resolution fails


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2805

PR #2369, opened 2021-11-14
RUBY-2842: Sleep 200ms before reporting failed server heartbeat back

mic-kul:
This prevents from issuing too many requests and causing memory use spike


p-mongo:
The throttling in push monitor was already implemented, and this code was removed in https://jira.mongodb.org/browse/RUBY-2990 because the entire push monitor is now stopped when it fails with a network error.

PR #2370, opened 2021-11-16
RUBY-2841 Ignore read pref if any pre-5.0

comandeo:
Ignore read preference for $out/$merge on secondaries if any servers are pre-5.0.


behackett (approved):
I'm hopeless at Ruby, but the logic looks right to me.

PR #2372, opened 2021-11-23
RUBY-2405 Add Azure KMS support

comandeo:


lib/mongo/crypt/kms.rb
```diff
@@ -0,0 +1,77 @@
...
+          else
+            nil
+          end
+        end
+      end
+    end
+  end
+end
+
+require "mongo/crypt/kms/credentials.rb"
```
p-mongo, line 73: `.rb` shouldn't be needed here?

lib/mongo/crypt/kms/aws.rb
```diff
@@ -0,0 +1,128 @@
...
+          # @return [ String ] AWS KMS key.
+          attr_reader :key
+
+          # @return [ String | nil ] AWS KMS endpoint.
+          attr_reader :endpoint
+
+          FORMAT_HINT = "AWS key document  must be in the format: " +
+                        "{ region: 'REGION', key: 'KEY' }"
+
+        # Creates a master key document object form a parameters hash.
```
p-mongo, line 90: Missing one level of indentation in this comment.

lib/mongo/crypt/binding.rb
```diff
@@ -265,6 +265,8 @@ def self.setopt_log_handler(handle, log_callback)
       #   @param [ Integer ] aws_secret_access_key_len The length of the AWS
       #     secret access key (or -1 for a null-terminated string).
       #   @return [ Boolean ] Returns whether the option was set successfully.
+      #
+      #   @deprecated Use mongocrypt_setopt_kms_providers instead.
```
p-mongo, line 269: This entire module if I am not mistaken is api private, therefore if these methods aren't used they should be deletable altogether.

lib/mongo/crypt/kms.rb
```diff
@@ -0,0 +1,77 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module Mongo
+  module Crypt
+    module KMS
+      module Validations
```
p-mongo, line 21: I would mark this whole module api private rather than its only function.

p-mongo (approved):
Do the FLE configs test both AWS and Azure? If yes am I also understanding it correctly that if, e.g., one of the variables with credentials/azure params gets cleared out, all azure tests will be silently skipped and nobody would know this? If so having some kind of a check against this situation would be handy, e.g. if FLE is enabled then require that all needed credentials are present either during test setup or as one of the tests. If this or a similar check is already implemented and I missed it, my apologies.

lib/mongo/crypt/kms.rb
comandeo: Fixed, thanks

lib/mongo/crypt/kms/aws.rb
comandeo: Done

lib/mongo/crypt/binding.rb
comandeo: Removed all deprecated methods and corresponding tests. 👍

lib/mongo/crypt/kms.rb
comandeo: Done

comandeo:
Added a check that fails if `FLE` env variable is set, but some of KMS provider ones are missing.

PR #2373, opened 2021-11-30
RUBY-2742 Remove OP_KILL_CURSORS, OP_INSERT, OP_UPDATE, OP_DELETE implementations + tests

Neilshweky:
https://jira.mongodb.org/browse/RUBY-2742


lib/mongo/auth/aws/request.rb
```diff
@@ -86,7 +86,7 @@ def initialize(access_key_id:, secret_access_key:, session_token: nil,
         # @return [ String ] formatted_time ISO8601-formatted time of the
         #   request, as would be used in X-Amz-Date header.
         def formatted_time
-          @formatted_time ||= @time.utc.strftime('%Y%m%dT%H%M%SZ')
+          @formatted_time ||= @time.getutc.strftime('%Y%m%dT%H%M%SZ')
```
Neilshweky, line 89: Not sure where this came from! Seems to be something I did 2 summers ago that never got merged.

lib/mongo/operation/shared/write.rb
```diff
@@ -37,13 +37,7 @@ module Write
...
         server.with_connection(service_id: context.service_id) do |connection|
           validate!(connection)
-          op = if connection.features.op_msg_enabled?
-              self.class::OpMsg.new(spec)
-            elsif !acknowledged_write?
-              self.class::Legacy.new(spec)
-            else
-              self.class::Command.new(spec)
-            end
+          op = self.class::OpMsg.new(spec)
```
p-mongo, line 40: This is correct but for a different reason than I stated in our conversation - looking at this code again, I'm assuming the reason why Legacy was used was because the Command protocol doesn't have a way of specifying how to perform an unacknowledged write (i.e. it doesn't have a provision for the respective flag).

p-mongo (approved):

comandeo (approved):

lib/mongo/operation/shared/write.rb
Neilshweky: Ah gotcha!

PR #2374, opened 2021-12-01
RUBY-2405 Add GCP KMS support

comandeo:


comandeo:
@p-mongo Could you please take a look on this - https://github.com/mongodb/mongo-ruby-driver/pull/2374/files#diff-ed17cd6ab0eaf7614281a1b87af00303cc1ce60290a3209502ee385b787b209cR98

I am wondering if there is a better way.

docs/reference/client-side-encryption.txt
```diff
@@ -389,9 +390,16 @@ use that information to generate a data key.
 
 If you have created an Azure master key, note the tenant id, the client id, and
 the client secret of the application that has permissions to use the key.
-Additionally, note the key name, key version (id any), and key vault endpoint
+Additionally, note the key name, key version (if any), and key vault endpoint
 for your master key. You will use that information to generate a data key.
 
+If you have created an GCP KMS master key, note the email and the private key
```
p-mongo, line 396: *a* GCP KMS master key

docs/reference/client-side-encryption.txt
```diff
@@ -389,9 +390,16 @@ use that information to generate a data key.
...
+the client secret of the application that has permissions to use the key.
```
p-mongo, line 397: I think grammar is broken between the previous line and this line.

lib/mongo/crypt/kms/azure.rb
```diff
@@ -105,7 +105,7 @@ class MasterKeyDocument
         #
         # @raise [ ArgumentError ] If required options are missing or incorrectly.
           def initialize(opts)
-            if opts.is_a?(Hash)
+            unless opts.is_a?(Hash)
```
p-mongo, line 108: Interesting...

lib/mongo/crypt/kms/gcp.rb
```diff
@@ -0,0 +1,151 @@
...
+    module KMS
+      module GCP
+        # GCP Cloud Key Management Credentials object contains credentials for
+        # using GCP KMS provider.
+        #
+        # @api private
+        class Credentials
+          include KMS::Validations
+
+          # @return [ String ] GCP email to authenticate.
```
p-mongo, line 29: This should probably say "authenticate *with*"?

lib/mongo/crypt/hooks.rb
```diff
@@ -88,6 +88,19 @@ def hash_sha256(input)
...
       end
       module_function :hash_sha256
+
+      # An RSASSA-PKCS1-v1_5 with SHA-256 signature function.
+      #
+      # @param [ String ] key The PKCS#8 encrypted private key.
+      # @param [ String ] input The data to be signed.
+      #
+      # @return [ String ] The signature.
+      def rsaes_pkcs_signature(key, input)
```
p-mongo, line 98: Where are the requirements/specification for this method?

docs/reference/client-side-encryption.txt
```diff
@@ -412,6 +420,10 @@ for your master key. You will use that information to generate a data key.
         tenant_id: 'AZURE-TENANT-ID',
         client_id: 'AZURE-CLIENT-ID',
         client_secret: 'AZURE-CLIENT-SECRET'
+      },
+      gcp: {
+        email: 'GCP-EMAIL',
+        private_key: 'GCP-PRIVATE-KEY',
```
p-mongo, line 426: I would specify what format the private key can be in here, in light of the test code wrapping it with pem line markers. Is it some format that GCP provides?

p-mongo (approved):

lib/mongo/crypt/kms/azure.rb
comandeo: Yeah, in the previous PR there were no test for Data Key for Azure. Therefore this error got through. In this PR I added missing tests for Azure, and for GCP.

lib/mongo/crypt/hooks.rb
comandeo: Some information is available here - https://docs.google.com/document/d/18rO1QYORNvKs8rYWtXKdrT-2YaORMJ-r8xcM8g1SJ8Q/edit#heading=h.h2339z2vnbng

There are also some further links:
https://www.rfc-editor.org/rfc/rfc7518.html#section-3.3
https://www.rfc-editor.org/rfc/rfc3447#section-8.2

p-mongo: I think it would be helpful to add a unit test that signs the "data to sign" as per that section in the google doc. The implementation looks good to me. The other part that that test would possibly clarify is whether the signature as produced by ruby's openssl binding is in the expected format vs mongocrypt/that doc (e.g. base64 encoded?)

docs/reference/client-side-encryption.txt
comandeo: Fixed

docs/reference/client-side-encryption.txt
comandeo: Fixed, thanks.

lib/mongo/crypt/kms/azure.rb
comandeo: Fixed

lib/mongo/crypt/hooks.rb
comandeo: Test added.

docs/reference/client-side-encryption.txt
comandeo: We managed to accept both DER and PEM. Documentation updated accordingly.

PR #2375, opened 2021-12-01
RUBY-2770 Use OP_MSG for exhaust cursors if the server supports OP_MSG

Neilshweky:


lib/mongo/server/monitor.rb
```diff
@@ -282,9 +282,7 @@ def check
           result = server.round_trip_time_averager.measure do
             begin
               doc = @connection.check_document
-              cmd = Protocol::Query.new(
-                Database::ADMIN, Database::COMMAND, doc, :limit => -1
-              )
+              cmd = Protocol::Msg.new([], {}, doc.merge({'$db' => Database::ADMIN}))
```
Neilshweky, line 285: Is it okay to always use OP_MSG, being that the Ruby driver doesn't support servers older than 3.6?

p-mongo: The impact of this change should be less clear error reporting when the driver connects to old servers.

Without this change, the driver should fail the wire version range check and report to the application that the driver supports wire versions X-Y but the server returned Z.

With this change, the driver won't be able to speak to the server at all (successfully) and will produce a monitoring/ismaster error of some kind, in my estimation.

Given that there are many places in the tree that use `Protocol::Query` right now, my suggestion would be to defer this change until the driver completely removes pre-OP_MSG protocol implementations.

Neilshweky: What does that mean in terms of this ticket?

PR #2379, opened 2021-12-07
RUBY-2525 Expose the Reason an Operation Fails Document Validation

Neilshweky:
I decided to add a details property to the `OperationFailure` class to better conform with the crud spec. 
This attribute is defined as follows:

- For WriteConcernErrors this is `document['writeConcernError']['errInfo']`.
- For WriteErrors this is `document['writeErrors'][0]['errInfo']`.
- For all other errors this is `nil`.

Now, this makes a lot of sense for the OperationFailure class, but for BulkWriteError it doesn't quite make sense. The problem is that BulkWriteError can have multiple writeErrors inside it. This is troublesome because I'm not sure what to do with the `details` now. Should it be a list of the `errInfo`s of each `writeError`? This didn't seem to make sense, so I just left the details off the `BulkWriteError` class. If the user wants to find that information, they can get there from the `results` `attr_reader`.

[Evergreen Patch](https://spruce.mongodb.com/version/61b0c81f3e8e862ec357dc60/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/lib/mongo/error/bulk_write_error.rb
+++ b/lib/mongo/error/bulk_write_error.rb
@@ -59,11 +59,20 @@ module Mongo
         errors = @result['writeErrors']
         return nil unless errors
 
-        fragment = errors.first(10).map do |error|
-          "[#{error['code']}]: #{error['errmsg']}"
-        end.join('; ')
+        fragment = ""
+        cut_short = false
+        errors.first(10).each_with_index do |error, i|
+          fragment += "; " if fragment.length > 0
+          fragment += "[#{error['code']}]: #{error['errmsg']}"
+          fragment += " -- #{error['errInfo'].to_json}" if error['errInfo']
 
-        fragment += '...' if errors.length > 10
+          if fragment.length > 3000
+            cut_short = i < [9, errors.length].min
+            break
+          end
+        end
+
+        fragment += '...' if errors.length > 10 || cut_short
 
         if errors.length > 1
           fragment = "Multiple errors: #{fragment}"
--- a/lib/mongo/error/bulk_write_error.rb
+++ b/lib/mongo/error/bulk_write_error.rb
@@ -55,6 +55,21 @@ module Mongo
 
       private
 
+      # Generates an error message when there are multiple write errors.
+      # The message is generated as follows:
+      #
+      # col has validation { 'validator' => { 'x' => { '$type' => 'string' } } }
+      # col.insert_many([{_id: 1}, {_id: 2}], ordered: false)
+      #
+      # Multiple errors:
+      #   [121]: Document failed validation --
+      #     {"failingDocumentId":1,"details":{"operatorName":"$type",
+      #     "specifiedAs":{"x":{"$type":"string"}},"reason":"field was
+      #     missing"}};
+      #   [121]: Document failed validation --
+      #     {"failingDocumentId":2, "details":{"operatorName":"$type",
+      #     "specifiedAs":{"x":{"$type":"string"}}, "reason":"field was
+      #     missing"}}
       def build_message
         errors = @result['writeErrors']
         return nil unless errors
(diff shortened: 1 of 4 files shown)


lib/mongo/error/operation_failure.rb
```diff
@@ -306,6 +316,28 @@ def unsupported_retryable_write?
...
+      private
+
+      # Retrieve the details from a document
+      #
+      # @return [ Hash | nil ] the details extracted from the document
+      def retrieve_details(document)
+        return nil unless document
+        if document['writeConcernError']
+          return document['writeConcernError']['errInfo']
+        elsif document['writeErrors'] && document['writeErrors'][0]
```
Neilshweky, line 329: Is there a way to do this that's more idiomatic to ruby?

alexbevi: @Neilshweky Ruby's [safe navigation operator](https://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/) might work for this:

```ruby
d1 = {}
d2 = {}
d2["writeErrors"] = [1,2,3]
d1
 => {} 
d2
=> {"writeErrors"=>[1, 2, 3]} 
d1["writeErrors"]&.first
=> nil
d2["writeErrors"]&.first
=> 1 
```

so maybe something along the lines of the following might work?

```ruby
elsif document['writeErrors']&.first
```

lib/mongo/error/operation_failure.rb
```diff
@@ -224,6 +224,14 @@ def write_concern_error?
       # @since 2.10.0
       attr_reader :write_concern_error_code_name
 
+      # @return [ String | nil ] The details of the error.
+      #   For WriteConcernErrors this is `document['writeConcernError']['errInfo']`.
+      #   For WriteErrors this is `document['writeErrors'][0]['errInfo']`.
+      #   For all other errors this is nil.
+      #
+      # @since 2.11.0
```
p-mongo, line 232: We no longer use `@since` tags and this one isn't accurate, since the functionality will be added in 2.18. You can remove this `@since` tag and, if others are in your way going forward, feel free to get rid of them also.

lib/mongo/error/operation_failure.rb
p-mongo: You cound do `document['writeConcernError']&.[]('errInfo') || document['writeErrors']&.first&.[]('errInfo')` but whether this is more readable is open to interpretation.

I would assign intermediate values to variables to avoid repeated traversal for both runtime performance and (hopefully) ease of readability:

```

        if wce = document['writeConcernError']
          return wce['errInfo']
...
```

lib/mongo/error/bulk_write_error.rb
```diff
@@ -59,11 +59,20 @@ def build_message
...
 
-        fragment = errors.first(10).map do |error|
-          "[#{error['code']}]: #{error['errmsg']}"
-        end.join('; ')
+        fragment = ""
+        cut_short = false
+        errors.first(10).each_with_index do |error, i|
+          fragment += "; " if fragment.length > 0
+          fragment += "[#{error['code']}]: #{error['errmsg']}"
+          fragment += " -- #{error['errInfo'].to_json}" if error['errInfo']
```
p-mongo, line 67: Can you please add a comment somewhere in this file that shows an example message that is generated in this way, i.e. one containing multiple errors and the double dashes.

lib/mongo/error/operation_failure.rb
Neilshweky: done

spec/mongo/error/operation_failure_heavy_spec.rb
```diff
@@ -58,4 +61,50 @@
...
+        }
+      }].create
+      subscribed_client[:collection_name]
+    end
+
+    context 'when there is a write error' do
+      it 'succeeds and prints the error' do
+        begin
+          collection.insert_one({x: 1})
+        rescue Mongo::Error => e
```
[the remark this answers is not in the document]
Neilshweky: done

p-mongo (approved):
@comandeo ready for you

comandeo (approved):

PR #2380, opened 2021-12-09
RUBY-2746 RUBY-2851 RUBY-2737 Provide options to limit number of mongos servers used in connecting to sharded clusters

Neilshweky:
Adding ticket [RUBY-2851](https://jira.mongodb.org/browse/RUBY-2851) to this PR. This ticket is for removing the DNS spec tests that conflict with url-options spec tests.
Also adding [RUBY-2737](https://jira.mongodb.org/browse/RUBY-2737) to this PR. This ticket is for adding the `srvServiceName` option, and is completely implemented by this ticket.


lib/mongo/srv/resolver.rb
```diff
@@ -51,13 +61,18 @@ def timeout
...
       #
       # In the event that a record with a mismatched domain is found or no
       # records are found, if the :raise_on_invalid option is true,
       # an exception will be raised, otherwise a warning will be logged.
       #
       # @param [ String ] hostname The hostname whose records should be obtained.
+      # @param [ String | nil ] srv_service_name The SRV service name for the DNS query.
+      #   If nil, 'mongodb' is used.
+      # @param [ Integer | nill ] srv_max_hosts The maximum number of records to return.
+      #   If this value is zero or nil, return all of the records.
```
p-mongo, line 75: I think it would be better to change zero to nil sometime during client construction and everywhere downstream of the client require and expect (with validation, if necessary) that the value is either nil or a positive integer.

lib/mongo/uri.rb
```diff
@@ -538,6 +538,28 @@ def validate_uri_options!
           raise_invalid_error_no_fmt!("loadBalanced=true cannot be used with replicaSet option")
         end
       end
+
+      # Can I do this check here? because I have to check the scheme of the URI.
```
p-mongo, line 542: It's  unconventional but I don't see obvious problems with doing this.

lib/mongo/uri.rb
```diff
@@ -538,6 +538,28 @@ def validate_uri_options!
...
+      unless self.is_a?(URI::SRVProtocol) # TODO: is it better practice to leave off the self?
```
p-mongo, line 543: Either way is fine with me.

lib/mongo/uri.rb
```diff
@@ -538,6 +538,28 @@ def validate_uri_options!
...
+
+        if uri_options[:srv_service_name]
+          raise_invalid_error_no_fmt!("srvServiceName cannot be used on non-SRV URI")
+        end
+      end
+
+      # byebug
+      if uri_options[:srv_max_hosts] && uri_options[:srv_max_hosts] > 0
+        if uri_options[:replica_set]
+          raise_invalid_error_no_fmt!(":srv_max_hosts > 0 cannot be used with replica_set option")
```
p-mongo, line 556: These messages should use the URI option names, e.g. srvMaxHosts, replicaSet.

spec/mongo/client_construction_spec.rb
```diff
@@ -1513,6 +1513,86 @@
...
+          end
+
+          it 'it is rejected' do
+            expect do
+              client
+            end.to raise_error(ArgumentError, /:srv_max_hosts > 0 cannot be used with load_balanced=true/)
+          end
+        end
+
+        context 'srv_max_hosts > 0 and replicaSet' do
```
p-mongo, line 1530: `replica_set`, please check the naming of other options in this file too.

p-mongo (commented):
Is there a test that specifying srv_max_hosts actually reduces the number of seeds compared to what is in the DNS? Apologies if I missed it, if not it should be added.

Neilshweky:
> Is there a test that specifying srv_max_hosts actually reduces the number of seeds compared to what is in the DNS? Apologies if I missed it, if not it should be added.

There are spec tests that do it (`srvMaxHosts-less_than_srv_records.yml`) but I didn't write any Ruby tests. Should I?

p-mongo:
The srvMaxHosts-less_than_srv_records.yml test doesn't assert the resulting seed list, therefore I expect it to pass if the driver e.g. happens to replace the seed list  with a single bogus value. Looking at the test runner implementation, as far as I can tell that test would also be silently not performed if the file format changed (as opposed to producing an error due to unexpected test format).

Based on this, I suggest adding a test which uses one of the test URIs, create a client with it, then inspect the seed list and verify that the seed is one of expected values.

Neilshweky:
> The srvMaxHosts-less_than_srv_records.yml test doesn't assert the resulting seed list, therefore I expect it to pass if the driver e.g. happens to replace the seed list with a single bogus value. Looking at the test runner implementation, as far as I can tell that test would also be silently not performed if the file format changed (as opposed to producing an error due to unexpected test format).
> 
> Based on this, I suggest adding a test which uses one of the test URIs, create a client with it, then inspect the seed list and verify that the seed is one of expected values.

It's difficult to assert the actual seedless, since if max hosts is less than the number of hosts, it selects a random subset of hosts of size `srv_max_hosts`. I can assert that the hosts that it does come up with are valid, though.

docs/reference/create-client.txt
```diff
@@ -694,6 +694,20 @@ Ruby Options
...
+     - The maximum number of mongos connections that may be created
+       for sharded topologies. If this option is set to 0, there will
+       be no maximum number of connections. If the given URI resolves
+       to more hosts than ``:srv_max_hosts``, the client will ramdomly
+       choose an ``:srv_max_hosts`` sized subset of hosts.
+     - ``Integer``
+     - 0
+
+   * - ``:srv_service_name``
+     - The SRV service name to use in the DNS query.
```
p-mongo, line 707: I think "The service name to use in the SRV DNS query" would be more clear.

docs/reference/create-client.txt
```diff
@@ -694,6 +694,20 @@ Ruby Options
      - ``Float``
      - none
 
+   * - ``:srv_max_hosts``
+     - The maximum number of mongos connections that may be created
+       for sharded topologies. If this option is set to 0, there will
+       be no maximum number of connections. If the given URI resolves
+       to more hosts than ``:srv_max_hosts``, the client will ramdomly
+       choose an ``:srv_max_hosts`` sized subset of hosts.
```
p-mongo, line 702: I would add here a note that the hosts that the driver ignores during client construction will never be used, i.e., if a deployment has 10 total mongoses and this option limits them to 5, and those 5 happen to become unavailable, the client will quit working completely even though the deployment overall still has 5 functional mongoses elsewhere.

docs/reference/create-client.txt
```diff
@@ -694,6 +694,20 @@ Ruby Options
      - ``Float``
      - none
 
+   * - ``:srv_max_hosts``
+     - The maximum number of mongos connections that may be created
+       for sharded topologies. If this option is set to 0, there will
```
p-mongo, line 699: I would adjust the first sentence to something like "The maximum number of mongoses that the driver will communicate with", because the number of connections is controlled by pool size and can exceed 1 per mongos.

lib/mongo/client.rb
```diff
@@ -92,6 +92,8 @@ class Client
       :server_api,
       :server_selection_timeout,
       :socket_timeout,
+      :srv_max_hosts,
+      :srv_service_name,
```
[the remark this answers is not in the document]
p-mongo: This still needs to be done right?

spec/mongo/uri/srv_protocol_spec.rb
```diff
@@ -263,15 +263,68 @@
...
+            expect(uri.servers.length).to eq 2
+            uri.servers.should =~ hosts
+          end
+        end
+
+        context 'with srvMaxHosts' do
+          let(:options) { '/?srvMaxHosts=1' }
+          it 'returns an array with only one of the parsed servers' do
+            expect(uri.servers.length).to eq 1
+            expect(hosts.include?(uri.servers.first)).to be true
```
p-mongo, line 291: :+1:

p-mongo (approved):

comandeo (approved):
👍

PR #2383, opened 2021-12-14
RUBY-2833 Add KMIP Support

comandeo:


.evergreen/config-atlas.yml
```diff
@@ -420,6 +420,20 @@ functions:
...
           SERVERLESS=1 SSL=ssl RVM_RUBY="${RVM_RUBY}" SINGLE_MONGOS="${SINGLE_MONGOS}" MULTI_ATLASPROXY_SERVERLESS_URI="${MULTI_ATLASPROXY_SERVERLESS_URI}" SINGLE_ATLASPROXY_SERVERLESS_URI="${SINGLE_ATLASPROXY_SERVERLESS_URI}" .evergreen/run-tests-serverless.sh
 
+  "stat kmip server":
+    - command: shell.exec
+      params:
+        background: true
+        shell: bash
+        working_dir: "src"
+        script: |
+          source .evergreen/csfle/activate_venv.sh
```
p-mongo, line 430: `source` is a bashism, the portable equivalent of it is `.` (the dot):

```
. .evergreen/csfle/activate_venv.sh
```

.evergreen/config.yml
```diff
@@ -471,6 +485,7 @@ tasks:
   - name: "test-fle"
     commands:
       - func: "export FLE credentials"
+      - func: "stat kmip server"
```
p-mongo, line 488: start

.evergreen/run-tests.sh
```diff
@@ -139,6 +139,25 @@ if test -n "$FLE"; then
 
   export LIBMONGOCRYPT_PATH=`pwd`/rhel-70-64-bit/nocrypto/lib64/libmongocrypt.so
   test -f "$LIBMONGOCRYPT_PATH"
+
+  echo "Waiting for mock KMS servers to start..."
+   wait_for_kms_server() {
+      for i in $(seq 60); do
+         # Exit code 7: "Failed to connect to host".
+         if curl -s "localhost:$1"; test $? -ne 7; then
```
p-mongo, line 147: Wouldn't it better to require status 0 than anything other than 7?

docs/reference/client-side-encryption.txt
```diff
@@ -523,6 +546,50 @@ keys are stored in the ``admin`` database in the ``datakeys`` collection:
...
+      kms_providers: {
+        kmip: {
+          endpoint: 'KMIP-SERVER-HOST'
+        }
+      },
+      kms_tls_options: {
+        kmip: {
+          ssl_ca_cert: 'PATH-TO-CA-FILE',
+          ssl_cert: 'PATH-TO-CLIENT-CERT-FILE',
+          ssl_key: 'PATH-TO-CLIENT-KEY-FILE'
```
p-mongo, line 586: We have an open ticket to rename ssl prefixes to tls. Do you have an idea how feasible it would be to use the tls prefixes in the FLE changes you've been making recently?

p-mongo (approved):

.evergreen/config-atlas.yml
comandeo: Fixed

.evergreen/config.yml
comandeo: Fixed, thanks.

.evergreen/run-tests.sh
comandeo: Some of mock servers we start have bad configurations - wrong certificates, etc. So, curl won't be able to connect to them "good enough" to return 0. Hence check for "anything but Failed to connect".

docs/reference/client-side-encryption.txt
comandeo: The spec says `The TLS options SHOULD be consistent with the existing TLS options for MongoDB server TLS connections.` I understand this as we should have the same set of TLS options for our client, and for KMIP.

.evergreen/run-tests.sh
p-mongo: How about adding the `-k` option in this case?

docs/reference/client-side-encryption.txt
p-mongo: Let's punt on the renaming and do it all together later.

.evergreen/run-tests.sh
comandeo: Just adding `-k` didnt't work. I end up with this - `curl -d '' -X POST -k https://localhost:8002/`, which works (== returns 0) for almost all mock servers. Unfortunately, an attempt to connect to mock server with `--require_client_cert` option will give us error code 56.

p-mongo: OK, whatever you think is best is good with me.

PR #2384, opened 2021-12-14
RUBY-2814 listCollectionNames should use authorizedCollections argument

Neilshweky:
All the tests that I had to edit here are for server versions >= 4.0. Should I be adding tests specifically for < 4.0 being that nothing changed for those versions? 

UPDATE: This PR is no longer necessary as after the update of [DRIVERS-1393](https://jira.mongodb.org/browse/DRIVERS-1393), the ruby driver already fulfills it's requirements. Closing this PR.


p-mongo:
:+1:

PR #2404, opened 2022-01-11
RUBY-2822 Add support for comment field

comandeo:


lib/mongo/collection/view/readable.rb
```diff
@@ -281,6 +289,7 @@ def estimated_document_count(opts = {})
                   db_name: database.name,
                   read: read_pref,
                   session: session,
+                  comment: opts[comment],
```
Neilshweky, line 292: I was looking at the spec for `estimatedDocumentCount` and it doesn't look like the comment field is included there. Is that a mistake here or there? Or I could just be reading the spec wrong.

Neilshweky: Or are all the options from CountOptions included in EstimatedCountOptions as well?

Neilshweky (approved):
LGTM, just fix those typos on the comments

docs/release-notes.txt
```diff
@@ -35,6 +35,10 @@ This release includes the following new features:
   derived from Object. BSON::DBRef retains all fields, unlike Mongo::DBRef which
   only allowed $ref, $id, and $db. BSON::DBRef also reorders the fields if
   necessary to place $ref, $id, and $db first, in that order.
+- CRUD methods, methods for listing databases, collection, and indexes management
+  methods now supports a new option ``comment``. This option enables users to
```
p-mongo, line 39: support (singular)

spec/mongo/client_spec.rb
```diff
@@ -584,6 +592,17 @@
...
     end
+
+    context 'with comment' do
+      min_server_version '4.4'
+
+      it 'returns a list of database names and send comment' do
+        result = monitored_client.database_names({}, comment: "comment")
+        expect(result).to include('admin')
+        command = subscriber.command_started_events("listDatabases").last&.command
+        expect(command&.fetch("comment")).to eq("comment")
```
p-mongo, line 603: I would assert that the values here aren't nil, same with other tests.

lib/mongo/client.rb
```diff
@@ -925,6 +927,8 @@ def database_names(filter = {}, opts = {})
     #   See https://docs.mongodb.com/manual/reference/command/listDatabases/
     #   for more information and usage.
     # @option opts [ Session ] :session The session to use.
+    # @option options [ String, Bson::Document, Hash ] :comment A user-provided
```
p-mongo, line 930: Having both Hash and BSON::Document (and `BSON` needs to be capitalized there) is weird because the latter is a subclass of the former. In some of the methods, when Hash is specified what is implied is that the keys must be of the correct type (i.e. either symbols or strings - the driver will not cast Hash to B::D for indifferent access). When B::D is specified, the driver either may typecast to B::D which makes it actually accept Hash instances with wrong key types, or the driver may require actual B::D instances and fail if Hash instances are given with wrong types. So this can be confusing.

In case of comment, since the driver is not interpreting the value at all, I would tend toward documenting it as a Hash if the only two allowed types are String and Hash. If all types are allowed per Neil's comment, I suggest `Object` as the type.

lib/mongo/index/view.rb
```diff
@@ -78,25 +79,30 @@ class View
...
       #
+      # @param [ String, Bson::Document, Hash ] comment A user-provided
+      #   comment to attach to this command.
+      #
       # @return [ Result ] The response.
       #
       # @since 2.0.0
-      def drop_all
-        drop_by_name(Index::ALL)
+      def drop_all(comment: nil)
```
p-mongo, line 104: I think explicitly specifying options like this is problematic because it is not forward-compatible, i.e. if a future driver version adds another option, application cannot specify that option when running on an older driver. I would use '**options` or `options` as we do elsewhere.

lib/mongo/client.rb
```diff
@@ -1013,6 +1019,8 @@ def start_session(options = {})
     #   changes that occurred at or after the specified timestamp. Any command run
     #   against the server will return a cluster time that can be used here.
     #   Only recognized by server versions 4.0+.
+    # @potion options [ String, Bson::Document, Hash ] :comment A user-provided
```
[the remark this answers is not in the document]
comandeo: Thanks, fixed.

lib/mongo/collection.rb
```diff
@@ -386,7 +387,8 @@ def find(filter = nil, options = {})
     # @option options [ true, false ] :bypass_document_validation Whether or
     #   not to skip document level validation.
     # @option options [ Hash ] :collation The collation to use.
-    # @option options [ String ] :comment Associate a comment with the aggregation.
+    # @potion options [ String, Bson::Document, Hash ] :comment A user-provided
```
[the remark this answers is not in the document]
comandeo: Good point, thanks! Actually, the comment can be of any type. I replaced with `Object` as Oleg suggested below.

lib/mongo/collection/view/readable.rb
comandeo: I think I just forgot to add `comment` to `EstimatedCountOptions`. This helper actually uses `count` command, so `comment` works.
I'll create a DRIVERS ticket and PR to fix `EstimatedCountOptions`. I think we do not need to wait until spec is updated, and can leave this change.

docs/release-notes.txt
comandeo: Fixed.

lib/mongo/index/view.rb
comandeo: Thanks, changed to `options`.

lib/mongo/client.rb
comandeo: Yes, good point. Changed to `Object` since comment can be of any type.

spec/mongo/client_spec.rb
comandeo: Updated.

lib/mongo/collection/view/readable.rb
```diff
@@ -127,7 +128,7 @@ def batch_size(batch_size = nil)
         # @note Set profilingLevel to 2 and the comment will be logged in the profile
         #   collection along with the query.
         #
-        # @param [ String ] comment The comment to be associated with the query.
+        # @param [ String, Hash, BSON::Document ] comment The comment to be associated with the query.
```
p-mongo, line 131: `Object` ?

lib/mongo/collection/view/readable.rb
```diff
@@ -153,6 +154,8 @@ def comment(comment = nil)
         # @option opts [ Hash ] :read The read preference options.
         # @option opts [ Hash ] :collation The collation to use.
         # @option opts [ Mongo::Session ] :session The session to use for the operation.
+        # @option options [ Object ] :comment A user-provided
```
p-mongo, line 157: `opts`

lib/mongo/index/view.rb
```diff
@@ -78,25 +79,32 @@ class View
...
       #
       # @param [ String ] name The name of the index.
+      # @param [ Object ] comment A user-provided
+      #   comment to attach to this command.
       #
       # @return [ Result ] The response.
       #
       # @since 2.0.0
-      def drop_one(name)
+      def drop_one(name, comment: nil)
```
p-mongo, line 88: This also needs to use  `options`.

lib/mongo/collection/view/readable.rb
comandeo: Fixed.

p-mongo (approved):

PR #2417, opened 2022-02-08
RUBY-2869 Do not use mutexes in finalizers

comandeo:
Finalizers are called from signal trap, so mutexes do not work there (see https://github.com/ruby/ruby/blob/master/thread_sync.c#L288).

### DIFF
--- a/lib/mongo/cluster/reapers/cursor_reaper.rb
+++ b/lib/mongo/cluster/reapers/cursor_reaper.rb
@@ -44,6 +44,7 @@ module Mongo
         @to_kill = {}
         @active_cursor_ids = Set.new
         @mutex = Mutex.new
+        @pipe_read, @pipe_write = IO.pipe
       end
 
       attr_reader :cluster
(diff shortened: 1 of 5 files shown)


Neilshweky (approved):

lib/mongo/cluster/reapers/cursor_reaper.rb
```diff
@@ -56,12 +62,14 @@ def initialize(cluster)
...
-          end
-        end
+        msg = [
+          server.address.seed,
+          kill_spec.cursor_id,
+          kill_spec.coll_name,
+          kill_spec.db_name,
+          kill_spec.service_id
+        ].join(MSG_SEPARATOR)
+        @pipe_write.puts(msg)
```
p-mongo, line 72: This call can block if the pipe is full. My understanding is there needs to be  a thread that is constantly reading the pipe to ensure the writing side doesn't block indefinitely.

p-mongo: Have you considered https://ruby-doc.org/core-2.5.0/Queue.html ?

comandeo: I have, but `Queue` is thread safe, and according to its docs _"The [Queue](https://ruby-doc.org/core-2.5.0/Queue.html) class implements all the required locking semantics."_ So, we may encounter the same problem with queue. I'll do some testing, though. If it works, I would prefer queue over pipe here.

comandeo:
> I think it is more correct to use server addresses (`Address` instances) rather than seeds. The seeds may be missing port numbers.

Good point, changed accordingly.

PR #2418, opened 2022-02-17
Use encoding objects for comparison instead of encoding names

casperisfine:
Comparing the names is much less efficient than
comparing the instance directly.

It may also change in the future: https://bugs.ruby-lang.org/issues/18576


p-mongo (approved):

casperisfine:
> The performance impact of this comparison is not significant because it happens in the test suite.

Yeah sorry. I just sent 31 similar pull requests, so I just copy pasted the commit message.

Neilshweky (approved):

PR #2419, opened 2022-02-22
RUBY-2909 Snapshot Query Examples for the Manual

Neilshweky:


Neilshweky:
After removing `require_stress` a lot of executions failed because they took too long. I added it back.

spec/integration/snapshot_query_examples_spec.rb
```diff
@@ -0,0 +1,115 @@
+# frozen_string_literal: true
+# encoding: utf-8
+
+require 'spec_helper'
+
+describe 'Snapshot Query Examples' do
+  require_topology :replica_set, :sharded
+  require_stress
```
p-mongo, line 8: I don't think these tests will be executed when you `require_stress` but the file is not under `spec/stress`:

```
.evergreen/run-tests.sh:  bundle exec rspec spec/integration/fork*spec.rb spec/stress/fork*spec.rb
.evergreen/run-tests.sh:  bundle exec rspec spec/integration/fork*spec.rb spec/stress
```

Either this invocation needs to be adjusted or the file needs to be moved under spec/stress.

spec/integration/snapshot_query_examples_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    "mongodb://#{SpecConfig.instance.addresses.join(',')}/?w=majority"
+  end
+
+  context "Snapshot Query Example 1" do
+    before do
+      client = Mongo::Client.new(uri_string, database: "pets")
+      client['cats'].delete_many({})
+      client['dogs'].delete_many({})
+
+      client['cats'].insert_one({
```
p-mongo, line 21: While the curly braces `{}` in this case can be provided, they are redundant and most Ruby developers won't use them.  I suggest removing them.

spec/integration/snapshot_query_examples_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+                            startDate: "$$NOW",
+                            unit: "day",
+                            amount: 1
+                         }
+                      }
+                   ]
+                 }
+              }
+          },
+          { "$count": "totalDailySales" }
```
p-mongo, line 105: I would also rename this to `total_daily_sales`.

spec/integration/snapshot_query_examples_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+  min_server_fcv '5.0'
+
+  let(:uri_string) do
+    "mongodb://#{SpecConfig.instance.addresses.join(',')}/?w=majority"
+  end
+
+  context "Snapshot Query Example 1" do
+    before do
+      client = Mongo::Client.new(uri_string, database: "pets")
+      client['cats'].delete_many({})
```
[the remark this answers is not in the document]
Neilshweky: done

spec/integration/snapshot_query_examples_spec.rb
Neilshweky: done

spec/integration/snapshot_query_examples_spec.rb
Neilshweky: done

spec/integration/snapshot_query_examples_spec.rb
Neilshweky: done

p-mongo:
I tested this PR without the stress constraint in https://github.com/mongodb/mongo-ruby-driver/pull/2422 and the first failures I see are in auth configurations. Because the tests in this PR are creating a client with no options, they are subject to the default 30 second timeout when the server requires auth and the client doesn't provide it. If the client instantiation is part of the example and needs to stay the way it is, the easiest solution here in my opinion is to use `require_no_auth` to skip the tests on auth configurations and subsequently check remaining failures to skip on other configurations as needed (e.g. TLS). If the client instantiation is not part of the example and can be modified, the appropriate method to use is `new_local_client` with `SpecConfig.instance.test_options` to get the correct options for the configuration being tested.

.evergreen/config/standard.yml.erb
```diff
@@ -184,7 +184,7 @@ buildvariants:
     matrix_spec:
       stress: on
       ruby: ["ruby-2.7"]
-      mongodb-version: ["4.4", '4.2', '4.0', '3.6']
+      mongodb-version: ['5.0', '4.4', '4.2', '4.0', '3.6']
```
p-mongo, line 187: This is a good change to make, can you please put it into its own PR? You would also need to run `.evergreen/update-evergreen-configs` to update the yaml files that evergreen actually uses.

Neilshweky:
@p-mongo the creation of a client is in the the example test so I put it here. I think it should be there...

.evergreen/config/standard.yml.erb
Neilshweky: Cool will do.

p-mongo (approved):

Neilshweky:
@p-mongo before I merge, is it a problem that there are some failures related to these tests?

Neilshweky:
Error in CI on sharded cluster (Mongoid 5.0, ruby 2.5):
```
Mongo::Error::OperationFailure:
733
       [246:SnapshotUnavailable]: command failed because can not establish a snapshot :: caused by :: Unable to read from a snapshot due to pending collection catalog changes; please retry the operation. Snapshot timestamp is Timestamp(1645805826, 4). Collection minimum is Timestamp(1645805826, 5) (on localhost:27017, modern retry, attempt 1)
```

Neilshweky:
> Error in CI on sharded cluster (Mongo 5.0, ruby 2.5):
> 
> ```
> Mongo::Error::OperationFailure:
> 733
>        [246:SnapshotUnavailable]: command failed because can not establish a snapshot :: caused by :: Unable to read from a snapshot due to pending collection catalog changes; please retry the operation. Snapshot timestamp is Timestamp(1645805826, 4). Collection minimum is Timestamp(1645805826, 5) (on localhost:27017, modern retry, attempt 1)
> ```

This error is happening even on replica sets. Doing the distinct operation on each mongos does not solve this problem

spec/integration/snapshot_query_examples_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+        adoptable: true
+      )
+
+      client['dogs'].insert_one(
+        name: "Pebbles",
+        color: "Brown",
+        age: 10,
+        adoptable: true
+      )
+      run_mongos_distincts "pets", "cats"
```
p-mongo, line 36: `mongos_each_direct_client` does nothing when topology is not sharded. Can you try here

```
if ClusterConfig.instance.topology == :sharded
  run_mongos_distincts "pets", "cats"
else
  authorized_client.use('pets')['test'].distinct('foo')
  authorized_client.use('cats')['test'].distinct('foo')
end
```

Neilshweky: done!

Neilshweky:
@p-mongo I added that code and the error is still happening, any other ideas?

Neilshweky:
This looks good to me! You can merge when ready! Thanks for picking this up!

PR #2420, opened 2022-02-23
Explicitly specify python2 for ubuntu2004


p-mongo:


Neilshweky (approved):
LGTM, would you mind putting a small explanation of why this is necessary in the PR for our future selves?

p-mongo:
That is actually a good question. I believe the reason python 2 is currently used is because at some point in the past, some distros (like ubuntu1204 for example) shipped  ancient python 3 that didn't work, and to avoid tracking which python we are using we simply used python 2 everywhere. It is likely that on all of the currently used oses there is a recent enough python 3 such that the use of python 2 is no longer necessary.

p-mongo:
But, the reason for this particular PR is that newer distros (ubuntu2004) does not contain a `python` binary, requiring us to explicitly specify `python2` or `python3`.

p-mongo:
I just realized that the issue only occurs in docker, not in evergreen. An alternate fix would have been to install `python-is-python2` in docker. But I like explicitly specifying the python version more.

Neilshweky (approved):
Thanks for the explanation!

PR #2426, opened 2022-03-05
RUBY-2868 Pin to connections for load balanced

comandeo:


p-mongo:
The exception is part of the public API and needs to continue to be defined (but should be marked deprecated).

lib/mongo/error/notable.rb
```diff
@@ -70,6 +70,13 @@ def add_note(note)
       # @api experimental
       attr_accessor :service_id
 
+      # Returns global id of the connection on which the error occurred.
+      #
+      # @return [ Object | nil ] Service id.
```
p-mongo, line 75: Connection id?

lib/mongo/error/notable.rb
```diff
@@ -70,6 +70,13 @@ def add_note(note)
...
+      #
+      # @api experimental
```
p-mongo, line 77: Could be private if you don't have a use for applications to read it.

lib/mongo/server/connection_pool/generation_manager.rb
```diff
@@ -30,33 +30,32 @@ def initialize(server:)
...
-              raise ArgumentError, "Generation scoping to services is only available in load-balanced mode, but the server at #{server.address} is not a load balancer"
-            end
-          else
-            if server.load_balancer?
+        def generation(connection_global_id: nil)
+          key = if server.load_balancer?
+            if connection_global_id.nil?
               raise ArgumentError, "The server at #{server.address} is a load balancer and therefore does not have a single global generation"
             end
+            connection_global_id
```
p-mongo, line 38: If I am reading this correctly, generation must still be based on the service id: https://github.com/mongodb/specifications/blob/master/source/load-balancers/load-balancers.rst#initial-handshake-errors

spec/mongo/crypt/binding/version_spec.rb
```diff
@@ -14,7 +14,7 @@
     end
 
     it 'is in the x.y.z-tag format' do
-      expect(version).to match(/\A(\d+.){2}(\d+)?(-[A-Za-z\+\d]+)?\z/)
+      expect(version).to match(/\A(\d+.){2}(\d+)?(-*[A-Za-z\+\d]+)?\z/)
```
p-mongo, line 17: Could you explain what this change is for please?

lib/mongo/server/connection_pool/generation_manager.rb
comandeo: Wow, good catch! Thank you, I missed this moment

lib/mongo/error/notable.rb
comandeo: Done

spec/mongo/crypt/binding/version_spec.rb
comandeo: Latest version of libmongocrypt is `1.4.0alpha0+20220311gitd1f6bd4c9b`, and it does not match our regex. I reverted this change here because it is not related to this PR. I'll clarify why it happened and create a separate PR, if necessary.

p-mongo: I see. This regexp actually suggests a certain version number shape, and the description on line 16 doesn't match the version with `+` in it. I would suggest either complexifying the regexp to  something like `/\A(\d+.){2}(\d+\w*)?([-+][A-Za-z\+\d]+)?\z/` (untested) to retain the shape suggestion, perhaps removing the last `+` allowance which I don't quite understand the purpose of to `/\A(\d+.){2}(\d+\w*)?([-+][A-Za-z\d]+)?\z/`,  or simplifying this drastically to match all kinds of present and future versions.

p-mongo: Though I am also wondering whether we should be testing against unreleased versions which is what `1.4.0alpha0+20220311gitd1f6bd4c9b` looks like to me, the release should be `1.4.0alpha0` and perhaps it would be a better idea to test only against stable releases as we do with server versions (currently 1.3.1).

lib/mongo/server.rb
```diff
@@ -594,7 +592,7 @@ def update_description(description)
       @description = description
     end
 
-    # @param [ Object ] :service_id Close connections with the specified
+     # @param [ Object ] :service_id Close connections with the specified
```
p-mongo, line 595: Extra whitespace

lib/mongo/server/connection_pool.rb
```diff
@@ -307,11 +307,12 @@ def check_out(service_id: nil)
...
-                connection = next_available_connection(service_id: service_id)
+                connection = next_available_connection(
+                  connection_global_id: connection_global_id
+                )
 
-                # If service_id is not nil, connection may be nil here
+                # If connection_global_id is not nil, connection may be nil here
                 # even if there are available connections in the pool
-                # (they could be to other services).
                 break unless connection
```
p-mongo, line 317: Unlike with service id where a connection with the wanted service id might spontaneously appear in the future, this cannot happen with connection global ids. Therefore at this point I suggest checking if any connections in the pool have the wanted cgid and if none exist, raise an exception here immediately.

spec/mongo/crypt/binding/version_spec.rb
comandeo: Apparently, libmongocrypt versions should contain dash:
```
/**
 * @def MONGOCRYPT_VERSION
 * The version string describing libmongocrypt.
 * Has the form x.y.z-<pre>+<date>+git<sha>.
 */
 ```
The latest version has a malformed version, there is a bug for this - https://jira.mongodb.org/browse/MONGOCRYPT-412
I made our test pending until it is fixed.

File: lib/mongo/server/connection_pool.rb [hunk 7]
```diff
@@ -321,7 +322,9 @@ def check_out(service_id: nil)
                   next
                 end
 
-                if connection.generation != generation(service_id: connection.service_id)
+                if connection.generation != generation(
+                  service_id: connection.service_id
+                )
                   # Stale connections should be disconnected in the clear
                   # method, but if any don't, check again here
```
p-mongo, line 329: In my understanding this operation shouldn't be done if the connection is pinned. The language in https://github.com/mongodb/specifications/blob/master/source/load-balancers/load-balancers.rst#connection-pinning and subsequent sections says that connection is not returned to the pool while it is pinned, which in effect is how it avoids nuking the connection if (hopefully this is correct) an election happens while a cursor is being iterated. I think master is already somewhat deficient in this manner in that it might close a connection for a service leaving operations for that service to fail in the future, and I'm not sure whether master is already deficient for cursors.

What I think needs to happen is: 1) the connection needs to know that something is pinned to it (session or cursor), 2) while anything is pinned to connection it shouldn't be closed due to generation becoming stale, max idle time being exceeded, etc.

In theory every connection that something is pinned to should be used eventually for an operation which would unpin the connection (killcursors/aborttransaction for example), and when that operation completes and the connection is checked in it could be closed, but the unpinning logic might need to be added e.g. for the cursors.

lib/mongo/server/connection_pool.rb
```diff
@@ -495,8 +498,6 @@ def check_in(connection)
       # @option options [ true | false ] :lazy If true, do not close any of
       #   the idle connections and instead let them be closed during a
       #   subsequent check out operation.
-      # @option options [ Object ] :service_id Discard state for the specified
-      #   service id only.
```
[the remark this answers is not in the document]
comandeo: It is used, I just deleted a duplicate comment.

lib/mongo/server.rb
comandeo: Done

lib/mongo/server/connection_pool.rb
```diff
@@ -471,7 +474,7 @@ def check_in(connection)
             # Connection was closed - for example, because it experienced
             # a network error. Nothing else needs to be done here.
             @populate_semaphore.signal
-          elsif connection.generation != generation(service_id: connection.service_id)
+           elsif connection.generation != generation(service_id: connection.service_id)
```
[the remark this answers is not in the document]
comandeo: Fixed

lib/mongo/server/connection_pool.rb
```diff
@@ -341,8 +344,8 @@ def check_out(service_id: nil)
                 throw(:done)
               end
 
-              if service_id
-                # If we need a connection to a particular service, we can't
+              if @server.load_balancer? && connection_global_id
+                # If we need a  particular connection, we can't
                 # create one if we don't already have one, but we can wait
                 # for an in-progress operation to return such a connection
                 # to the pool, or for the populator to create a suitable
```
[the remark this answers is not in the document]
comandeo: Thanks, totally forgot to update comments. Fixed.

lib/mongo/server/connection_pool.rb
comandeo: Good catch, thanks. Fixed.

lib/mongo/monitoring/event/cmap/pool_cleared.rb
```diff
@@ -37,7 +37,7 @@ class PoolCleared < Base
           # Create the event.
           #
           # @param [ Address ] address
-          # @param [ Object ] service_id The service id, if any.
+          # @param [ Object ] service:service_id The service id, if any.
```
p-mongo, line 40: Typo?

lib/mongo/error/connection_not_exist.rb
```diff
@@ -0,0 +1,25 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module Mongo
+  class Error
+
+    # Exception raised when trying to check out connection that does not exist.
+    class ConnectionNotExist < Error
```
p-mongo, line 22: We have no *NotExist and no *NotExists exceptions, but we have a bunch of Missing* ones.  I think `MissingConnection` might be more in line with other exceptions and avoid the singular/plural awkwardness?

lib/mongo/error/connection_not_exist.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+# encoding: utf-8
+
+# Copyright (C) 2020 MongoDB Inc.
```
p-mongo, line 4: 2022

p-mongo (approved):

lib/mongo/error/connection_not_exist.rb
comandeo: Fixed

lib/mongo/error/connection_not_exist.rb
comandeo: Thank you, fixed.

lib/mongo/server/connection_pool.rb
comandeo: Created https://jira.mongodb.org/browse/RUBY-2936

PR #2428, opened 2022-03-09
RUBY-2820 create session not materialized error

Neilshweky:


lib/mongo/error/session_not_materialized.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+# encoding: utf-8
+
+# Copyright (C) 2020 MongoDB Inc.
```
p-mongo, line 4: 2022

lib/mongo/error/session_not_materialized.rb
```diff
@@ -0,0 +1,29 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module Mongo
+  class Error
+
+    # This exception is raised when a session is attempted to be used but the
```
p-mongo, line 21: s/ the//

spec/mongo/session_spec.rb
```diff
@@ -268,6 +268,18 @@
         end.to raise_error(Mongo::Error::SessionEnded)
       end
     end
+
+    context "when the sesion is not materialized" do
+      let(:session) { authorized_client.get_session(implicit: true) }
+
+      it "raises SessionNotMaterialized" do
+        expect(session.materialized?).to be false
```
p-mongo, line 276: Can this line be in a `before` block to separate this precondition check from the actual test please?

p-mongo (approved):

p-mongo:
The reason Mongoid CI is failing is that Mongoid has had the session registry implementation copied into it and now the Mongoid copy is outdated (and doesn't work). This PR is correct to fix the missing exception definition in the driver, however to repair Mongoid it needs to get the updated session registry implementation, which should be accomplished by moving the driver's copy into MRSS, deleting the copy that exists in Mongoid and using MRSS's version from both projects.

p-mongo (approved):

PR #2429, opened 2022-03-14
Add support of MongoDB Extended JSON to Mongo::Crypt::Handle

Seitk:
`keyId` in Auto-encryption schema map config is required to be in BSON::Binary type to communicate with mongocryptd, for projects that working with a static config file (like Mongoid) is unable to set keyId with MongoDB Extended JSON syntax like `$binary` or `$uuid`.

This PR add support on raw JSON string schema map and also hash object parsed by static mongoid.yml with Mongoid

### DIFF
--- a/lib/mongo/crypt/handle.rb
+++ b/lib/mongo/crypt/handle.rb
@@ -49,20 +49,19 @@ module Mongo
           Binding.method(:mongocrypt_destroy)
         )
 
-        @kms_tls_options =  kms_tls_options
+        @kms_tls_options = kms_tls_options
 
         @schema_map =
           case options[:schema_map].class
           when String
             begin
-              # Support raw JSON string
+              # Rebuild to raw JSON string as schema map implementation requires BSON variable instances
               BSON::ExtJSON.parse(options[:schema_map])
             rescue JSON::ParserError
               nil
             end
-          when Hash
-            # Rebuild to BSON as schema map implementation requires BSON variable instances
-            BSON::ExtJSON.parse options[:schema_map].to_json
+          else
+            options[:schema_map]
           end
 
         set_schema_map if @schema_map
--- a/lib/mongo/crypt/handle.rb
+++ b/lib/mongo/crypt/handle.rb
@@ -52,7 +52,7 @@ module Mongo
         @kms_tls_options =  kms_tls_options
 
         # Rebuild to BSON as schema map implementation requires BSON variable instances
-        @schema_map = BSON::ExtJSON.parse options[:schema_map].to_json
+        @schema_map = BSON::ExtJSON.parse(options[:schema_map].is_a?(String) ? options[:schema_map] : options[:schema_map].to_json)
 
         set_schema_map if @schema_map
 
--- a/lib/mongo/crypt/handle.rb
+++ b/lib/mongo/crypt/handle.rb
@@ -51,8 +51,19 @@ module Mongo
 
         @kms_tls_options =  kms_tls_options
 
-        # Rebuild to BSON as schema map implementation requires BSON variable instances
-        @schema_map = BSON::ExtJSON.parse(options[:schema_map].is_a?(String) ? options[:schema_map] : options[:schema_map].to_json)
+        @schema_map =
+          case options[:schema_map].class
+          when String
+            begin
+              # Support raw JSON string
+              BSON::ExtJSON.parse(options[:schema_map])
+            rescue JSON::ParserError
+              nil
+            end
+          when Hash
+            # Rebuild to BSON as schema map implementation requires BSON variable instances
+            BSON::ExtJSON.parse options[:schema_map].to_json
+          end
 
         set_schema_map if @schema_map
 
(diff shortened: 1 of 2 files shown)


Seitk:
@comandeo thanks for your suggestion! The `BSON::ExtJSON.parse` logic was added when I trying to figure out how to make it work with mongoid. That should work perfectly when we have all the configuration in `mongoid.yml`.

The gem `mongoid-client-side-encryption` is trying to provide better developer experience on setting the encryption options along with the model definition. It provides a generator to build the schema map in JSON format, and then be loaded into `mongoid.yml` as following

```
clients:
  default:
    ...
    options:
      auto_encryption_options:
        key_vault_namespace: ...
        kms_providers: ...
        schema_map: <%= File.read('config/mongodb_schema_map.json') %>
    ...
```

I found that it's quite hard to embed or load external yml in ERB, that's why I modified a bit and let the `lib/mongo/crypt/handle.rb:59` to take JSON string as well. Is there any suggestion on how the generator can set `!ruby/object:BSON::Binary` data into the yml?

Seitk:
I am thinking maybe I can discard the changes on extra `BSON::ExtJSON.parse` when `schema_map` is a `Hash`
So that might make less impact to existing test cases but at least our gem will work with JSON string instead

Seitk:
Hi @comandeo, if you don't want to change on the mongo ruby driver I might propose to build a custom class for the extra parsing work and whitelist into Mongoid `load_yaml` process

Workflow as following

(1) in our case we will pre-build the schema map into a JSON string
```ruby
{"shopline_dev.users":{"bsonType":"object","encryptMetadata":{"keyId":[{"$uuid":"2ffe54ab-2b04-4556-80ca-7f96e709b2b6"}]}}}
```

(2) I add a custom class for working with JSON string and Extended BSON

```ruby
module MongoidClientSideEncryption
  class SchemaMap
    attr_reader :data, :initialized

    def initialize(data)
      @initialized = false
      @data = data
    end

    # Proxy all method to data hash
    def method_missing(m, *args, &block)
      lazy_initialize!
      @data.send(m, *args, &block)
    end

    # Convert raw JSON string to a hash
    def lazy_initialize!
      return if @initialized
      @data = @data.to_json if @data.is_a?(Hash)
      @data = BSON::ExtJSON.parse(@data)
    rescue JSON::ParserError, ::BSON::Error::ExtJSONParseError
      @data = {}
    ensure
      @initialized = true
    end

    # Hackaround for Mongo::Crypt::Handle to use this class as Hash
    def is_a?(klass)
      @data.is_a?(klass)
    end

    def to_hash
      lazy_initialize!
      @data
    end
  end
end
```

Then setup our `mongoid.yml` like this
```yml
development:
  clients:
    default:
      ...
      options:
        ...
        auto_encryption_options:
          key_vault_namespace: 'admin.datakeys'
          kms_providers:
            ...
          schema_map: !ruby/object:MongoidClientSideEncryption::SchemaMap
            data: <%= File.read('config/mongodb_schema_map.json') %>
          ...
```

In Mongoid, we might need to allow the application to add their own whitelisted classes into the `#load_yaml` like

```ruby
def load_yaml(path, environment = nil)
  ...
  custom_permitted_classes = ...
  data = if RUBY_VERSION.start_with?("2.5")
    YAML.safe_load(ERB.new(contents).result, [Symbol], [], true)
  else
    YAML.safe_load(ERB.new(contents).result, permitted_classes: [Symbol].concat(custom_permitted_classes), aliases: true)
  end
  ...
end
```

Since we are using Mongoid v7.0.5 which it's still using `YAML.load`, we are good for this moment with the following change in our gem lol
https://github.com/shoplineapp/mongoid-client-side-encryption/pull/2/files

What do you think?

comandeo:
Hi @Seitk , I like the idea to add changes to Mongoid instead of driver for this issue. I also do like your solution with a custom class! I think we can even add it to Mongoid itself, and explicitly allow loading it in `load_yaml` method. 

@p-mongo @Neilshweky what do you think?

p-mongo:
My proposed solution is in https://github.com/mongodb/mongoid/pull/5251.

johnnyshields:
We should regard this `auto_encryption_options` config as experimental. The usage may change before the official Mongoid 8.0 release--we should discuss it further in light of how ActiveRecord does things (`encrypts` macro on model)

p-mongo:
`auto_encryption_options` are part of the driver's FLE configuration and are described in https://www.mongodb.com/docs/ruby-driver/master/reference/client-side-encryption/#automatic-encryption. They are not experimental and they are not part of Mongoid.

comandeo:
Hi @Seitk, we discussed the issue in the team, and our proposed solution is described here - https://jira.mongodb.org/browse/RUBY-2997. We are going to allow the driver to load `schemaMap` from a file; so, we will be able to generate the `schemaMap` in the `mongoid-client-side-encryption`, and then load it into the driver.

Please let us know what do you think. If it sounds reasonable to you, I'll close this PR.

Seitk:
@comandeo sounds great, thanks!

PR #2432, opened 2022-03-24
RUBY-2933 Fix getMore with comment specs

comandeo:
https://jira.mongodb.org/browse/RUBY-2912
https://jira.mongodb.org/browse/RUBY-2933
https://jira.mongodb.org/browse/RUBY-2934


lib/mongo/cursor.rb
```diff
@@ -424,12 +424,10 @@ def get_more_operation
...
         end,
-        comment: if view.respond_to?(:options) && view.options.is_a?(Hash)
-          view.options[:comment]
-        else
-          nil
-        end,
       }
+      if view.respond_to?(:options) && view.options.is_a?(Hash)
+        spec[:comment] = view.options[:comment] unless view.options[:comment].nil?
+      end
```
Neilshweky, line 430: This might be a dumb question, but why don't you check for `get_more_comment_enabled?` here as well?

Neilshweky (approved):

lib/mongo/cursor.rb
comandeo: We need a `connection` object to do the feature check, but we do not have it here, unfortunately.

Neilshweky: I see, and we don't get the connection until much later... Is this going to lead to unnecessarily sending comments when the feature is not enabled?

comandeo: It shouldn't. We check if the feature is enabled on the operation layer, when we do have a connection - https://github.com/mongodb/mongo-ruby-driver/pull/2432/files#diff-091c8fb9685f4516c8ef985fd2e1abe701fa8c45bf94c44103159f073e905649R34
I do not think there is a chance that a `comment` will be send if the feature is not enabled.

Neilshweky: okay I understand now, thanks for the explanation!

p-mongo (approved):

PR #2433, opened 2022-03-25
Fix RUBY-2932 map_reduce changes broke read preference passing from Mongoid


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2932

PR #2445, opened 2022-04-05
RUBY-2923 Fix slow spawn mongocrypt

comandeo:
https://jira.mongodb.org/browse/RUBY-2923


lib/mongo/cluster.rb
```diff
@@ -962,25 +962,11 @@ def validate_session_support!
       # No data bearing servers known - perform server selection to try to
       # get a response from at least one of them, to return an accurate
       # assessment of whether sessions are currently supported.
```
p-mongo, line 964: This comment and the docstring on top of the method need to be adjusted. Suggest something like "If the client isn't connected to any servers and doesn't find any servers for the duration of server selection timeout, this method will raise NoServerAvailable. This method is called from the operation execution flow, and if it raises NoServerAvailable the entire operation will fail with that exception, since the operation execution has waited for the server selection timeout for any server to become available (which would be a superset of the servers suitable for the operation being attempted) and none materialized."

p-mongo (approved):

lib/mongo/cluster.rb
comandeo: Updated.

PR #2447, opened 2022-04-06
RUBY-2715 Add documentation for Time Series collections

Neilshweky:
A lot of this is taken from the MongoDB docs on Time Series collections here: https://www.mongodb.com/docs/manual/core/timeseries-collections/#time-series-collections

In general I opted for less detail on exactly what everything does and more on how to do it in Ruby. If the users wants to know more about time series collections in general they should refer to the MongoDB docs. I have linked it in multiple places.


p-mongo (approved):
This is awesome!

PR #2452, opened 2022-04-14
Treat 0 limit as no limit in query caching

mikebaldry:
Related to https://jira.mongodb.org/browse/MONGOID-5285

MongoDB treats a `limit` of `0` as having no limit at all but the query caching does not take this in to consideration, returning 0 records, nor using existing cached items with a 0 limit when they could be used.

### DIFF
--- a/lib/mongo/collection/view/iterable.rb
+++ b/lib/mongo/collection/view/iterable.rb
@@ -71,6 +71,9 @@ module Mongo
             # re-use results from an earlier query with the same or larger
             # limit, and then impose the lower limit during iteration.
             limit_for_cached_query = respond_to?(:limit) ? limit : nil
+
+            # For the purposes of caching, a limit of 0 means no limit, as mongo treats it as such.
+            limit_for_cached_query = nil if limit_for_cached_query == 0
           end
 
           if block_given?
--- a/lib/mongo/query_cache.rb
+++ b/lib/mongo/query_cache.rb
@@ -180,6 +180,9 @@ module Mongo
       # @api private
       def get(**opts)
         limit = opts[:limit]
+        # For the purposes of caching, a limit of 0 means no limit, as mongo treats it as such.
+        limit = nil if limit == 0
+
         _namespace_key = namespace_key(**opts)
         _cache_key = cache_key(**opts)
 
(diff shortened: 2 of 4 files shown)


p-mongo (commented):
Very thorough, thank you for your work @mikebaldry

lib/mongo/query_cache.rb
```diff
@@ -208,6 +210,14 @@ def get(**opts)
         end
       end
 
+      def normalized_limit(limit)
```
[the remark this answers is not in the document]
p-mongo: Though a limit doesn't require creation of a cursor.

I would prefer collection to not reference QueryCache for implementation parts.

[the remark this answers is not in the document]
mikebaldry: I did consider CachingCursor exposing a method to get a limited set of values and handling the normalization internally there, some how. Didn't get that far though, happy to attempt it though

lib/mongo/collection/view/iterable.rb
```diff
@@ -70,7 +70,7 @@ def each
             # If a query with a limit is performed, the query cache will
             # re-use results from an earlier query with the same or larger
             # limit, and then impose the lower limit during iteration.
-            limit_for_cached_query = respond_to?(:limit) ? limit : nil
+            limit_for_cached_query = respond_to?(:limit) ? QueryCache.normalized_limit(limit) : nil
           end
 
           if block_given?
```
mikebaldry, line 76: Can't comment on line 89 because it's not in the diff but I'll ask on this one..

If we call `each` without a block, it'll just return an `Enumerable`, which won't have the cached limit applied. 

This case probably should be handled by always doing the positive case, and then calling `to_enum` on the result of it

mikebaldry: line 79 could end up being something like 
```ruby
result = if use_query_cache?
  @cursor.limited_to(limit) # is there a case where limit doesn't exist? CachingCursor inherits Cursor which implements limit, so don't see why we have the `respond_to?(:limit)` check
else
  @cursor
end

if block_given? 
  # iterate result and yield 
else
  result.to_enum
end
```

Just ideas at them moment, let me know what you think

mikebaldry: cc @p-mongo (not sure if you'd normally get notifications for these comments on someone elses PR)

p-mongo: I do @mikebaldry , but thank you for the highlight. We'll discuss this PR with the team to figure out the best course of action.

comandeo:
The issue in now under the driver project - https://jira.mongodb.org/browse/RUBY-2961

comandeo:
The issue in now under the driver project - https://jira.mongodb.org/browse/RUBY-2961

comandeo (approved):

PR #2453, opened 2022-04-14
RUBY-2957 Use release version of libmongocrypt


p-mongo:


Neilshweky (approved):
I'm not going to pretend to know what all of this means/does but I read all of it and nothing jumps out to me as incorrect. LGTM!

comandeo (approved):

PR #2455, opened 2022-04-15
RUBY-2956 Make allowDiskUse opt-out rather  than opt-in in 6.0

Neilshweky:
RUBY-2956

### DIFF
--- a/spec/runners/unified/crud_operations.rb
+++ b/spec/runners/unified/crud_operations.rb
@@ -244,6 +244,9 @@ module Unified
       if batch_size = args.use('batchSize')
         opts[:batch_size] = batch_size
       end
+      if !(allow_disk_use = args.use('allowDiskUse')).nil?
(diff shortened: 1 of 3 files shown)


spec/runners/unified/crud_operations.rb
```diff
@@ -244,6 +244,9 @@ def aggregate(op)
       if batch_size = args.use('batchSize')
         opts[:batch_size] = batch_size
       end
+      if !(allow_disk_use = args.use('allowDiskUse')).nil?
```
Neilshweky, line 247: When this option was false, `allow_disk_use = args.use('allowDiskUse')` would evaluate to false and the code inside it would not execute. I had to get a little creative here.

p-mongo: Another way that hopefully works with whatever class `args` is should be `if args.key?('allowDiskUse')`.

p-mongo (approved):

PR #2459, opened 2022-04-20
RUBY-2945 Add AWS Lambda examples

comandeo:


spec/integration/aws_lambda_examples_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      # Use the client to return the name of the connected database.
+      client.database.name
+    end
+
+    # End AWS Lambda Example 1
+
+    client.close
+  end
+
+  it "to the deployment using AWS IAM authentication" do
```
Neilshweky, line 30: I'm not sure what this test string actually means.

Neilshweky (approved):
LGTM awesome job!

spec/integration/aws_lambda_examples_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+require "spec_helper"
+
+describe "AWS Lambda examples in Ruby" do
+  require_aws_auth
+
+  it "shares the client" do
+    # Start AWS Lambda Example 1
+
+    # Import the dependency.
```
p-mongo, line 12: I  would say something like "require the driver library".

spec/integration/aws_lambda_examples_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    require "mongo"
+
+    # Create a Mongo::Client instance.
+    # CRITICAL: You must create client outside the handler so that the client
+    # can be reused across function invocations.
+    client = Mongo::Client.new(ENV["MONGODB_URI"])
+
+    def lambda_handler(event:, context:)
+      # Use the client to return the name of the connected database.
```
p-mongo, line 21: Suggest "configured database", since this will return regardless of whether the connections succeed (and even when the client is constructed with an empty host list).

p-mongo: Or actually perform an operation which would perhaps be more meaningful.

spec/integration/aws_lambda_examples_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    # Import the dependency.
+    require "mongo"
+
+    # Create a Mongo::Client instance using AWS IAM authentication.
+    # CRITICAL: You must create client outside the handler so that the client
+    # can be reused across function invocations.
+    client = Mongo::Client.new([ENV["MONGODB_HOST"]],
+                               auth_mech: :aws,
+                               user: ENV["AWS_ACCESS_KEY_ID"],
```
p-mongo, line 48: `ENV.fetch(...)`  would also  be good here I would say.

spec/integration/aws_lambda_examples_spec.rb
p-mongo: I can't find where these examples are actually specified but I'm thinking this should be a `context` with something like "when using AWS IAM authentication" with the example under `it` under the context.

p-mongo (approved):

spec/integration/aws_lambda_examples_spec.rb
comandeo: The examples are specified here - https://www.mongodb.com/docs/atlas/manage-connections-aws-lambda/
and in the scope document:

> The scope of these code examples is limited to the instructions on:
> * How to share the client (“Example 1”. See first snippet in [this](https://www.mongodb.com/docs/atlas/manage-connections-aws-lambda/#connection-examples) section for structure)
> * How to connect to the deployment using AWS IAM authentication (“Example 2”. Same structure as “Example 1”, but see the blue note box in the [docs](https://www.mongodb.com/docs/atlas/manage-connections-aws-lambda) for which lines need to be replaced to work with AWS IAM auth)

PR #2463, opened 2022-04-23
MONGOID-4611: [Proof-of-Concept] Option to convert $set: nil to $unset when writing

johnnyshields:
This is a proof-of-concept to add a mode to the driver where `$set: { name: nil }` is replaced with `$unset: { name: true }`. The former persists the document key with value "BSON Null".

The reasons are:
- It makes your database more compact if you have many nil values, as I often do.
- "BSON Null" does not work with sparse indexes (or partial_filter_expression `$exists` operator); it counts as "existing" in the database.
- Mongoid treats "BSON Null" and key-not-present identically when deserializing (demongoizing) from database.

Note users in this mode would never want to use "BSON Null" type, however, if you ever did want it, you could still do `$set: { name: BSON::NilClass }` and it would persist it.

The ticket was originally marked for Mongoid, however, after investigating I thought it was better to handle at the driver level to ensure uniformity, and then have Mongoid pass-in its global config value (`Mongoid.unset_nil_values`) in the driver options. Otherwise, we'll end up doing the `$set` --> `$unset` conversion in many more places in Mongoid, including things like `delayed_atomic_sets`, etc. it becomes quite a mess.

Please take a look and let me know what you think.

### DIFF
--- a/lib/mongo/collection/view/writable.rb
+++ b/lib/mongo/collection/view/writable.rb
@@ -155,6 +155,8 @@ module Mongo
         #   See the server documentation for details.
         # @option options [ Object ] :comment A user-provided
         #   comment to attach to this command.
+        # @option options [ Boolean ] :unset_nil_values Converts $set operations
+        #   with nil values to $unset.
         #
         # @return [ BSON::Document ] The document.
         #
@@ -171,6 +173,8 @@ module Mongo
             end
 
             QueryCache.clear_namespace(collection.namespace)
+
+            convert_set_nil_to_unset(spec) if opts.delete(:unset_nil_values)
 
             cmd = {
               findAndModify: collection.name,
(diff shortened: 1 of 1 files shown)


p-mongo:
Hi @johnnyshields , thank you for the PR.

Can you please add a code snippet showing how a user would use this functionality?

johnnyshields:
@p-mongo I am planning a PR for Mongoid to enable this. That PR is *not yet raised*, but it will look like:

```yml
# in config/mongoid.yml
  options:
    unset_nil_values: true
```

Then, when I do:

```ruby
class Person
  include Mongoid::Document

  field :name, type: String
  field :hometown, type: String
end

person = Person.all.first
person.update_attributes!({ name: "Bob", hometown: nil })
```

Internally, Mongoid calls the Ruby driver's `update_one` method (I think). Mongoid will pass-in the `unset_nil_values` option to the driver.

The driver will then process this:

```
# **without** unset_nil_values
{ "$set" => { name: "Bob", hometown: nil } }

# with unset_nil_values
{ "$set" => { name: "Bob" }, "$unset" => { hometown: true } }
```

The result of this is that the `hometown` field will not have a key set in the database. (Otherwise, it would be set to `BSON Null` type.)

Please let me know if this makes sense.

p-mongo:
Per the discussion in the ticket, we decided that the driver should not perform query transformation and that Mongoid should generate the $unsets based on changes in models.

PR #2465, opened 2022-04-27
RUBY-2973 Add the acknowledged? field to the BulkWrite::Result class

Neilshweky:
All of the other Result classes descend form Operation::Result super class and therefore already have the acknowledged? method. This PR only adds it to the BulkWrite Result class.

### DIFF
--- a/lib/mongo/bulk_write/result.rb
+++ b/lib/mongo/bulk_write/result.rb
@@ -23,6 +23,11 @@ module Mongo
     # @since 2.0.6
     class Result
 
+      # @return [ Boolean ] Is the result acknowledged?
+      def acknowledged?
+        @acknowledged
+      end
+
       # Constant for number removed.
       #
       # @since 2.1.0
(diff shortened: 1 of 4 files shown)


lib/mongo/bulk_write/result.rb
```diff
@@ -96,8 +101,11 @@ def deleted_count
       # @param [ BSON::Document, Hash ] results The results document.
       #
       # @since 2.1.0
-      def initialize(results)
+      #
+      # @api private
+      def initialize(results, acknowledged)
```
p-mongo, line 106: `acknowledged` needs to be documented in the docstring.

spec/mongo/bulk_write_spec.rb
```diff
@@ -2390,4 +2390,42 @@
       end
     end
   end
+
+  describe ".acknowledged?" do
```
p-mongo, line 2394: `#acknowledged?` since it's an instance method, `.acknowledged?` would be a class method.

p-mongo (approved):

spec/mongo/bulk_write_spec.rb
Neilshweky: done

lib/mongo/bulk_write/result.rb
Neilshweky: done

PR #2468, opened 2022-04-28
RUBY-2935 Remove use of admin database FLE docs

comandeo:


p-mongo (approved):
I see the requirement to use `__keyVault` but this is ugly and not Ruby style, I would just use `keyvault`.

PR #2477, opened 2022-05-04
RUBY-2583 remove v2 test calls

Neilshweky:


spec/spec_tests/crud_spec.rb
```diff
@@ -5,18 +5,8 @@
 
 require 'runners/crud'
 
-describe 'CRUD v1 spec tests' do
+describe 'CRUD spec tests' do
```
p-mongo, line 8: The `v1` is meaningful here, I suggest keeping it.

Neilshweky: done

p-mongo (approved):

comandeo (approved):

PR #2479, opened 2022-05-05
RUBY-2736 Add server connectionId to command monitoring events

Neilshweky:


comandeo (approved):

lib/mongo/monitoring/event/command_failed.rb
```diff
@@ -53,6 +53,9 @@ class CommandFailed < Mongo::Event::Base
         # @return [ Integer ] request_id The request id.
         attr_reader :request_id
 
+        # @api private
```
p-mongo, line 56: This actually needs to be public because this is a cross-driver feature delivered for users.

Neilshweky: done

PR #2481, opened 2022-05-06
RUBY-2978 Move end session to delayed finalizer

comandeo:


spec/mongo/cursor_spec.rb
```diff
@@ -353,7 +353,7 @@
         cluster.instance_variable_get(:@periodic_executor).flush
         expect do
           cursor.to_a
-        end.to raise_exception(Mongo::Error::OperationFailure, /[cC]ursor.*not found/)
+        end.to raise_exception(Mongo::Error::SessionEnded)
```
p-mongo, line 356: This exception shouldn't be raised to users, right? If so I suppose this test is OK with a note explaining why this exact exception is raised (cursor cleanup triggers session cleanup, subsequent operation with this session fails).

If this is triggerable by users I would say cursor should track its state and raise a better error prior to attempting an operation on closed session.

lib/mongo/cluster/reapers/cursor_reaper.rb
```diff
@@ -195,6 +195,12 @@ def kill_cursors
             connection_global_id: kill_spec.connection_global_id,
           }
           op.execute(server, context: Operation::Context.new(options: options))
+
+          if session = kill_spec.session
+            if session.implicit? && !session.ended?
```
p-mongo, line 200: Is this cleanup now happening earlier compared to when it happened before?

spec/mongo/cursor_spec.rb
comandeo: This exception won't be raised to users. Code that causes is is scheduled in finalizer, so the cursor is gone when the code is executed.

lib/mongo/cluster/reapers/cursor_reaper.rb
comandeo: I wouldn't say so. First, before it never happened because it tried to use a mutex in a finalizer. Second, this is now called in a job that is done by a periodic executor, so it is most probably delayed. And finally, it is the last operation in the killing cursor routine.

Neilshweky (approved):

lib/mongo/cluster/reapers/cursor_reaper.rb
comandeo: Done

lib/mongo/cluster/reapers/cursor_reaper.rb
```diff
@@ -197,7 +197,10 @@ def kill_cursors
           op.execute(server, context: Operation::Context.new(options: options))
 
           if session = kill_spec.session
-            if session.implicit? && !session.ended?
+            if session.ended?
+              raise Error::LintError, "Ended session for a garbage collected cursor"
```
p-mongo, line 201: This actually needs to be conditioned on linting being enabled (`if Lint.enabled?`)

PR #2485, opened 2022-05-09
RUBY-2480 to_bson getting called 6 times on a collection insert

Neilshweky:
This PR reduces the number of times that `to_bson` is called to two. The first time that `to_bson` is called, it's results are cached, however if it is not called with `validating_keys` set to true, when it is called with validations, an error will not be raised since we won't actually call `to_bson`. Therefore if `validating_keys` is set to true, and it has not already been validated, trigger the calculation of `to_bson`.

After allowing dots and dollars in Ruby, `to_bson` is only called once as there is no longer any validation.


p-mongo:
I changed the module to a class which I think is easier to reason about with respect to 1) whether the driver modifies user-provided objects, and 2) whether the data can be mutated while it's cached (there are many ways to modify a hash besides `[]=`).

spec/mongo/protocol/caching_hash_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+          let(:validating_keys) { b }
+
+          it "is false" do
+            expect(needs_validation?).to be false
+          end
+        end
+      end
+    end
+  end
+end
```
Neilshweky, line 82:
```suggestion
end

```

Neilshweky:
I'm a little confused... I remember some of the documents being modified after the fact and therefore I had to clear the cache so the bytes would be recalculated... How does that happen in your changes?

p-mongo (approved):

PR #2487, opened 2022-05-10
RUBY-2971 Add createCollection and collMod spec tests for changeStreamPreAndPostImages option

Neilshweky:
RUBY-2971. Did not add the collMod spec tests since we don't have a modify collections helper function.

EDIT: Must review this PR first: https://github.com/mongodb/mongo-ruby-driver/pull/2488

I was able to add the modifyCollections test back after skipping unsupported operations.

### DIFF
--- a/lib/mongo/collection.rb
+++ b/lib/mongo/collection.rb
@@ -58,11 +58,12 @@ module Mongo
     # @since 2.1.0
     CHANGEABLE_OPTIONS = [ :read, :read_concern, :write, :write_concern ].freeze
 
-    # Options that can be used for creating a time-series collection.
-    TIME_SERIES_OPTIONS = {
+    # Options map to transform create collection options.
+    CREATE_COLLECTION_OPTIONS = {
       :time_series => :timeseries,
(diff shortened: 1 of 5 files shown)


lib/mongo/collection.rb
```diff
@@ -58,11 +58,12 @@ class Collection
     # @since 2.1.0
     CHANGEABLE_OPTIONS = [ :read, :read_concern, :write, :write_concern ].freeze
 
-    # Options that can be used for creating a time-series collection.
-    TIME_SERIES_OPTIONS = {
+    # Options map to transform create collection options.
+    CREATE_COLLECTION_OPTIONS = {
```
p-mongo, line 62: This is technically an API change.

p-mongo (approved):
Approved with the constant made api private + changes in the other PR for skipping.

lib/mongo/collection.rb
Neilshweky: done

comandeo (approved):

p-mongo (approved):

PR #2489, opened 2022-05-11
RUBY-2528 Permit inserting dollar-prefixed or dotted keys

Neilshweky:
RUBY-2528

I decided not to completely remove the `validating_keys` functionality across the code base in case we want to use that in the future, and since some of the methods it's used in are not api private. I just set it to false everywhere. 

I am going to create a ticket for implementing replaceOne and update validation as specified here: 
https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#update-vs-replace-validation

The following files need to be added back: 
- updateOne-validation.yml
- updateMany-validation.yml
- replaceOne-validation.yml
- bulkWrite-update-validation.yml
- transactions/errors-client.yml (the newest version)
- First test in poc-transactions


spec/runners/unified/assertions.rb
```diff
@@ -51,6 +51,31 @@ def assert_result_matches(actual, expected)
             end
           end
 
+          %w(acknowledged).each do |k|
```
Neilshweky, line 54: I decided to move all comparisons with result object to this method. This is not necessarily applicable to dots and dollars.

spec/mongo/collection_crud_spec.rb
```diff
@@ -720,10 +720,10 @@ def generate
         { 'testing.test' => 'value' }
       end
 
-      it 'raises a BSON::String::IllegalKey exception' do
+      it 'does not raise a BSON::String::IllegalKey exception' do
         expect {
           authorized_collection.insert_one(doc)
-        }.to raise_exception(BSON::String::IllegalKey)
+        }.to_not raise_exception(BSON::String::IllegalKey)
```
p-mongo, line 726: This should verify the document actually got inserted and with the correct keys & values.

Neilshweky: done

p-mongo (approved):

comandeo (approved):

PR #2492, opened 2022-05-12
RUBY-2980 Add FLE2 api to client encryption

comandeo:


p-mongo (approved):

p-mongo (approved):

lib/mongo/crypt/binding.rb
```diff
@@ -78,6 +78,13 @@ class Binding
         end
       end
 
+      # Minimum version of libmongocrypt required by this version of the driver.
+      # An attempt to use the driver with any previous version of libmongocrypt
+      # will cause a `LoadError`.
+      #
+      # @api private
+      MIN_LIBMONGOCRYPT_VERSION = Gem::Version.new("1.5.0.pre.alpha0")
```
p-mongo, line 86: How does `1.5.0.alpha1` compare to `1.5.0.pre.alpha0`? Maybe this should be `1.5.0.alpha`.

lib/mongo/crypt/binding.rb
```diff
@@ -87,6 +94,11 @@ class Binding
       #   @return [ String ] A version string for libmongocrypt.
       attach_function :mongocrypt_version, [:pointer], :string
 
+      if Gem::Version.new(mongocrypt_version(nil)) < MIN_LIBMONGOCRYPT_VERSION
```
p-mongo, line 97: Can this be something like

```
if (actual_version = Gem::Version.new(mongocrypt_version(nil))) < MIN_LIBMONGOCRYPT_VERSION

```

to avoid copy-pasting the entire expression in the exception message below please.

lib/mongo/crypt/binding.rb
```diff
@@ -87,6 +94,11 @@ class Binding
...
+        raise LoadError, "libmongocrypt version #{MIN_LIBMONGOCRYPT_VERSION} or above is reauired, " +
```
p-mongo, line 98: required

p-mongo (approved):
The lmc version requirement needs to be in the documentation somewhere, either in the FLE page or in the compat page with the other matrices.

lib/mongo/crypt/binding.rb
comandeo: Done

lib/mongo/crypt/binding.rb
comandeo: Done

lib/mongo/crypt/binding.rb
comandeo: Done

comandeo:
> The lmc version requirement needs to be in the documentation somewhere, either in the FLE page or in the compat page with the other matrices.

There is a dedicated ticket for docs, I'd do all docs at once.

Neilshweky (approved):
LGTM!

comandeo:
CI failures seem to be irrelevant to this branch.

PR #2495, opened 2022-05-16
RUBY-2996 Implement update and replace validation

Neilshweky:
The CI failures all look unrelated and additionally they come up on other PRs.

### DIFF
--- a/lib/mongo/bulk_write.rb
+++ b/lib/mongo/bulk_write.rb
@@ -315,11 +315,19 @@ module Mongo
       end
     end
 
+    # Checks the request documents to make sure they are valid. For update
+    # documents this means they only contain atomic modifiers, and for
+    # replacement documents this means they don't contain modifiers.
+    # Note that as per the spec, we only have to examine the first element
+    # in the update document.
+    #
+    # @raise [ Error::InvalidUpdateDocument, Error::InvalidReplacementDocument ]
+    #   if the document is invalid.
     def validate_requests!
       @requests.each do |req|
         if op = req.keys.first
           if [:update_one, :update_many].include?(op)
-            if doc = wrap(req.dig(op, :update))&.first
+            if doc = maybe_first(req.dig(op, :update))
               if key = doc.keys&.first
                 unless key.start_with?("$")
                   raise Error::InvalidUpdateDocument.new(key)
@@ -337,8 +345,14 @@ module Mongo
       end
     end
 
-    def wrap(elt)
-      elt.is_a?(Array) ? elt : [elt]
+    # If the given object is an array return the first element, otherwise
+    # return the given object.
+    #
+    # @param [ Object ] obj The given object.
+    #
+    # @return [ Object ] The first element of the array or the given object.
+    def maybe_first(obj)
+      obj.is_a?(Array) ? obj.first : obj
     end
   end
 end
--- a/lib/mongo/bulk_write.rb
+++ b/lib/mongo/bulk_write.rb
@@ -329,14 +329,14 @@ module Mongo
           if [:update_one, :update_many].include?(op)
             if doc = maybe_first(req.dig(op, :update))
               if key = doc.keys&.first
-                unless key.start_with?("$")
+                unless key.to_s.start_with?("$")
                   raise Error::InvalidUpdateDocument.new(key)
                 end
               end
             end
           elsif op == :replace_one
             if key = req.dig(op, :replacement)&.keys&.first
-              if key.start_with?("$")
+              if key.to_s.start_with?("$")
                 raise Error::InvalidReplacementDocument.new(key)
               end
             end
--- a/lib/mongo/bulk_write.rb
+++ b/lib/mongo/bulk_write.rb
@@ -315,11 +315,14 @@ module Mongo
       end
     end
 
-    # Checks the request documents to make sure they are valid. For update
-    # documents this means they only contain atomic modifiers, and for
-    # replacement documents this means they don't contain modifiers.
-    # Note that as per the spec, we only have to examine the first element
-    # in the update document.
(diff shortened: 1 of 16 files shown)


lib/mongo/error/invalid_replacement_document.rb
```diff
@@ -19,23 +19,15 @@ module Mongo
...
 
     # Exception raised if the object is not a valid replacement document.
-    #
-    # @since 2.0.0
     class InvalidReplacementDocument < Error
 
-      # The error message.
-      #
-      # @since 2.0.0
-      MESSAGE = 'Invalid replacement document provided'.freeze
```
Neilshweky, line 29: I am aware that this is technically in the public API but it shouldn't be so I don't have a problem removing it.

lib/mongo/bulk_write.rb
```diff
@@ -313,5 +314,45 @@ def can_hint?(connection)
...
+    end
+
+    # If the given object is an array return the first element, otherwise
+    # return the given object.
+    #
+    # @param [ Object ] obj The given object.
+    #
+    # @return [ Object ] The first element of the array or the given object.
+    def maybe_first(obj)
+      obj.is_a?(Array) ? obj.first : obj
```
comandeo, line 355: I think this can be replaced with `Array(obj).first`

lib/mongo/collection/view/writable.rb
```diff
@@ -538,6 +541,44 @@ def update_one(spec, opts = {})
...
+        # Checks the update documents to make sure they only have atomic modifiers.
+        # Note that as per the spec, we only have to examine the first element
+        # in the update document.
+        #
+        # @param [ Hash | Array<Hash> ] spec The update document or pipeline.
+        #
+        # @raise [ Error::InvalidUpdateDocument ] if the first key in the
+        #   document does not start with a $.
+        def validate_update_documents!(spec)
+          if update = spec.is_a?(Array) ? spec&.first : spec
```
comandeo, line 556: Here we could also use `if update = Array(spec).first`

lib/mongo/error/invalid_replacement_document.rb
comandeo: Am I right that this exception was never raised before this PR? If yes, I think we can remove this constant.

comandeo (approved):

lib/mongo/error/invalid_replacement_document.rb
p-mongo: I personally would leave it with a `@deprecated` note, since it doesn't appear to be in the way.

lib/mongo/collection/view/writable.rb
Neilshweky: see my comment above

lib/mongo/error/invalid_replacement_document.rb
Neilshweky: okay great

lib/mongo/error/invalid_replacement_document.rb
```diff
@@ -19,23 +19,20 @@ module Mongo
...
       # Instantiate the new exception.
       #
-      # @example Instantiate the exception.
-      #   Mongo::Error::InvalidReplacementDoc.new
-      #
-      # @since 2.0.0
-      def initialize
-        super(MESSAGE)
+      # @param [ String ] key The invalid key.
+      def initialize(key)
```
p-mongo, line 32: This is an API change, the key should be defaulted to nil and my preference would be for it to be a keyword argument.

lib/mongo/error/invalid_update_document.rb
```diff
@@ -25,17 +25,16 @@ class InvalidUpdateDocument < Error
...
       # Instantiate the new exception.
       #
-      # @example Instantiate the exception.
-      #   Mongo::Error::InvalidUpdateDocument.new
-      #
-      # @since 2.0.0
-      def initialize
-        super(MESSAGE)
+      # @param [ String ] key The invalid key.
+      def initialize(key)
```
p-mongo, line 34: Same for this added constructor argument.

lib/mongo/bulk_write.rb
```diff
@@ -313,5 +314,45 @@ def can_hint?(connection)
         end
       end
     end
+
+    # Checks the request documents to make sure they are valid. For update
```
p-mongo, line 318: This description isn't matching the implementation, because the documents are only partially checked. A more accurate description would be something like:

> Perform the request document validation required by driver specifications. This method validates the first key of each update request document to be an operator (i.e. start with `$`) and the first key of each replacement document to not be an operator (i.e. not start with `$`). The request document may be invalid without this method flagging it as such (for example an update or replacement document containing some keys which are operators and some which are not), in which case the driver expects the server to fail the operation with an error.

p-mongo (approved):

lib/mongo/error/invalid_replacement_document.rb
Neilshweky: done

lib/mongo/error/invalid_update_document.rb
Neilshweky: done

lib/mongo/bulk_write.rb
Neilshweky: done

lib/mongo/error/invalid_replacement_document.rb
```diff
@@ -29,7 +29,7 @@ class InvalidReplacementDocument < Error
       # Instantiate the new exception.
       #
       # @param [ String ] key The invalid key.
```
Neilshweky, line 31:
```suggestion
      # @param [ String ] :key The invalid key.
```

PR #2498, opened 2022-05-19
RUBY-2968 Create and drop FLE2 collections

comandeo:


lib/mongo/operation/create/op_msg.rb
```diff
@@ -31,7 +31,19 @@ class OpMsg < OpMsgBase
...
           # Collation is always supported on 3.6+ servers that would use OP_MSG.
-          spec[:selector].merge(collation: spec[:collation]).compact
+          spec[:selector].merge(
+            collation: spec[:collation],
+            encryptedFields: spec[:encrypted_fields],
+          ).compact.tap do |sel|
+            if sel[:encryptedFields] && sel[:encryptedFields].key?(:fields)
+              sel[:encryptedFields][:fields] = sel[:encryptedFields][:fields].map do |field|
+                if field[:queries] && field[:queries].key?(:contention)
+                  field[:queries][:contention] = BSON::Int64.new(field[:queries][:contention])
```
p-mongo, line 41: This ultimately mutates user input, doesn't it?

p-mongo: If `spec` is a BSON::Document then perhaps it doesn't, but I think there should be a test for this.

comandeo: Added the test

spec/mongo/operation/create/op_msg_spec.rb
```diff
@@ -0,0 +1,285 @@
...
+    end
+  end
+
+  describe '#selector' do
+    min_server_fcv '3.6'
+
+    it 'does not mutate user input' do
+      user_input = spec.dup
+      described_class.new(user_input).send(:selector, connection)
+      expect(user_input).to eq(spec)
```
p-mongo, line 81: This does not catch deep mutation. We have https://github.com/dkubb/ice_nine attached to the test suite, so  I think what would work here is

```
user_input = IceNine.deep_freeze(spec.dup)
```

and if this is modified I expect an exception to be raised.

comandeo: Thanks, good point! Changed the test as suggested. Seems that we are fine, no exception raised.

comandeo: Good point, thank you! Updated the test.

p-mongo (approved):

Neilshweky (approved):

lib/mongo/client.rb
```diff
@@ -1141,6 +1141,14 @@ def canonicalize_ruby_options(options)
       end
     end
 
+    # Returns encrypted field map has if provided when creating the client.
```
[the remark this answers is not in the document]
comandeo: Fixed, thanks

lib/mongo/collection/queryable_encryption.rb
```diff
@@ -0,0 +1,122 @@
...
+            session: session,
+          )
+          do_drop(operation, session, context)
+        end
+        yield
+      end
+
+      private
+
+      # Checks is names for auxiliary collections are set and returns them,
```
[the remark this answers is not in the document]
comandeo: Fixed, thanks

PR #2508, opened 2022-05-25
RUBY-2947 Update distros to ubuntu1804/ubuntu2004 where possible


p-mongo:
https://jira.mongodb.com/browse/RUBY-2947


comandeo (approved):
:+1:

PR #2515, opened 2022-05-27
RUBY-2996 add feature flag for update/replace

Neilshweky:


lib/mongo/bulk_write.rb
```diff
@@ -327,20 +327,22 @@ def can_hint?(connection)
...
     #   if the document is invalid.
     def validate_requests!
-      @requests.each do |req|
-        if op = req.keys.first
-          if [:update_one, :update_many].include?(op)
-            if doc = maybe_first(req.dig(op, :update))
-              if key = doc.keys&.first
-                unless key.to_s.start_with?("$")
-                  raise Error::InvalidUpdateDocument.new(key: key)
+      if Mongo.validate_update_replace
```
p-mongo, line 330: I would return if the value is false to keep the nesting in the rest of the method down.

lib/mongo/config.rb
```diff
@@ -0,0 +1,31 @@
...
+    option :validate_update_replace, default: false
+
+    # Set the configuration options.
+    #
+    # @example Set the options.
+    #   config.options = { validate_update_replace: true }
+    #
+    # @param [ Hash ] options The configuration options.
+    def options=(options)
+      if options
```
p-mongo, line 23: The argument is defined to be a hash, making this check unnecessary. If you would like to keep it I would prefer ArgumentError raised if the argument is not a hash.

lib/mongo/config.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require "mongo/config/options"
+require "mongo/config/validators/option"
+
+module Mongo
+
+  # This module defines configuration options for Mongo.
+  module Config
```
p-mongo, line 9: This should be `@api private` I think.

lib/mongo/config/validators/option.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Mongo
+  module Config
+    module Validators
+
+      # Validator for configuration options.
+      module Option
```
p-mongo, line 8: `@api private` ?

lib/mongo/bulk_write.rb
p-mongo: Actually, shouldn't this produce a warning if the flag is off? And I think in this case the warning would need to be throttled otherwise it can spam the logs.

lib/mongo.rb
```diff
@@ -75,8 +75,16 @@
 require 'mongo/version'
 require 'mongo/write_concern'
 require 'mongo/utils'
+require 'mongo/config'
 
 module Mongo
+  extend Forwardable
+  extend self
```
p-mongo, line 82: Instead of `extend self` I prefer `class << self` syntax so that the methods are defined in the one scope where they are supposed to be used from instead of two scopes.

lib/mongo.rb
```diff
@@ -75,8 +75,16 @@
...
+
+  # Take all the public instance methods from the Config singleton and allow
+  # them to be accessed through the Mongo module directly.
+  def_delegators Config, *(Config.public_instance_methods(false))
```
p-mongo, line 86: I would sleep better if this was an explicit enumeration of methods, how difficult would that be to implement?

comandeo (commented):
I think we should add documentation explaining how to set feature flags.

lib/mongo/config.rb
Neilshweky: done

lib/mongo.rb
Neilshweky: done

lib/mongo/config/validators/option.rb
Neilshweky: done

lib/mongo/error/invalid_config_option.rb
```diff
@@ -0,0 +1,18 @@
...
+  class Error
+
+    # This error is raised when a bad configuration option is attempted to be
+    # set.
+    class InvalidConfigOption < Error
+
+      # Create the new error.
+      #
+      # @param [ Symbol, String ] name The attempted config option name.
+      def initialize(name)
```
p-mongo, line 13: This should be `@api private`.

p-mongo (approved):
The flag still needs user documentation but this PR is good to go in my opinion.

lib/mongo/bulk_write.rb
p-mongo: Mongoid does https://github.com/mongodb/mongoid/blob/master/lib/mongoid/query_cache.rb#L82, the flag can perhaps be stored in a new module or in `Config` if everything related to it is marked `@api private`.

lib/mongo.rb
```diff
@@ -75,8 +75,29 @@
...
+  class << self
+    extend Forwardable
+
+    # Delegate the given option along with its = and ? methods to the given
+    # object.
+    #
+    # @param [ Object ] obj The object to delegate to.
+    # @param [ Symbol ] opt The method to delegate.
+    def self.delegate_option(obj, opt)
+      o = opt.to_s
```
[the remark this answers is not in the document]
Neilshweky: done

lib/mongo.rb
```diff
@@ -75,8 +75,29 @@
...
+      def_delegators obj, opt, :"#{o}=", :"#{o}?"
```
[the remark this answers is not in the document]
Neilshweky: done

lib/mongo/error/invalid_config_option.rb
Neilshweky: done

Neilshweky:
Here is what the feature flags docs looks like... I put it at the bottom of `create-client.txt` because that was the closest thing we had to "configuration" docs:
Screen Shot 2022-06-01 at 12 20 01 PM

p-mongo (approved):

comandeo (approved):

PR #2520, opened 2022-06-03
RUBY-3016 Retry BSON size limits tests

Neilshweky:


p-mongo (approved):
Could you reference https://jira.mongodb.org/browse/RUBY-3016 in a comment above the added `retry_test` line and make sure the ticket # makes it into the squashed commit?

PR #2522, opened 2022-06-07
RUBY-2986 Add queryable encryption docs

comandeo:


docs/reference/client-side-encryption.txt
```diff
@@ -324,6 +324,203 @@ master key and create data keys, see the following sections of this tutorial:
...
+
+  kms_providers = {
+    local: {
+      key: local_master_key
+    }
+  }
+
+  # The key vault client is a Mongo::Client instance connected to the collection
+  # that will store your data keys.
+  key_vault_client = Mongo::Client.new(['localhost:27017'])
```
p-mongo, line 360: This should specify the `keyvault` database I think as the default is `admin`.

p-mongo: Or, if this does not matter, remove "connected to the collection" from the above text.

docs/reference/client-side-encryption.txt
```diff
@@ -324,6 +324,203 @@ master key and create data keys, see the following sections of this tutorial:
...
+            queries: {
+              queryType: 'equality'
+            }
+          }
+        ]
+      }
+    }
+
+  # Configure the client for automatic encryption
+  client = Mongo::Client.new(
```
p-mongo, line 393: I suggest specifying a database here also.

docs/reference/client-side-encryption.txt
```diff
@@ -324,6 +324,203 @@ master key and create data keys, see the following sections of this tutorial:
...
+    ['localhost:27017'],
+    auto_encryption_options: {
+      key_vault_namespace: 'encryption.__keyVault',
+      kms_providers: kms_providers,
+      encrypted_fields_map: encrypted_fields_map,
+    }
+  )
+
+  # Make sure there is no data in the collection.
+  client.use('encryption_db').database.drop
```
p-mongo, line 403: ... instead of giving it repeatedly here

spec/integration/client_side_encryption/explicit_queryable_encryption_spec.rb
```diff
@@ -5,7 +5,7 @@
 
 describe 'Explicit Queryable Encryption' do
   require_libmongocrypt
-  min_server_version '6.0'
+  min_server_version '6.0.0-rc8'
```
p-mongo, line 8: Honestly I would just keep it at `6.0`. Is your revision to version checks handling the `-rc8`?

spec/integration/client_side_encryption/queryable_encryption_examples_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+  require_libmongocrypt
+  min_server_version '6.0.0-rc8'
+  require_topology :replica_set, :sharded, :load_balanced
+
+  include_context 'define shared FLE helpers'
+
+  it 'uses queryable encryption' do
+    #  Drop data from prior test runs.
+    authorized_client.use('docs_examples').database.drop
+    authorized_client.use(key_vault_db)[key_vault_coll].drop
```
p-mongo, line 16: `key_vault_db_name`, since we do have a `Database` object that this could be confused with.

p-mongo: And probably `key_vault_coll_name` too.

docs/reference/client-side-encryption.txt
```diff
@@ -324,6 +324,203 @@ master key and create data keys, see the following sections of this tutorial:
 - `Creating A Master Key`_,
 - `Creating A Data Key`_,
 
+Queryable Encryption
+====================
+Queryable encryption is a new feature in MongoDB 6.0. It also requires
+libmongocrypt version 1.5.0-rc1 or above.
+
+Usage of automatic queryable encryption is very similar to previously
+known client-side Encryption. Below is an example of automatic queryable
```
p-mongo, line 333: Can this say what the differences are? Is it just using `:encrypted_fields` instead of `:schema_map`?

Also Encryption is capitalized while nothing else is.

p-mongo (approved):

docs/reference/client-side-encryption.txt
```diff
@@ -324,6 +324,203 @@ master key and create data keys, see the following sections of this tutorial:
...
+  )
+
+  data_key_id = client_encryption.create_data_key('local')
+  # => <BSON::Binary... type=ciphertext...>
+
+  #######################################################
+  # Step 3: Configure Mongo::Client for auto-encryption #
+  #######################################################
+
+  # Create am encrypted fields map, which tells the Mongo::Client which fields to encrypt.
```
Neilshweky, line 376:
```suggestion
  # Create an encrypted fields map, which tells the Mongo::Client which fields to encrypt.
```

Neilshweky (approved):

spec/integration/client_side_encryption/explicit_queryable_encryption_spec.rb
comandeo: Apparently, it will not run on 6.0.0-rc8 if we specify just `6.0`
```ruby
Gem::Version.new('6.0.0-rc8') < Gem::Version.new('6.0')
# => true
```

docs/reference/client-side-encryption.txt
comandeo: Apparently, some server docs were published, so I added a link.

spec/integration/client_side_encryption/explicit_queryable_encryption_spec.rb
p-mongo: I see, that is good then.

p-mongo:
:+1:

PR #2524, opened 2022-06-07
RUBY-3019 Optionally sleep before retrying tests

Neilshweky:


spec/integration/change_stream_spec.rb
```diff
@@ -4,7 +4,7 @@
 require 'spec_helper'
 
 describe 'Change stream integration' do
-  retry_test 4
+  retry_test n: 4
```
p-mongo, line 7: Can the option names please be named using words.

`n` is confusing because of https://github.com/NoRedInk/rspec-retry/pull/109, it should be either "try count" or "retry count" to reflect current reality.

`t` can also be expanded to either `delay` or `sleep` or `wait` or something along those lines.

Neilshweky: @p-mongo done, check out https://github.com/mongodb-labs/mongo-ruby-spec-shared/pull/37

p-mongo (approved):

comandeo (approved):

PR #2525, opened 2022-06-07
RUBY-2963 Improved change stream event visibility for C2C Replication

Neilshweky:


spec/runners/unified/ddl_operations.rb
```diff
@@ -66,10 +69,16 @@ def rename(op)
       collection = entities.get(:collection, op.use!('object'))
       use_arguments(op) do |args|
         to = args.use!('to')
-        collection.client.use(:admin).command({
+        opts = {
```
p-mongo, line 72: This looks like a command rather than options. It doesn't make sense to execute options (by definition they are attached to *something*, that thing is the entity being executed).

p-mongo (approved):

comandeo (approved):

spec/runners/unified/ddl_operations.rb
Neilshweky: done

PR #2532, opened 2022-06-10
RUBY-2976 Raise an error on empty list of operations to BulkWrite

Neilshweky:


lib/mongo/bulk_write.rb
```diff
@@ -324,9 +324,15 @@ def can_hint?(connection)
...
     #
-    # @raise [ Error::InvalidUpdateDocument, Error::InvalidReplacementDocument ]
+    # Raise an ArgumentError if requests is empty.
+    #
+    # @raise [ Error::InvalidUpdateDocument, Error::InvalidReplacementDocument,
+    #   ArgumentError ]
     #   if the document is invalid.
     def validate_requests!
+      if @requests.empty?
+        raise ArgumentError, "Bulk write requests cannot be empty."
```
p-mongo, line 334: Exception messages in Ruby do not (generally) have the period at the end.

spec/mongo/bulk_write_spec.rb
```diff
@@ -55,6 +55,19 @@
...
+      context 'when providing no requests' do
+
+        let(:requests) do
+          []
+        end
+
+        it 'raises an exception' do
+          expect {
+            bulk_write.execute
+          }.to raise_error(ArgumentError)
```
p-mongo, line 67: I would add the expected exception message as the second argument, to ensure that the expected exception is produced (ArgumentError is a common exception and can be triggered elsewhere).

p-mongo (approved):

lib/mongo/bulk_write.rb
Neilshweky: done

spec/mongo/bulk_write_spec.rb
Neilshweky: done

comandeo (approved):

PR #2535, opened 2022-06-16
Fix RUBY-2427 Server description on 4.4+ servers flips between having compression on & off


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2427

PR #2537, opened 2022-06-22
RUBY-3032 Do not wrap CryptError with KmsError, return KmsError directly


p-mongo:
https://jira.mongodb.com/browse/RUBY-3032


comandeo (approved):

Neilshweky (approved):
Is this a backwards breaking change?

p-mongo:
No, the same exception is returned out of the driver (KmsError). The difference is that in the exception message there is no longer a wrapped CryptError. An example is given in the ticket.

PR #2538, opened 2022-06-23
RUBY-3010 Allow unlimited connection pool

comandeo:
https://jira.mongodb.org/browse/RUBY-3010
https://jira.mongodb.org/browse/RUBY-3011

### DIFF
--- a/docs/reference/create-client.txt
+++ b/docs/reference/create-client.txt
@@ -550,7 +550,7 @@ Ruby Options
      - The maximum size of the connection pool for each server.
        Setting this option to zero sets no limit for a maximum size of the connection pool.
      - ``Integer``
-     - 5
+     - 20
 
    * - ``:max_read_retries``
      - The maximum number of read retries, when legacy read retries are used.
--- a/docs/reference/create-client.txt
+++ b/docs/reference/create-client.txt
@@ -548,7 +548,7 @@ Ruby Options
 
    * - ``:max_pool_size``
      - The maximum size of the connection pool for each server.
-       Setting this option to zero sets no limit for a maximum size of the connection pool.
+       Setting this option to zero removes the max size limit from the connection pool, permitting it to grow to any number of connections.
      - ``Integer``
      - 20
 
--- a/docs/reference/create-client.txt
+++ b/docs/reference/create-client.txt
@@ -548,6 +548,7 @@ Ruby Options
 
    * - ``:max_pool_size``
      - The maximum size of the connection pool for each server.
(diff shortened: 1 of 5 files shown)


docs/reference/create-client.txt
```diff
@@ -548,8 +548,9 @@ Ruby Options
 
    * - ``:max_pool_size``
      - The maximum size of the connection pool for each server.
+       Setting this option to zero sets no limit for a maximum size of the connection pool.
```
p-mongo, line 551:
```suggestion
       Setting this option to zero removes the max size limit from the connection pool, permitting it to grow to any number of connections.
```

spec/mongo/server/connection_pool_spec.rb
```diff
@@ -83,11 +83,21 @@
...
+      end
+    end
+
+    context 'when min size is provided and max size is zero (unlimited)' do
+      let (:options) do
+        { min_size: 10, max_size: 0 }
+      end
+
+      it 'sets max size to zero (unlimited)' do
+        expect(pool.max_size).to eq(0)
```
p-mongo, line 100: There should also be an assertion that the min size is respected and unchanged from user-provided value, in this or a new test.

lib/mongo/server/connection_pool.rb
```diff
@@ -377,7 +378,7 @@ def check_out(connection_global_id: nil)
                 #
                 # Ruby does not allow a thread to lock a mutex which it already
                 # holds.
-                if unsynchronized_size < max_size
+                if max_size == 0 || unsynchronized_size < max_size
```
p-mongo, line 381: I am not a fan of max_size attribute of the pool being zero but the alternative I can think of (have max_size be nil for an unlimited pool) is also confusing: nil max_size in constructor argument is changed to max_size=20, max_size=0 in constructor argument is changed to max_size nil. Unless you have some other ideas we can leave this the way it is typed here for now.

p-mongo (approved):
There are 2 test failures in ./spec/mongo/client_construction_spec.rb which need to be remedied.

It wouldn't hurt to have some more tests giving 0 as max size to client constructor and verifying either that the zero is correctly forwarded to pool (via method expectations) or asserting actual pool attributes.

lib/mongo/server/connection_pool.rb
comandeo: We could theoretically use `Float::INFINITY` for the `max_size`, but it is (obviously) float.

Neilshweky (approved):

PR #2540, opened 2022-06-24
RUBY-3036 Remove contention conversion for FLE

comandeo:


spec/runners/transactions/test.rb
```diff
@@ -71,10 +71,11 @@ def initialize(crud_spec, data, test)
           Operation.new(self, op)
         end
 
-        @expectations = BSON::ExtJSON.parse_obj(test['expectations'], mode: :bson)
+        mode = if bson_types then :bson else nil end
+        @expectations = BSON::ExtJSON.parse_obj(test['expectations'], mode: mode)
 
         if test['outcome']
-          @outcome = Mongo::CRUD::Outcome.new(BSON::ExtJSON.parse_obj(test['outcome'], mode: :bson))
+          @outcome = Mongo::CRUD::Outcome.new(BSON::ExtJSON.parse_obj(test['outcome'], mode: mode))
```
p-mongo, line 78: Why can this not always be `:bson` ?

p-mongo (approved):

spec/runners/transactions/test.rb
comandeo: Fixed

PR #2542, opened 2022-06-24
RUBY-1869 allow max_await_time_ms to be passed through to change stream

Neilshweky:


lib/mongo/database.rb
```diff
@@ -495,8 +495,11 @@ def aggregate(pipeline, options = {})
     #
     # @since 2.6.0
     def watch(pipeline = [], options = {})
+      view_options = { await_data: true }.merge(options) if options[:max_await_time_ms]
```
p-mongo, line 498:
```suggestion
      # TODO remove await_data: true when RUBY-3041 is done
      view_options = { await_data: true }.merge(options) if options[:max_await_time_ms]
```

spec/mongo/collection_spec.rb
```diff
@@ -798,6 +798,29 @@
...
+              authorized_collection.insert_one(a: 2)
+              authorized_collection.insert_one(a: 3)
+            end
+          end
+
+          let(:change_stream) do
+            authorized_collection.watch([], max_await_time_ms: 1500)
+          end
+
+          it 'returns the documents in the batch size specified' do
```
p-mongo, line 815: update the context & it labels to refer to this code

Neilshweky: done

p-mongo (approved):

PR #2543, opened 2022-06-25
RUBY-3038: Fix lite_spec_helper on Windows

johnnyshields:
The following changes to `lite_spec_helper.rb` are required for specs to run on Windows:
- Use "timeout" instead of "timeout_interrupt"
- Skip loading debug gems if neither is present

### DIFF
--- a/spec/lite_spec_helper.rb
+++ b/spec/lite_spec_helper.rb
@@ -32,7 +32,10 @@ else
   begin
     require 'byebug'
   rescue LoadError
-    require 'ruby-debug' rescue LoadError
+    begin
+      require 'ruby-debug'
+    rescue LoadError
+    end
   end
 end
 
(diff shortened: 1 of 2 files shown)


spec/lite_spec_helper.rb
```diff
@@ -32,7 +32,7 @@
   begin
     require 'byebug'
   rescue LoadError
-    require 'ruby-debug'
+    require 'ruby-debug' rescue LoadError
```
p-mongo, line 35: This isn't the correct syntax to rescue LoadError.

p-mongo (changes requested):
This can be merged if  the LoadError rescue is repaired.

spec/lite_spec_helper.rb
johnnyshields: Fixed

p-mongo (approved):

comandeo (approved):

PR #2547, opened 2022-06-28
RUBY-2699 make readable methods recognize chained options

Neilshweky:


spec/mongo/collection/view/readable_spec.rb
```diff
@@ -151,6 +151,70 @@
...
       end
     end
+
+    context "when using methods to set aggregate options" do
+
+      let(:aggregate) do
+        view.send(opt, param).aggregate(pipeline, options)
+      end
+
+
```
Neilshweky, line 161:
```suggestion
```

lib/mongo/collection/view/readable.rb
```diff
@@ -58,6 +71,7 @@ module Readable
         #
         # @since 2.0.0
         def aggregate(pipeline, options = {})
+          options = @options.slice(*AGGREGATE_OPTIONS).merge(options)
```
p-mongo, line 74: The AGGREGATE_OPTIONS is an array of strings but the `options` parameter is documented as a Hash with symbol keys. This usage will only work if `options` is actually a BSON::Document.

You can make AGGREGATE_OPTIONS etc. have the values as both strings and symbols or explicitly convert (better in my opinion) or perform way more manipulations here (to avoid forcing `options` into a BSON::Document if it isn't one).

lib/mongo/collection/view/readable.rb
```diff
@@ -24,6 +24,19 @@ class View
...
+        #
+        # @api private
+        COUNT_OPTIONS = ["hint", "limit", "max_time_ms", "skip", "comment"]
+
+        # The aggregate options potentially included in the options hash, using the
+        # configure method.
+        #
+        # @api private
+        AGGREGATE_OPTIONS = ["allow_disk_use", "batch_size", "max_time_ms",
+          "max_await_time_ms", "comment", "hint"]
```
p-mongo, line 38: We freeze arrays assigned to constants to prevent unintentional modification.

lib/mongo/collection/view/readable.rb
```diff
@@ -24,6 +24,19 @@ class View
       # @since 2.0.0
       module Readable
 
+        # The count options potentially included in the options hash, using the
+        # configure method.
+        #
+        # @api private
+        COUNT_OPTIONS = ["hint", "limit", "max_time_ms", "skip", "comment"]
```
p-mongo, line 31: The change overall needs a release note, and specifically for count options, passing `limit` and `skip` would change behavior meaningfully wouldn't it?

lib/mongo/collection/view/readable.rb
```diff
@@ -254,11 +270,12 @@ def estimated_document_count(opts = {})
           end
 
           %i[limit skip].each do |opt|
-            if @options.key?(opt)
+            if options.key?(opt) || opts.key?(opt)
               raise ArgumentError, "Cannot call estimated_document_count when querying with #{opt}"
```
p-mongo, line 274: This would cause existing use of `estimated_document_count` to fail when view has skip/limit applied where it worked previously with skip/limit being ignored, right?

lib/mongo/collection/view/readable.rb
```diff
@@ -384,7 +402,8 @@ def limit(limit = nil)
         #
         # @since 2.0.0
         def map_reduce(map, reduce, options = {})
-          MapReduce.new(self, map, reduce, @options.merge(options))
+          options = @options.slice(:limit, :sort).merge(options)
+          MapReduce.new(self, map, reduce, options)
```
p-mongo, line 406: This also looks like a meaningful behavior change.

lib/mongo/collection/view/readable.rb
Neilshweky: done

lib/mongo/collection/view/readable.rb
Neilshweky: yes it would, but you were already able to pass in limit and skip view the options, this just fixes it for the chained methods...

now that I think about it this change is pretty backwards breaking if you've been using these methods and didn't realize they weren't working... How would you like to handle that?

lib/mongo/collection/view/readable.rb
Neilshweky: Previously skip/limit would only cause a failure when it was in the view options

After this change it will also failed when being passed into the method parameters.

lib/mongo/collection/view/readable.rb
Neilshweky: See my comment on this above. What do you think about this change?

lib/mongo/collection/view/readable.rb
p-mongo: Since we now have feature flags, let's put the change behind an opt-in feature flag.

Neilshweky: done!

docs/reference/create-client.txt
```diff
@@ -2064,6 +2064,12 @@ The following is a list of feature flags that the Mongo Ruby Driver provides:
 
    * - Flag
      - Description
+   * - ``broken_view_options``
+     - When this flag is turned off, the view options will be correctly
+       propagated to the ``aggregate``, ``count``, ``count_documents``,
+       ``distinct``, and ``estimated_document_count`` mehods. When this flag is
+       switched on, the view options will be ignored in those methods.
+       (default: false)
```
p-mongo, line 2072: Default should be true to have same behavior as 2.17 right?

p-mongo (approved):
The code looks to be correct, just docstring written inverted.

docs/reference/create-client.txt
Neilshweky: yes, default should say true

p-mongo (approved):

PR #2548, opened 2022-06-28
RUBY-3007 make comment its own option

Neilshweky:
When I originally implemented this I incorrectly put the comment option inside the command... This PR moves it to its own option, which is consistent with the other commands.


p-mongo:
No tests?

Neilshweky:
There are additional tests in #2547 for checking that setting the comment through view options puts it in the right place.

PR #2551, opened 2022-06-30
RUBY-3027 Change removeKeyAltName helper

comandeo:


Neilshweky (approved):

lib/mongo/crypt/encryption_io.rb
```diff
@@ -192,18 +192,27 @@ def get_keys
       # Removes a key_alt_name from the key_alt_names array of the key document
       # in the key vault collection with the given id.
       def remove_key_alt_name(id, key_alt_name)
-        result = key_vault_collection.find_one_and_update(
+        key_vault_collection.find_one_and_update(
           { _id: id },
-          { '$pull' => { keyAltNames: key_alt_name } },
+          [
+            {
+              '$set' => {
```
p-mongo, line 199: You can do `:$set` if you prefer to keep all keys as symbols.

p-mongo (approved):

PR #2564, opened 2022-07-21
RUBY-2886 Add FLE shared lib support

comandeo:


lib/mongo/crypt/handle.rb
```diff
@@ -70,13 +79,29 @@ def initialize(kms_providers, kms_tls_options, options={})
...
         set_crypto_hooks
 
         Binding.setopt_kms_providers(self, kms_providers.to_document)
+
         initialize_mongocrypt
+
+        @crypt_shared_lib_required = !!options[:crypt_shared_lib_required]
+        if @crypt_shared_lib_required && crypt_shared_lib_version == 0
+          raise Mongo::Error::CryptError.new(
+            "Crypt shared library is required, but cannot be loaded"
```
p-mongo, line 102:
```suggestion
            "Crypt shared library is required, but cannot be loaded according to libmongocrypt"
```

p-mongo: I assume libmongocrypt doesn't provide any diagnostics, if it does they should be incorporated in this exception message.

p-mongo (approved):

Neilshweky (approved):

lib/mongo/crypt/handle.rb
comandeo: Fixed.

PR #2577, opened 2022-08-02
RUBY-2774 Split unified runner valid-fail tests with malformed operations into separate files

Neilshweky:


spec/runners/unified.rb
```diff
@@ -69,11 +69,11 @@ def define_unified_spec_tests(base_path, paths, expect_failure: false)
                   test.assert_events
                 # HACK: other errors are possible and likely will need to
                 # be added here later as the tests evolve.
-                rescue Mongo::Error::OperationFailure, Unified::Error::UnsupportedOperation
+                rescue ArgumentError, Mongo::Error::OperationFailure, Unified::Error::UnsupportedOperation, Unified::Error::EntityMissing
                 rescue => e
-                  fail "Expected to raise Mongo::Error::OperationFailure or Unified::Error::UnsupportedOperation, got #{e}"
+                  fail "Expected to raise ArgumenError, Mongo::Error::OperationFailure, Unified::Error::UnsupportedOperation or Unified::Error::EntityMissing, got #{e}"
```
Neilshweky, line 74:
```suggestion
                  fail "Expected to raise ArgumentError, Mongo::Error::OperationFailure, Unified::Error::UnsupportedOperation or Unified::Error::EntityMissing, got #{e}"
```

spec/runners/unified.rb
```diff
@@ -69,11 +69,11 @@ def define_unified_spec_tests(base_path, paths, expect_failure: false)
                   test.assert_events
                 # HACK: other errors are possible and likely will need to
                 # be added here later as the tests evolve.
-                rescue Mongo::Error::OperationFailure, Unified::Error::UnsupportedOperation
+                rescue ArgumentError, Mongo::Error::OperationFailure, Unified::Error::UnsupportedOperation, Unified::Error::EntityMissing
```
Neilshweky, line 72: I don't love checking against `EntityMissing` but for the tests that give a change stream that doesn't exist and expect failure, this is what we raise. I'm not sure how else to test that.

comandeo (approved):

spec/runners/unified.rb
p-mongo: EntityMissing is probably OK but why is ArgumentError here?

spec/runners/unified.rb
```diff
@@ -69,11 +69,11 @@ def define_unified_spec_tests(base_path, paths, expect_failure: false)
...
                 rescue => e
-                  fail "Expected to raise Mongo::Error::OperationFailure or Unified::Error::UnsupportedOperation, got #{e}"
+                  fail "Expected to raise ArgumentError, Mongo::Error::OperationFailure, Unified::Error::UnsupportedOperation or Unified::Error::EntityMissing, got #{e}"
                 else
-                  fail "Expected to raise Mongo::Error::OperationFailure or Unified::Error::UnsupportedOperation, but no error was raised"
+                  fail "Expected to raise error, but none was raised"
```
p-mongo, line 76: Why change to this message?

spec/runners/unified/crud_operations.rb
```diff
@@ -146,7 +146,7 @@ def insert_one(op)
         if session = args.use('session')
           opts[:session] = entities.get(:session, session)
         end
-        collection.insert_one(args.use!('document'), **opts)
+        collection.insert_one(args.use('document'), **opts)
```
p-mongo, line 149: What happened here?

spec/runners/unified/change_stream_operations.rb
```diff
@@ -37,5 +37,11 @@ def iterate_until_document_or_error(op)
       object = entities.get(:change_stream, object_id)
       object.to_enum.next
     end
+
+    def close(op)
+      object_id = op.use!('object')
+      object = entities.get(:change_stream, object_id)
```
p-mongo, line 43: `close` is an operation on cursor, not change stream: https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#close

spec/runners/unified.rb
Neilshweky: It was just getting really long and I didn't feel like it was necessary to list all the errors if none were raised.

spec/runners/unified/crud_operations.rb
Neilshweky: One of the tests added was to test that if you supply a `nil` document to insert_one, it fails. 

We do, in fact, fail in this case, and we raise an ArgumentError. This is why I added the ArgumentError above the errors we check for.

spec/runners/unified/change_stream_operations.rb
Neilshweky: We do define a method close on change_stream: https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/collection/view/change_stream.rb#L237

Neilshweky: oh I see, this was meant to go on the cursor... Okay I will switch it.

spec/runners/unified.rb
Neilshweky: see my comment: https://github.com/mongodb/mongo-ruby-driver/pull/2577#discussion_r937035251

spec/runners/unified/crud_operations.rb
p-mongo: This can be handled in one of two ways: eitheir `use` can be modified to check for key presence via `key?` and return the value of that key even if it's nil, or the `ArgumentError` needs to be reraised as a unified spec runner exception (like the other exceptions appearing in this PR) and that specific exception can be rescued here. The second option retains an early check on values being `use`d being not nil everywhere else in the spec runners which I think is good, thus would be my preference.

PR #2579, opened 2022-08-03
RUBY-3081 Avoid creating mongocryptd client

comandeo:


Neilshweky (approved):

spec/mongo/crypt/auto_encrypter_spec.rb
```diff
@@ -157,6 +157,13 @@
         expect(auto_encrypter.metadata_client).not_to eq(client)
       end
     end
+
+    context 'when crypt shared library is available' do
+      it 'does not create a mongocryptd client' do
+        allow_any_instance_of(Mongo::Crypt::Handle).to receive(:"crypt_shared_lib_available?").and_return true
+        expect(auto_encrypter.mongocryptd_client).to be_nil
+      end
```
p-mongo, line 165: While there is nothing wrong with this test, there should ideally be an integration test that asserts the client is functional, it doesn't spawn mongocryptd, and it doesn't wait for mongocryptd for 10 seconds.

lib/mongo/crypt/auto_encrypter.rb
```diff
@@ -117,7 +117,7 @@ def initialize(options)
           @crypt_handle.crypt_shared_lib_available? ||
           @options[:extra_options][:crypt_shared_lib_required]
 
-        if !@options[:extra_options][:crypt_shared_lib_required]
+        unless @options[:extra_options][:crypt_shared_lib_required] || @crypt_handle.crypt_shared_lib_available?
```
p-mongo, line 120: Personally I only use `unless` if there is a single clause in the condition, i.e. I would've phrased this as `if !a && !b` which I think is easier to read.

PR #2585, opened 2022-08-05
RUBY-3083 Fix AWS auth test configuration

comandeo:


p-mongo (approved):
I was wondering about that. What is `-v` supposed to do?

comandeo:
`-v` was supposed to check whether `FLE` variable is set. It turned out it is always set, but it is empty for non-fle configurations.

PR #2586, opened 2022-08-05
RUBY-2154 Add link to BSON docs to CSFLE tutorial

comandeo:


p-mongo:
This would be much more usable if it linked to the correct section, I can investigate that.

Neilshweky (approved):

PR #2587, opened 2022-08-05
RUBY-1627 Document collection insert options

Neilshweky:
I know that the Insert constructor takes the following option:
```
# @option options :flags [Array] The flags for the insertion message.
#   Supported flags: +:continue_on_error+
```
but it's not tested anywhere and it's not part of the specification. 

I don't think collation is relevant in this case.

### DIFF
--- a/lib/mongo/collection.rb
+++ b/lib/mongo/collection.rb
@@ -644,13 +644,13 @@ module Mongo
     # @param [ Hash ] document The document to insert.
     # @param [ Hash ] opts The insert options.
     #
-    # @option opts [ Session ] :session The session to use for the operation.
-    # @option opts [ Hash ] :write_concern The write concern options.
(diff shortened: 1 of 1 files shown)


lib/mongo/collection.rb
```diff
@@ -690,6 +698,10 @@ def insert_one(document, opts = {})
     # @option options [ true | false ] :ordered Whether the operations
     #   should be executed in order.
     # @option options [ Session ] :session The session to use for the operation.
+    # @option options [ Hash ] :write_concern The write concern options.
+    #   Can be :w => Integer, :fsync => Boolean, :j => Boolean.
+    # @option options [ true, false ] :bypass_document_validation Whether or
+    #   not to skip document level validation.
```
Neilshweky, line 704: bulk write doesn't seem to take a comment....

comandeo (approved):
:+1:

p-mongo (approved):

PR #2588, opened 2022-08-05
RUBY-1673 Document collection options

Neilshweky:


lib/mongo/collection.rb
```diff
@@ -204,17 +216,37 @@ def write_concern_with_session(session)
       wc
     end
 
-    # Provides a new collection with either a new read preference or new write concern
-    # merged over the existing read preference / write concern.
+    # Provides a new collection with either a new read preference, new read
+    # concern or new write concern merged over the existing read preference /
+    # read concern / write concern.
```
Neilshweky, line 221: As per RUBY-3084, the `CHANGEABLE_OPTIONS` contains both `read` and `read_concern`. The `read` option corresponds to the `read_preference` options. Now, this method has only documented the ability to change the read preference and write concern, while allowing the changing of read preference, **read concern**, and write concern.

There are two options here: 
- document that `with` also allows changing read concern, which is what's done here
- change `with` to not allow the changing of read concern as per RUBY-3084.

This PR implements the former. I can also move this snippet to its own PR if that would be better.

lib/mongo/collection.rb
```diff
@@ -107,12 +107,24 @@ def ==(other)
...
     # @option options [ Integer ] :expire_after Number indicating
     #   after how many seconds old time-series data should be deleted.
     # @options clustered_index [ Hash ] :clustered_index Create a clustered index.
     #   This option specifies how this collection should be clustered on _id.
     #   See https://www.mongodb.com/docs/v5.3/reference/method/db.createCollection/#std-label-db.createCollection.clusteredIndex
     #   for more information about this option.
+    # @option options [ Hash ] :collation The collation to use.
     #
     # @since 2.0.0
     def initialize(database, name, options = {})
```
alexbevi, line 130: `:view_on` doesn't appear to be documented (https://github.com/mongodb/mongo-ruby-driver/blob/85ac96b1d282d59ab9b7dea235d6e32d3f9e325f/lib/mongo/collection.rb#L75). Might as well add that too.

alexbevi: Note it doesn't appear `:pipeline`, which the [`create` command](https://www.mongodb.com/docs/manual/reference/command/create/) would use to filter the view is defined here.

Neilshweky: done!

Neilshweky:
This is ready for review.

spec/mongo/collection_crud_spec.rb
```diff
@@ -4412,4 +4412,24 @@ def generate
       expect(result).to be_nil
     end
   end
+
+  context "when creating collection with view_on and pipeline" do
```
Neilshweky, line 4416: Shoutout to @alexbevi for this test

alexbevi (approved):
These changes LGTM

lib/mongo/collection.rb
p-mongo: https://jira.mongodb.org/browse/RUBY-3086

p-mongo (approved):

lib/mongo/collection.rb
p-mongo: I made the ticket before reading this PR, I see that you've done several of the changes listed there in this PR already.

comandeo (approved):

PR #2589, opened 2022-08-05
RUBY-1554 Mention to_json in logging examples in tutorial

Neilshweky:
Is this what you had in mind @p-mongo?

### DIFF
--- a/docs/reference/monitoring.txt
+++ b/docs/reference/monitoring.txt
@@ -94,6 +94,11 @@ Sample output:
   D, [2018-09-23T13:47:31.258020 #4692] DEBUG -- : COMMAND | 127.0.0.1:27027 | test.hello | STARTED | {"hello"=>1, "$readPreference"=>{"mode"=>"primary"}, "lsid"=>{"id"=><BSON::Binary:0x47111693353080 type=uuid data=0x730341e880dc40a2...>}}
   D, [2018-09-23T13:47:31.259145 #4692] DEBUG -- : COMMAND | 127.0.0.1:27027 | test.hello | SUCCEEDED | 0.000791175s
 
+.. note::
+
+  The default inspection of a command which is a BSON document gets truncated
+  in the middle. To get the full rendering of the command, the ``to_json``
+  method can be called on the document: ``event.command.to_json``.
 
 .. _sdam:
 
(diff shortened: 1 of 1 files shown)


docs/reference/monitoring.txt
```diff
@@ -94,6 +94,11 @@ Sample output:
   D, [2018-09-23T13:47:31.258020 #4692] DEBUG -- : COMMAND | 127.0.0.1:27027 | test.hello | STARTED | {"hello"=>1, "$readPreference"=>{"mode"=>"primary"}, "lsid"=>{"id"=><BSON::Binary:0x47111693353080 type=uuid data=0x730341e880dc40a2...>}}
   D, [2018-09-23T13:47:31.259145 #4692] DEBUG -- : COMMAND | 127.0.0.1:27027 | test.hello | SUCCEEDED | 0.000791175s
 
+.. note::
+
+  The default inspection of a command which is a BSON document gets truncated
+  in the middle. To get the full rendering of the command, the ``to_json``
+  method can be called on the document: ``event.command.to_json``.
```
Neilshweky, line 102:
```suggestion


```

docs/reference/monitoring.txt
```diff
@@ -94,6 +94,12 @@ Sample output:
   D, [2018-09-23T13:47:31.258020 #4692] DEBUG -- : COMMAND | 127.0.0.1:27027 | test.hello | STARTED | {"hello"=>1, "$readPreference"=>{"mode"=>"primary"}, "lsid"=>{"id"=><BSON::Binary:0x47111693353080 type=uuid data=0x730341e880dc40a2...>}}
   D, [2018-09-23T13:47:31.259145 #4692] DEBUG -- : COMMAND | 127.0.0.1:27027 | test.hello | SUCCEEDED | 0.000791175s
 
+.. note::
+
+  The default inspection of a command which is a BSON document gets truncated
```
comandeo, line 99: I would add this as a comment inside the `started` method, where we actually use `event.command`. But here is fine, too.

comandeo (approved):

docs/reference/monitoring.txt
Neilshweky: Sounds like a better idea, I'll throw that up tomorrow

p-mongo: I think this was my first impression too, `event` is not available in the suggested scope?

Neilshweky:
@p-mongo this is rfal when you return from vacation.

@comandeo  is this what you had in mind?

comandeo (approved):

docs/reference/monitoring.txt
```diff
@@ -41,7 +41,10 @@ logging subscriber used internally by the driver:
     include Mongo::Loggable
 
     def started(event)
-      log_debug("#{prefix(event)} | STARTED | #{format_command(event.command)}")
+      # The default inspection of a command which is a BSON document gets
+      # truncated in the middle. To get the full rendering of the command, the
+      # ``to_json`` method can be called on the document.
+      log_debug("#{prefix(event)} | STARTED | #{format_command(event.command.to_json)}")
```
p-mongo, line 47: I don't think this is quite correct:

```
irb(main):004:0> puts ({a:4}.to_json.inspect)
"{\"a\":4}"
=> nil
irb(main):005:0> puts ({a:4}.inspect)
{:a=>4}
=> nil
```
This code will output escaped json.

Perhaps
```
      log_debug("#{prefix(event)} | STARTED | #{event.command.to_json}")
```

PR #2595, opened 2022-08-10
RUBY-1613 Verify handling of write concern errors by user management commands function correctly

Neilshweky:


spec/mongo/auth/user/view_spec.rb
```diff
@@ -486,4 +486,100 @@
       end
     end
   end
+
+  context "when the result is a write concern error" do
+
+    let(:admin_client) { authorized_client.use('admin') }
```
Neilshweky, line 492:
```suggestion
    require_topology :replica_set
    let(:admin_client) { authorized_client.use('admin') }
```

p-mongo (approved):

PR #2596, opened 2022-08-12
RUBY-1768 change to_s to return original URI

Neilshweky:


p-mongo:
I think a bigger question here is what will happen if the original URI was partially invalid and the invalid parts were dropped during parsing. `to_s` would now return something that is technically an invalid URI right?

If you do decide to return `@string` I think you would need to `dup` it otherwise clients can end up mutating it.

p-mongo:
Also what happens for SRV URIs?

Neilshweky:
@p-mongo that's a good point, I learned that options with invalid values should be dropped (when srvMaxHosts is negative, for example), so I decided to reconstruct the uri from its parts. I have added tests for regular URIs and srv URIs.

spec/mongo/uri/srv_protocol_spec.rb
```diff
@@ -583,13 +637,15 @@
...
+
+          include_examples "roundtrips string"
         end
       end
 
       context 'read preference max staleness option provided' do
 
         let(:options) do
-          'readPreference=Secondary&maxStalenessSeconds=120'
+          'readPreference=secondary&maxStalenessSeconds=120'
```
p-mongo, line 648: Does this lowercasing behavior have a test?

spec/mongo/uri_spec.rb
```diff
@@ -1278,12 +1394,14 @@
     end
 
     context 'when an app name option is provided' do
-      let(:options) { "appname=uri_test" }
+      let(:options) { "appName=uri_test" }
```
p-mongo, line 1397: Or is it canonical casing? Whichever way it is should also be documented.

lib/mongo/uri.rb
```diff
@@ -335,11 +335,31 @@ def database
     #
     # @return [ String ] The uri string.
     def to_s
-      @string
+      reconstruct_uri
     end
 
     private
 
+    # Reconstruct the URI from its parts.
```
p-mongo, line 343: So this should say that invalid option values are dropped and option names are converted to lower|canonical case.

p-mongo (commented):
What do you think about having ability to retrieve both the original uri and the "correct" uri? And designated methods for this, e.g. `to_uri` and `original_uri`. The "URI" in class name makes this confusing because technically URI is the string, a URI parsed into components is strictly no longer a URI.

spec/mongo/uri_spec.rb
Neilshweky: I'm gonna put this back the way it was.

lib/mongo/uri.rb
Neilshweky: done

spec/mongo/uri/srv_protocol_spec.rb
Neilshweky: I'm going to put this back as well

Neilshweky:
@p-mongo this is rfal

Neilshweky:
@p-mongo this is rfal

comandeo (approved):

lib/mongo/uri.rb
```diff
@@ -328,8 +328,45 @@ def database
...
+    #   uri.to_s
+    #
+    # @return [ String ] The uri string.
+    def to_s
+      reconstruct_uri
+    end
+
     private
 
+    # Reconstruct the URI from its parts. Invalid options are drops and options
```
Neilshweky, line 343:
```suggestion
    # Reconstruct the URI from its parts. Invalid options are dropped and options
```

PR #2597, opened 2022-08-15
RUBY-1775 Make driver Ruby warning-clean

Neilshweky:


lib/mongo/auth/aws/request.rb
```diff
@@ -215,7 +215,7 @@ def canonical_request
         # @return [ String ] The signature.
         def signature
           hashed_canonical_request = Digest::SHA256.hexdigest(canonical_request)
-          hashed_body = Digest::SHA256.new.update(STS_REQUEST_BODY).hexdigest
+          Digest::SHA256.new.update(STS_REQUEST_BODY).hexdigest
```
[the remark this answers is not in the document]
comandeo: I believe this line can be safely removed.

comandeo (approved):

lib/mongo/auth/aws/request.rb
Neilshweky, line 218:
```suggestion
```

lib/mongo/auth/aws/conversation.rb
```diff
@@ -65,10 +65,6 @@ def continue(reply_document, connection)
         # @return [ Integer ] The conversation id.
         attr_reader :conversation_id
 
-        def client_first_payload
```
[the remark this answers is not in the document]
p-mongo: Yes, looks like the other payload methods return the serialized bson which this one doesn't so this is probably a leftover and should've been removed.

lib/mongo/auth/aws/request.rb
```diff
@@ -215,7 +215,6 @@ def canonical_request
         # @return [ String ] The signature.
         def signature
           hashed_canonical_request = Digest::SHA256.hexdigest(canonical_request)
-          hashed_body = Digest::SHA256.new.update(STS_REQUEST_BODY).hexdigest
```
p-mongo, line 218: I think this ended up being in the hashed canonical request.

lib/mongo/collection/view/readable.rb
```diff
@@ -272,19 +272,18 @@ def estimated_document_count(opts = {})
...
-                  selector: cmd,
-                  db_name: database.name,
-                  read: read_pref,
-                  session: session,
-                  comment: opts[:comment],
-                ).execute(server, context: context)
-                result.n.to_i
+                cmd[:readConcern] = Options::Mapper.transform_values_to_strings(read_concern)
               end
+              result = Operation::Count.new(
```
p-mongo, line 277: Interesting. I guess there is always a read concern by this point.

lib/mongo/auth/aws/request.rb
comandeo: I am wondering how is it possible. We create a new digest here, and it is not used anywhere else.

p-mongo: The body is part of canonical request. This line is likely a leftover from the in-progress implementation of aws auth.

PR #2598, opened 2022-08-24
RUBY-3033 update sdam specs

Neilshweky:
I did not add the following spec tests: 
- auth-network-timeout-error
- find-network-timeout-error

There's a very specific comment alerting that we don't want to raise in this case:
```
# possibly cluster is slow, do not give up on it
```
and we didn't implement these tests in the past, so I'm electing to leave them out now.

I also didn't implement `minPoolSize-error` spec since we don't have a poolReadyEvent, and we don't pause and resume the pool.


spec/runners/unified/thread_operations.rb
```diff
@@ -0,0 +1,73 @@
...
+    def wait_for_thread(op)
+      consume_test_runner(op)
+      use_arguments(op) do |args|
+        thread = entities.get(:thread, args.use!('thread'))
+        thread.context.signal_stop
+        thread.join
+      end
+    end
+  end
+end
```
Neilshweky, line 73:
```suggestion
end

```

p-mongo (approved):

p-mongo (approved):

PR #2603, opened 2022-08-29
Compatibility table updates

mongoKart:
Our driver documentation uses a circle asterisk to indicate partial compatibility with a MongoDB version. This PR adds that symbol to the Ruby compat table, as well as a legend that shows the reader what each symbol means.

### DIFF
--- a/docs/includes/compatibility-table-legend.rst
+++ b/docs/includes/compatibility-table-legend.rst
@@ -17,4 +17,4 @@ Compatibility Table Legend
    * - No mark
      - The Driver version is not tested with the MongoDB version.
    * - D
-     - Support for the MongoDB version is deprecated and will be removed in a future     Driver version.
+     - Support for the MongoDB version is deprecated and will be removed in a future Driver version.
--- a/docs/includes/compatibility-table-legend.rst
+++ b/docs/includes/compatibility-table-legend.rst
@@ -1,0 +1,20 @@
+Compatibility Table Legend
+++++++++++++++++++++++++++
+
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+   :class: compatibility
+
+   * - Icon
+     - Explanation
+
+   * - ✓
+     - All features are supported.
+   * - ⊛
+     - The Driver version will work with the MongoDB version, but not all
+       new MongoDB features are supported.
+   * - No mark
+     - The Driver version is not tested with the MongoDB version.
+   * - D
+     - Support for the MongoDB version is deprecated and will be removed in a future     Driver version.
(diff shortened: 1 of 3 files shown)


comandeo (approved):

p-mongo:
@mongoKart  Most of the empty spaces in the table that have been replaced with the wheel are for configurations that have not been tested. For example, 2.6-stable was tested against 4.0 server and no newer versions. If the explanations of the icons are correct, the checkmarks that have footnotes stating that some support is implemented in later driver versions should be replaced by wheels, and all empty spaces should remain empty (unless you've gone through the history and checked each stable branch for what it tested, but our server support table is generally quite accurate).

p-mongo:
If the wheel icon means the combination wasn't tested but is assumed to work because server shouldn't have changed behavior, then shouldn't all empty cells be replaced with wheels, e.g. going back to 2.0 driver?

mongoKart:
> @mongoKart Most of the empty spaces in the table that have been replaced with the wheel are for configurations that have not been tested. For example, 2.6-stable was tested against 4.0 server and no newer versions. If the explanations of the icons are correct, the checkmarks that have footnotes stating that some support is implemented in later driver versions should be replaced by wheels, and all empty spaces should remain empty (unless you've gone through the history and checked each stable branch for what it tested, but our server support table is generally quite accurate).

For context, [see this Jira ticket](https://jira.mongodb.org/browse/DOCS-15172). My understanding from conversations with @terakilobyte was that any language driver that is compatible with server 4.0 will also work with later versions of server, though all new features may not be supported. If this isn't true for the Ruby driver, please let me know.

Neilshweky:
> > @mongoKart Most of the empty spaces in the table that have been replaced with the wheel are for configurations that have not been tested. For example, 2.6-stable was tested against 4.0 server and no newer versions. If the explanations of the icons are correct, the checkmarks that have footnotes stating that some support is implemented in later driver versions should be replaced by wheels, and all empty spaces should remain empty (unless you've gone through the history and checked each stable branch for what it tested, but our server support table is generally quite accurate).
> 
> For context, [see this Jira ticket](https://jira.mongodb.org/browse/DOCS-15172). My understanding from conversations with @terakilobyte was that any language driver that is compatible with server 4.0 will also work with later versions of server, though all new features may not be supported. If this isn't true for the Ruby driver, please let me know.

I don't see any reason for 4.0+ to be compatible with the new versions of the driver, the only problem is that it's not being tested. 

If you want to change the wheel to indicate, "should work but untested" that would be accurate, but I'm not sure that's helpful.

Neilshweky:
Additionally, the rebase here is going to be slightly painful, we removed server version 5.1, 5.2 from our compat tables. I would do it but I don't have push permission to this branch.

Neilshweky (commented):
Left comments above

PR #2605, opened 2022-08-31
RUBY-2989 Obtain AWS credentials for CSFLE

comandeo:


Neilshweky (approved):

lib/mongo/crypt/binding.rb
```diff
@@ -775,13 +775,14 @@ def self.ctx_explicit_decrypt_init(context, doc)
 
       # An enum labeling different libmognocrypt state machine states
       enum :mongocrypt_ctx_state, [
```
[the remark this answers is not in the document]
p-mongo: This in an FFI feature (to go with C enums), not Ruby's. https://github.com/ffi/ffi/wiki/Enums

lib/mongo/crypt/context.rb
```diff
@@ -110,11 +114,24 @@ def run_state_machine
...
 
             Binding.ctx_kms_done(self)
+          when :need_kms_credentials
+            providers = {}
+            if kms_providers.aws&.empty?
+              begin
+                aws_credentials = Mongo::Auth::Aws::CredentialsRetriever.new.credentials
+              rescue Auth::InvalidConfiguration
+                raise Error::CryptError.new(
+                  "Could not locate AWS credentials (checked environment variables, ECS and EC2 metadata)"
```
p-mongo, line 124: Can you please include the original exception via the `#{e.class}: #{e}` syntax.

p-mongo (approved):

lib/mongo/crypt/context.rb
comandeo: The original exception will be confusing here, I believe. It says  `"Could not locate AWS credentials (checked Client URI and Ruby options, environment variables, ECS and EC2 metadata)"`, even though we do not check Client URI and Ruby options here (and we should not, according to the spec). Thats's why I replaced the original message here.

Would it be enough if we just include the original exception class?

p-mongo: There are 4 places where InvalidConfiguration is raised. The one with "Could not locate AWS credentials" could be replaced by a subclass, then the subclass would be raised in just that one place and then that exception can be rescued here and the message completely replaced as you are doing now.

comandeo: Good suggestion, thanks! Changed accordingly.

PR #2606, opened 2022-08-31
RUBY-3095 write tests, docs for option mapper

Neilshweky:


lib/mongo/uri/options_mapper.rb
```diff
@@ -290,11 +290,12 @@ def self.uri_option(uri_key, name, **extra)
       # Returns true for 'true', false for 'false', otherwise nil.
       #
       # @param [ String ] name Name of the URI option being processed.
-      # @param value [ String | true | false ] URI option value.
+      # @param [ String | true | false ] value URI option value.
       #
       # @return [ true | false | nil ] Converted value.
       def convert_bool(name, value)
         case value
+          # note do we want to include 1, 0, off, on etc? or is this good enough
```
Neilshweky, line 298:
```suggestion
```

lib/mongo/uri/options_mapper.rb
```diff
@@ -422,27 +452,38 @@ def revert_symbol(value)
...
       end
 
+      # Reverts an array.
+      #
+      # @param [ Array<String> ] value An array of strings.
+      #
+      # @return [ String ] The array as a string.
       def revert_array(value)
-        value
+        value.join(',')
```
p-mongo, line 466:
```suggestion
        value
```

lib/mongo/uri/options_mapper.rb
```diff
@@ -472,62 +513,82 @@ def convert_auth_mech_props(name, value)
...
 
+      # Reverts auth mechanism properties.
+      #
+      # @param [ Hash | nil ] value The auth mech properties.
+      #
+      # @return [ String | nil ] The auth mechanism properties as a string.
       def revert_auth_mech_props(value)
-        value
+        return if value.nil?
+        value.map { |k, v| "#{k}:#{v}" }.join(',')
```
p-mongo, line 523: This should go back to `value`

lib/mongo/uri/options_mapper.rb
```diff
@@ -472,62 +513,82 @@ def convert_auth_mech_props(name, value)
...
+        int = if value.is_a?(String) && /\A-?\d+\z/ =~ value
+          value.to_i
+        elsif value.is_a?(Integer)
+          value
+        end
 
-          if int == -1
-            int = nil
-          end
+        if int.nil?
```
p-mongo, line 541: When `value` is neither string nor integer this branch won't be taken and code will continue to the comparison with -1 which looks to be wrong.

lib/mongo/uri/options_mapper.rb
```diff
@@ -472,62 +513,82 @@ def convert_auth_mech_props(name, value)
...
+          log_warn("max staleness should be either 0 or greater than 90: #{value}")
+          int = nil
         end
 
-        log_warn("Invalid max staleness value: #{value}")
-        nil
+        int
       end
 
+      # Reverts max stalenss.
```
p-mongo, line 558:
```suggestion
      # Reverts max staleness.
```

lib/mongo/uri/options_mapper.rb
```diff
@@ -537,16 +598,21 @@ def convert_read_tags(name, value)
...
       end
 
+      # Reverts read tags.
+      #
+      # @param [ Array<Hash> | nil ] value The read tags.
+      #
+      # @return [ Array<String> | nil ] A list of read tags.
       def revert_read_tags(value)
-        value
+        value&.map { |ar| ar.map { |k, v| "#{k}:#{v}" }.join(',') }
```
p-mongo, line 607:
```suggestion
        value
```

lib/mongo/uri/options_mapper.rb
```diff
@@ -572,9 +638,14 @@ def convert_w(name, value)
...
+      # Reverts write concern.
+      #
+      # @param [ Integer | Symbol | String ] value The write concern.
+      #
+      # @return [ String ] The write concern as a string.
       def revert_w(value)
         case value
-        when Symbol
+        when Symbol, Integer
           value.to_s
```
p-mongo, line 649: Integer shouldn't be converted to string but returned as is.

lib/mongo/uri/options_mapper.rb
```diff
@@ -584,31 +655,38 @@ def revert_w(value)
...
-          end
+        i = if value.is_a?(String) && /\A-?\d+\z/ =~ value
+          value.to_i
+        elsif value.is_a?(Integer)
+          value
         end
 
-        log_warn("#{value} is not a valid zlibCompressionLevel")
-        nil
+        if i && (i >= -1 && i <= 9)
```
p-mongo, line 669: This also I think mishandles input which is neither string nor integer.

p-mongo (commented):
I would prefer to have tests covering the entire process in addition to (or even instead of, though you've already written a ton of tests here) unit tests for individual conversion methods. Meaning, for example, go from `{server_selection_timeout: 10}` as the set of Ruby options to `{serverSelectionTimeoutMS: 10000}` set of SMC options (via `ruby_to_smc_options`). The same data set could also be used to test the opposite direction (`smc_to_ruby_options`).

lib/mongo/uri/options_mapper.rb
Neilshweky: I don't think that's correct... If it's neither a string or an integer, int will be nil and then warn and exit before it gets to that point... I should just combine those branches, however

lib/mongo/uri/options_mapper.rb
Neilshweky: Again, when it's neither a string nor integer, i will be nil and exit (this time on the first part of the if condition)

Neilshweky: Again, when it's neither a string nor integer, i will be nil and exit (this time on the first part of the if condition)

Neilshweky:
@p-mongo  this is rfal

lib/mongo/uri/options_mapper.rb
p-mongo: I see, sneaky.

p-mongo (approved):

PR #2609, opened 2022-09-02
RUBY-3095 remove readConcernLevel special case


p-mongo:
https://jira.mongodb.com/browse/RUBY-3095


p-mongo:
The symbol->string conversion is handled by revert_symbol, thus this case shouldn't now be needed.

Neilshweky (approved):

PR #2610, opened 2022-09-02
RUBY-3104 support all options in collection create

Neilshweky:


lib/mongo/collection.rb
```diff
@@ -372,13 +372,13 @@ def capped?
     # @since 2.0.0
     def create(opts = {})
       # Passing read options to create command causes it to break.
-      # Filter the read options out.
+      # Filter the read options out. The same is true with session
+      # since it gets passed to the operation and fails BSON serialization.
```
p-mongo, line 376: Unless I'm misunderstanding the situation, the correct comment here would be that session is handled further below on line 391. It is ultimately passed to the server just using a different code path.

p-mongo (approved):

comandeo (approved):

p-mongo (approved):

PR #2614, opened 2022-09-05
RUBY-3099 Add Ruby 3.1 to compat tables

comandeo:


docs/reference/driver-compatibility.txt
```diff
@@ -365,6 +367,7 @@ for that Ruby version is deprecated.
      - |checkmark|
      - |checkmark|
      - |checkmark|
+     - |checkmark|
```
p-mongo, line 370: 2.17-stable is not tested against 3.1, if this change is merged into master it needs to be backported there and the configurations added.

p-mongo (approved):

Neilshweky (approved):

PR #2615, opened 2022-09-06
RUBY-3105 Do not raise errors when closing cursor

comandeo:


lib/mongo/cursor.rb
```diff
@@ -294,6 +291,8 @@ def close
       end
 
       nil
+    rescue Error::OperationFailure, Error::SocketError, Error::SocketTimeoutError => e
```
Neilshweky, line 294:
```suggestion
    rescue Error::OperationFailure, Error::SocketError, Error::SocketTimeoutError
```
Just so that we don't get a warning about an unused variable

Neilshweky (approved):

p-mongo (approved):

PR #2616, opened 2022-09-08
RUBY-3079 Propagate Original Error for Write Errors Labeled NoWritesPerformed

Neilshweky:


spec/integration/retryable_writes_errors_spec.rb
```diff
@@ -31,4 +31,60 @@
       end
     end
   end
+
+  context "when encountering a WriteConcernError with a RetryableWriteError label" do
+    require_topology :replica_set
```
Neilshweky, line 36:
```suggestion
    require_topology :replica_set
    require_fail_command
```

spec/integration/retryable_writes_errors_spec.rb
```diff
@@ -31,4 +31,61 @@
       end
     end
   end
+
+  context "when encountering a WriteConcernError with a RetryableWriteError label" do
```
p-mongo, line 35: Shouldn't this say something about NoWritesPerformed?

p-mongo (approved):

spec/integration/retryable_writes_errors_spec.rb
Neilshweky: done

PR #2617, opened 2022-09-09
RUBY-3111 Fix test failures in BSON master cause by the changing of the Illegal Key error

Neilshweky:


spec/integration/retryable_writes_errors_spec.rb
```diff
@@ -35,6 +35,7 @@
   context "when encountering a NoWritesPerformed error after an error with a RetryableWriteError label" do
     require_topology :replica_set
     require_fail_command
+    min_server_version '6.0'
```
Neilshweky, line 38: Following the Go drivers lead on this one for now... I have a comment in the ticket asking if this should be there.

Neilshweky: Apparently fail command error labels only work on 4.4+...

p-mongo (approved):

PR #2619, opened 2022-09-09
RUBY-3113 add client param to sdam_proc call

Neilshweky:


matthewdale (approved):
Looks good, thanks! 👍

p-mongo (approved):

PR #2621, opened 2022-09-13
RUBY-2752 Integration tests for observeSensitiveCommands

Neilshweky:


spec/runners/unified/assertions.rb
```diff
@@ -156,7 +156,7 @@ def assert_events
         else
           false
         end
-        actual_events = subscriber.wanted_events(@observe_sensitive)
+        actual_events = subscriber.wanted_events(@observe_sensitive[client])
```
p-mongo, line 159:
```suggestion
        actual_events = subscriber.wanted_events(@observe_sensitive.fetch(client))
```

p-mongo: I would also index with `client_id` rather than `client`.

spec/runners/unified/event_subscriber.rb
```diff
@@ -30,8 +30,13 @@ def wanted_events(observe_sensitive = false)
         events
       else
         events.reject do |event|
-          event.respond_to?(:command_name) &&
-            %w(authenticate getnonce saslStart saslContinue).include?(event.command_name)
+          if event.respond_to?(:command_name)
+            command = event.respond_to?(:command) ? event.command : event.started_event.command
```
p-mongo, line 34: When you use `event.started_event`, what is `event` in that expression?

spec/runners/unified/assertions.rb
p-mongo: Or can `nil` be passed to `wanted_events`?

Neilshweky: nil can (and should) be passed to `wanted_events`...

I indexed by client because that's what we do for the `subscribers` hash but I can change it.

spec/runners/unified/event_subscriber.rb
Neilshweky: The event is a CommandSucceeded event, which doesn't contain a command.

p-mongo: Could you extract that on its own line for clarity please, e.g.

```
started_event = event.respond_to?(:command) ? event : event.started_event
```

spec/runners/unified/assertions.rb
p-mongo: In that case it's fine the way it is.

p-mongo (approved):

spec/runners/unified/event_subscriber.rb
Neilshweky, line 34:
```suggestion
            # event could be a command started event or command succeeded event
            command = event.respond_to?(:command) ? event.command : event.started_event.command
```

Neilshweky: done

p-mongo (approved):

PR #2622, opened 2022-09-13
RUBY-2671 remove old drivers, remove rapid releases, update kerberos compat

Neilshweky:
This PR does three things: 
- Remove drivers <2.5
- Remove server rapid release versions (5.1, 5.2)
- Updates the kerberos versioning (I'm not sure if this is actually correct but I don't recall us dropping support)

If these three things are okay, I will update the ticket (RUBY-2671) with these changes.

### DIFF
--- a/docs/reference/driver-compatibility.txt
+++ b/docs/reference/driver-compatibility.txt
@@ -39,8 +39,6 @@ version.
 
    * - Ruby Driver
      - MongoDB 6.0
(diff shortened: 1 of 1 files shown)


p-mongo (approved):

p-mongo:
We haven't touched kerberos anything in years but according to CI it's working, so that update is good with me.

PR #2639, opened 2022-09-26
RUBY-1253 Detect cursor closure from other thread

comandeo:


lib/mongo/cursor.rb
```diff
@@ -164,6 +166,9 @@ def each
         # StopIteration raised by try_next ends this loop.
         loop do
           document = try_next
+          if explicitly_closed?
+            raise Error::InvalidCursorOperation, 'Cursor was explicitly closed.'
```
p-mongo, line 170:
```suggestion
            raise Error::InvalidCursorOperation, 'Cursor was explicitly closed'
```

lib/mongo/cursor.rb
```diff
@@ -172,6 +177,9 @@ def each
...
```
p-mongo, line 181:
```suggestion
            raise Error::InvalidCursorOperation, 'Cursor was explicitly closed'
```

p-mongo (approved):

PR #2640, opened 2022-09-28
RUBY-2808 Add a CMAP test that verifies the background thread hands over connections to threads doing checkout

neilshweky:
Look at #2641 when this is done

Still failing tests related to this. I cannot reproduce them.

### DIFF
--- a/spec/runners/cmap/verifier.rb
+++ b/spec/runners/cmap/verifier.rb
@@ -41,7 +41,9 @@ module Mongo
           end
         end
 
-        expect(actual_modified).to eq(expected)
+        expected.each do |k, v|
+          expect(actual_modified[k]).to eq(v)
+        end
       end
     end
   end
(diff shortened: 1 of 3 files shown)


spec/runners/cmap/verifier.rb
```diff
@@ -41,7 +41,9 @@ def verify_hashes(actual, expected)
           end
         end
 
-        expect(actual_modified).to eq(expected)
+        expected.each do |k, v|
```
Neilshweky, line 44: The spec describes the match pseudocode [here](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/tests/README.rst#L158):
```
 For every key/value in expected
        Assert that actual[key] MATCHES value
```

p-mongo: If it's in the spec...

p-mongo:
I think this is going to conflict with my PR so I'd rather wait here.

Neilshweky:
okay sure @p-mongo I don't mind waiting on this one.

Neilshweky:
These tests fail on JRuby in CI. I cannot reproduce locally.

PR #2641, opened 2022-09-28
RUBY-2808 add other cmap tests that work

Neilshweky:


spec/runners/cmap/verifier.rb
```diff
@@ -41,7 +41,9 @@ def verify_hashes(actual, expected)
           end
         end
 
-        expect(actual_modified).to eq(expected)
+        expected.each do |k, v|
```
p-mongo, line 44: What is in actual that is not in expected?

Neilshweky: BTW this PR depends on: https://github.com/mongodb/mongo-ruby-driver/pull/2640

Sometimes the expected specifies a `connectionId` for the `ConnectionCreated` event, and sometimes it doesn't.

comandeo (approved):

PR #2649, opened 2022-10-20
RUBY-2927 Preemptively cancel in progress operations when SDAM heartbeats timeout

Neilshweky:


spec/spec_tests/data/sdam_unified/hello-timeout.yml
```diff
@@ -156,7 +156,7 @@ tests:
           failPoint:
             configureFailPoint: failCommand
             mode:
-              times: 4
+              times: 2
```
Neilshweky, line 159: I would like to investigate this further

Neilshweky:
An explanation of the fix here: https://github.com/mongodb/mongo-ruby-driver/pull/2649/commits/f5dd980a9671c4f79ed7dfd4278783676bda0b49

What was happening was, we were getting into a state where there would be pools with open threads on disconnected clusters. Therefore, when the client was closed, the cluster wouldn't close the pools for its server, since it thinks its already disconnected. 

Now, how did we end up in a state where the cluster is disconnected but there are ready pools? There is a `:monitoring_io` option on a cluster (which is passed through to the pool) which creates a cluster with `connected?=false` and does not start the monitoring threads. The connection pool populator threads are being started, however. The solution to this is to not start the populator threads when we dont want monitoring... this causes problems with the cmap tests, however, since they need the populator thread to run in order for tests to pass. The populator thread does three things: 

1) Populates the pool to min_size
2) interrupts in use connections
3) reaps stale connections

In order to fix the cmap tests we have to manually start the populator threads, or modify the input to the connection pool constructor to not include monitoring_io...

Neilshweky:
I ended up creating a second option called `:populator_io` in order to differentiate between the creation of monitoring threads and populator threads. Note that if monitoring_io is off and populator_io is on, the populator has to be closed manually, since the cluster is still considered "closed"

lib/mongo/server/connection_pool.rb
```diff
@@ -112,6 +121,7 @@ def initialize(server, options = {})
         @options = options.freeze
```
Neilshweky, line 121:
```suggestion
        @options = options.freeze
        @options.merge!(backtrace: caller(0))
```

lib/mongo/server/connection_pool.rb
```diff
@@ -110,8 +119,10 @@ def initialize(server, options = {})
 
         @server = server
         @options = options.freeze
+        @options.merge!(backtrace: caller(0))
```
Neilshweky, line 122:
```suggestion
        @options = @options.merge(backtrace: caller(0))
```

lib/mongo/server.rb
```diff
@@ -649,14 +659,14 @@ def clear_description
     #   service id only.
     #
     # @api private
-    def clear_connection_pool(service_id: nil)
+    def clear_connection_pool(service_id: nil, interrupt_in_use_connections: false)
```
comandeo, line 662: Documenting comment for `interrupt_in_use_connections` is missing.

lib/mongo/server/connection_pool/generation_manager.rb
```diff
@@ -44,10 +46,28 @@ def generation_unlocked(service_id: nil)
...
+          @pipe_fds[service_id][@map[service_id]]
+        end
+
+        def remove_pipe_fds(generation, service_id: nil)
+          validate_service_id!(service_id)
+
+          r, w = @pipe_fds[service_id].delete(generation)
+          w.close
+          # Schedule the read end of the pipe to be closed. We cannot close it
+          # immediately since we need to wait for the Kernel#select call to
```
comandeo, line 59: 1. I think we should clarify here where the `select` call is.
2. I am not sure I understand how we wait for the `select` call to return.

lib/mongo/socket.rb
```diff
@@ -334,12 +334,36 @@ def read_from_socket(length, timeout: nil)
...
+          if fd
+            select_args = [[@socket, fd], nil, [@socket, fd], select_timeout]
+          else
+            select_args = [[@socket], nil, [@socket], select_timeout]
+          end
         else
           select_args = [nil, [@socket], [@socket], select_timeout]
         end
+
         rv = Kernel.select(*select_args)
```
comandeo, line 348: We should catch and IO error here, shouldn't we? Or is it propagated?

lib/mongo/server/connection.rb
```diff
@@ -282,6 +308,7 @@ def disconnect!(options = nil)
           @socket = nil
         end
         @closed = true
+        interrupted! if options && options[:interrupted]
```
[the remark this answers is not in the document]
Neilshweky: done

lib/mongo/server.rb
Neilshweky: done

lib/mongo/socket.rb
Neilshweky: it's propagated to the `map_exceptions` method where it's converted to a `Mongo::SocketError`

lib/mongo/error/pool_cleared_error.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+# encoding: utf-8
+
+# Copyright (C) 2019-2020 MongoDB Inc.
```
comandeo, line 4:
```suggestion
# Copyright (C) 2019-present MongoDB Inc.
```

lib/mongo/server/connection.rb
```diff
@@ -232,27 +253,35 @@ def connect!
...
-      #   returned socket.
-      private def do_connect
-        socket = add_server_diagnostics do
+      #
+      # @return [ Socket ] The created socket.
+      private def create_socket
+        add_server_diagnostics do
           address.socket(socket_timeout, ssl_options.merge(
-            connection_address: address, connection_generation: generation))
+            connection_address: address, connection_generation: generation, fd: options[:fd]))
```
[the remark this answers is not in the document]
Neilshweky: done

lib/mongo/server/connection_pool.rb
```diff
@@ -910,8 +947,10 @@ def next_available_connection(connection_global_id: nil)
       end
 
       def create_connection
+        r, _ = @generation_manager.pipe_fds(service_id: server.description.service_id)
         opts = options.merge(
           connection_pool: self,
+          fd: r
```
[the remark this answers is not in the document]
Neilshweky: done

lib/mongo/server/connection_pool.rb
```diff
@@ -966,6 +1000,48 @@ def create_and_add_connection
...
+        if conn = @available_connections.detect(&method(:connection_stale_unlocked?))
+          conn.disconnect!(reason: :stale)
+          @available_connections.delete(conn)
+          return true
+        end
+      end
+
+      # Interrupt connections scheduled for interruption.
+      def remove_interrupted_connections
+        gens = Set.new
```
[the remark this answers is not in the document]
Neilshweky: done

lib/mongo/server/connection_pool.rb
```diff
@@ -987,6 +1063,25 @@ def raise_if_closed!
...
 
+      # If the connection was interrupted, raise a pool cleared error. If it
+      # wasn't interrupted raise the original error.
+      #
+      # @param [ Connection ] The connection.
+      # @param [ Mongo::Error ] The original error.
+      #
+      # @raise [ Mongo::Error | Mongo::Error::PoolClearedError ] A PoolClearedError
+      #   if the connection was interrupted, the original error if not.
+      def raise_pool_cleared!(connection, e)
```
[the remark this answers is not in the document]
Neilshweky: done

lib/mongo/server/connection_pool.rb
```diff
@@ -1026,6 +1121,43 @@ def check_invariants
...
+            if conn.generation < @generation_manager.generation(service_id: service_id)
+              conn.disconnect!(reason: :stale, interrupted: true)
+              @populate_semaphore.signal
+              true
+            end
+          end
+        end
+      end
+
+      # Interrupt in use connections and check them back into the pool.
```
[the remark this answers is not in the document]
Neilshweky: done

lib/mongo/server/connection_pool.rb
```diff
@@ -668,17 +676,19 @@ def do_clear(options = nil)
...
+                @server.address,
+                service_id: service_id,
+                interrupt_in_use_connections: options&.[](:interrupt_in_use_connections)
+              )
             )
-          )
-          pause unless @server.load_balancer?
+            # Only pause the connection pool if the server was marked unknown,
+            # otherwise, allow the retry to be attempted with a ready pool.
+            do_pause if !@server.load_balancer? && @server.unknown?
```
Neilshweky, line 690: This is in response to the following test failure:
```
 Failure/Error: raise err
  
Mongo::Error::PoolClearedError:
   Connection to localhost:27018 interrupted due to server monitor timeout (for pool 0x856520) (modern retry, attempt 1, later retry failed: Mongo::Error::LintError: Server #<Server address=localhost:27018 PRIMARY replica_set=test-rs pool=#<ConnectionPool size=0 (0-3) used=0 avail=0 pending=0 paused>> is known but has non-ready pool)
```
We don't always mark the pool unknown for a SocketTimeoutError, so we should allow the pool to remain ready as long as the server remains known.

lib/mongo/socket.rb
```diff
@@ -334,12 +334,36 @@ def read_from_socket(length, timeout: nil)
             raise Errno::ETIMEDOUT, "Took more than #{_timeout} seconds to receive data"
           end
         end
+        pipe = options[:pipe]
```
comandeo, line 337: What happens if pipe is nil here? Will we fail with a meaningful error, or just keep working?

comandeo (approved):
Great job! :clap:

lib/mongo/socket.rb
Neilshweky: wherever a pipe is used we check for its presence. We will keep working.

lib/mongo/server/connection_pool.rb
```diff
@@ -119,6 +119,7 @@ def initialize(server, options = {})
 
         @server = server
         @options = options.freeze
+        @options = @options
```
Neilshweky, line 122:
```suggestion
```

lib/mongo/server/connection_pool.rb
```diff
@@ -612,6 +613,7 @@ def do_pause
           raise Error::LintError, "Attempting to pause pool for server #{@server.summary} which is known"
         end
 
+
```
Neilshweky, line 616:
```suggestion
```

PR #2650, opened 2022-10-21
RUBY-2496 Implement and test connection pool paused state

Neilshweky:


lib/mongo/server/connection_pool.rb
```diff
@@ -520,16 +563,36 @@ def check_in(connection)
...
+
+        return if paused?
+
+        @lock.synchronize do
+          @ready = false
+        end
+
+        stop_populator
+
+        clear(options)
```
Neilshweky, line 583: I'm not sure why we decided to automatically clear on pause here but that does not look right. I will attempt to repair this.

lib/mongo/cluster/sdam_flow.rb
```diff
@@ -188,7 +188,6 @@ def server_description_changed
 
       verify_invariants
       commit_changes
-      disconnect_servers
```
Neilshweky, line 191: I'm not sure why we were disconnecting the servers on every server description change... This was causing a lint error because on server disconnect we clear the pool, and we would occasionally clear the pool for a known server.

Neilshweky:
TODO: lb tests on retryable writes... pool not becoming unpaused.

lib/mongo/cluster.rb
```diff
@@ -497,13 +497,14 @@ def summary
     # events in the process. Stops SRV monitoring if it is active.
     # Marks the cluster disconnected.
     #
-    # @return [ true ] Always true.
+    # A closed cluster is no longer usable. If the client is reconnected,
+    # it will create a new cluster instance.
     #
-    # @since 2.1.0
-    def disconnect!
```
comandeo, line 503: Leave the existing `disconnect!` method, mark as deprecated
Add the new method `close` method, and use it when we should.

lib/mongo/monitoring/event/cmap/pool_ready.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+# encoding: utf-8
+
+# Copyright (C) 2019-2020 MongoDB Inc.
```
comandeo, line 4:
```suggestion
# Copyright (C) 2019-present MongoDB Inc.
```

lib/mongo/server.rb
```diff
@@ -260,30 +260,43 @@ def connectable?; end
     # are returned to their respective connection pools. Stop the server's
     # background monitor.
     #
-    # @return [ true ] Always true.
+    # @return [ nil ] Always nil.
```
comandeo, line 263: Do not change the return value.

lib/mongo/server/connection_pool/generation_manager.rb
```diff
@@ -45,6 +45,19 @@ def generation(service_id: nil)
           end
         end
 
+        def generation_unlocked(service_id: nil)
```
comandeo, line 48: Add `should_lock` argument to `generation` method instead of creating a whole new method. Or extract the duplicating code into a separate method.

Neilshweky: done

lib/mongo/server.rb
Neilshweky: done

lib/mongo/cluster.rb
Neilshweky: I think we can actually leave this and mark the new method @api private.... See the note on this method:

```
    # @note Applications should call Client#close to disconnect from
    # the cluster rather than calling this method. This method is for
    # internal driver use only.
```

Neilshweky: Let me know if you still want me to add it back.

comandeo (approved):

PR #2656, opened 2022-11-07
RUBY-3049 Provide access to raw result document when the server returns an error for a command

Neilshweky:
This one relies on #2655


Neilshweky:
@comandeo  this is ready for review

comandeo (approved):

PR #2662, opened 2022-11-11
RUBY-2598 Clear pool when connection establishment fails in populator

Neilshweky:


spec/runners/cmap.rb
```diff
@@ -307,6 +308,7 @@ def preprocess
 
       def configure_fail_point
         @client.database.command(@fail_point_command) if @fail_point_command
+        Mongo.broken_view_options = false if @fail_point_command
```
Neilshweky, line 311:
```suggestion
```

comandeo (approved):

PR #2664, opened 2022-11-14
RUBY-2539 Provide explicit guidance on handling command errors that occur before the handshake completes during operation execution

neilshweky:
RUBY-2539


comandeo (approved):

Neilshweky:
RUBY-2539

PR #2666, opened 2022-11-16
RUBY-2495/RUBY-2840/RUBY-2808 Rate limit new connection creations (maxConnecting)

neilshweky:
The way this currently works introduces a race condition that should be fixed in this PR.
```
check_out: 

acquire the lock:
loop:
check for available connections
check if there are maxConnections pending connections
create a connection or wait for another available connection
release the lock

Acquire the max connecting semaphore, thus indicating that there is another pending connection
connect the connection
release the semaphore
```
The race happens when two threads enter checkout simultaneously:
thread1 acquires the lock, sees that there are less than maxConnections pending connections, and creates a connection
thread1 releases the lock, but does not yet acquire the semaphore
thread2 acquires the lock, sees that there are less than maxConnections pending connections, and creates a connection
thread1 acquires the semaphore
thread2 waits for the semaphore
thread3 now checks a connection back into the pool
thread1 connects the connection and releases the semaphore
thread2 acquires the semaphore and connects its connection

thread2 should have used the connection that thread3 had just checked back in the pool, but it already created its own connection.


lib/mongo/server/connection_pool.rb
```diff
@@ -956,6 +825,23 @@ def next_available_connection(connection_global_id: nil)
         end
       end
 
+      # @return [ true | false ] if there is a connection available or there is
+      # room for another connecton.
+      def available_connection?
```
Neilshweky, line 830: probably remove this.

lib/mongo/error/connection_unavailable.rb
```diff
@@ -1,27 +0,0 @@
...
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
-module Mongo
-  class Error
-
-    # Exception raised when trying to check out a connection with a specific
-    # global id, and the connection for that global id no longer exists in the
-    # pool.
-    class ConnectionUnavailable < Error
```
comandeo, line 24: Is is a BC break? This class seems to be public.

lib/mongo/retryable.rb
```diff
@@ -131,7 +131,7 @@ def read_with_retry(session = nil, server_selector = nil, &block)
         server = select_server(cluster, server_selector, session)
         begin
           yield server
-        rescue Error::SocketError, Error::SocketTimeoutError, Error::ServerNotUsable, Error::PoolError, Error::OperationFailure => e
+        rescue Error::SocketError, Error::SocketTimeoutError, Error::ServerNotUsable, Error::ConnectionPerished, Error::PoolError, Error::OperationFailure => e
```
comandeo, line 134: This looks really ugly now, maybe we should introduce something like`Error::Retryable` mixin, and catch it here.

spec/mongo/session_transaction_spec.rb
```diff
@@ -152,7 +152,7 @@ class SessionTransactionSpecError < StandardError; end
         it 'times out' do
           start = Mongo::Utils.monotonic_time
 
-          10.times do |i|
+          11.times do |i|
```
comandeo, line 155: I am wondering why increasing this.

comandeo (commented):
Looks like LB tests fail

spec/mongo/session_transaction_spec.rb
Neilshweky: It's a long story but basically these expectations are ordered, so they're expecting exactly 11 calls to `monotonic_time`. I added a call to this method so it was throwing this test off.

lib/mongo/retryable.rb
Neilshweky: Let's file a followup ticket?

Neilshweky: RUBY-3191

lib/mongo/error/connection_unavailable.rb
Neilshweky: yeah good call

lib/mongo/retryable.rb
comandeo: 👍

comandeo (approved):
👏

PR #2671, opened 2022-11-21
RUBY-2495 rework the check_out method

Neilshweky:


lib/mongo/server/connection_pool.rb
```diff
@@ -1132,27 +994,27 @@ def check_invariants
         end
       end
 
-      # Close the connections in the given list.
+      # Close the available connections.
       #
       # @param [ Array<Connection> ] connections A list of connections.
       # @param [ Object ] service_id The service id.
-      def close_connections(connections, service_id)
+      def close_available_connections(service_id)
```
Neilshweky, line 1001: Changed this to explicitly close available connections because it was only being used on available connections, and I don't want it to be used for other connections in the future, since if it is used on pending or checked_out connections,  the size_cv has to be signaled as well.

lib/mongo/server/connection_pool.rb
```diff
@@ -689,6 +547,10 @@ def do_clear(options = nil)
             # otherwise, allow the retry to be attempted with a ready pool.
             do_pause if !@server.load_balancer? && @server.unknown?
           end
+
+          # Broadcast here to cause all of the threads waiting on the pool size
+          # pool size to decrease to break out of the wait loop and error.
```
Neilshweky, line 552:
```suggestion
          # to decrease to break out of the wait loop and error.
```

lib/mongo/condition_variable.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+    def signal
+      maybe_raise_error!
+      @cv.signal
+    end
+
+    def_delegators :@lock, :synchronize
+
+    def maybe_raise_error!
```
comandeo, line 50: This method should be private, I believe.

comandeo, line 50:
```suggestion
    def raise_unless_owned!
```

lib/mongo/server/connection_pool.rb
```diff
@@ -148,6 +148,16 @@ def initialize(server, options = {})
...
 
+        # Condition variable to enforce the first check in check_out: max_pool_size.
+        # This condition variable should be signaled when the number of
+        # unavailable connections decreases (pending + pending_connections +
+        # available_connections).
+        @size_cv = Mongo::ConditionVariable.new(@lock)
+        # This represents the number of threads that have made it past the size_cv
+        # gate but have not acquired a connection to add to the pending_connections
+        # set.
+        @pending = 0
```
comandeo, line 159:
```suggestion
        @pending_threads_count = 0
```

lib/mongo/server/connection_pool.rb
```diff
@@ -327,172 +345,10 @@ def check_out(connection_global_id: nil)
...
-
-          publish_cmap_event(
-            Monitoring::Event::Cmap::ConnectionCheckOutFailed.new(
-              @server.address,
-              Monitoring::Event::Cmap::ConnectionCheckOutFailed::CONNECTION_ERROR
-            ),
-          )
-          raise
-        end
+        raise_pool_closed! if closed?
```
comandeo, line 348:
```suggestion
        raise_if_pool_closed!
        # And move the if inside the method
```

lib/mongo/server/connection_pool.rb
```diff
@@ -327,172 +345,10 @@ def check_out(connection_global_id: nil)
...
+        raise_pool_paused! if paused?
```
comandeo, line 349:
```suggestion
        raise_if_pool_paused!
```

lib/mongo/server/connection_pool.rb
```diff
@@ -1181,6 +1047,239 @@ def clear_pending_connections
           @pending_connections.delete(connection)
         end
       end
+
+      # The lock should be acquired when calling this method.
```
comandeo, line 1051: 👍

comandeo (approved):
Great job!

lib/mongo/condition_variable.rb
Neilshweky: done

Neilshweky: I like `raise_unless_locked` better.

lib/mongo/server/connection_pool.rb
Neilshweky: I changed this to `@connection_requests` in the next PR if that's okay

lib/mongo/server/connection_pool.rb
Neilshweky: do you think it would be a good idea to raise if the lock is not owned in some of these methods?

lib/mongo/server/connection_pool.rb
comandeo: 👍

comandeo (approved):

lib/mongo/server/connection_pool.rb
comandeo: Yes, it is a very good idea!

Neilshweky: done

PR #2674, opened 2022-12-05
RUBY-2547 Add test that ensures the error returned from being evicted from the WaitQueue is retryable

Neilshweky:


Neilshweky:
@comandeo This is ready for review

comandeo (approved):

PR #2681, opened 2022-12-16
RUBY-2732 Add service account support

comandeo:


comandeo:
This also fixes https://jira.mongodb.org/browse/RUBY-3199

.evergreen/aws
```diff
@@ -45,6 +45,16 @@ when 'assume-role'
...
+when 'assume-web-role'
+  arn = ARGV.shift
+  token_file = ARGV.shift
+
+  orchestrator = AwsUtils::Orchestrator.new(**options)
+  credentials = orchestrator.assume_role_with_web_identity(arn, token_file)
+  puts "AWS_ACCESS_KEY_ID=#{credentials.access_key_id}"
+  puts "AWS_SECRET_ACCESS_KEY=#{credentials.secret_access_key}"
+  puts "AWS_SESSION_TOKEN=#{credentials.session_token}"
+  puts
```
jamis, line 57: I know it implies a larger refactoring here, but what if the bodies of each of these `when` clauses were refactored to separate methods? It would make the enclosing `case` statement more concise, which (for me, selfishly, as a newbie 😄) would be a win, since it would be easier to scan and see what the various intentions are.

lib/mongo/auth/aws/credentials_retriever.rb
```diff
@@ -188,6 +194,48 @@ def ecs_metadata_credentials
           return nil
         end
 
+        def assume_role_with_wed_identity_credentials
```
jamis, line 197: I know I'm the new guy, so forgive me if my suggestions are overstepping any bounds, but since I can't really comment (yet) on the functionality of the PR, I figured I might as well share my thoughts on style and architecture. :)

Especially as the new guy, looking at this method is a bit challenging; it requires careful picking apart to see what's going on and what the different pieces are being used for. However, there are independent sections of this method that can be extracted into auxiliary methods, which can make the original method more concise, and clarify its intent.

For example:

```ruby
def assume_role_with_web_identity_credentials
  web_identity_token, role_session_name = prepare_web_identity_inputs
  return if web_identity_token.nil?

  response = request_web_identity_role(web_identity_token, role_session_name)
  return if response.nil?

  credentials_from_web_identity_response(response)
end
```

jamis (approved):
Am I right in understanding that this PR is in response to a specification change? I tried reading the associated tickets in Jira, and I think I got the gist of what is happening here, but it's a lot to take in. :) Thanks for letting me "look over your shoulder", here! And please do let me know if I'm overstepping at all with my comments.

lib/mongo/auth/aws/credentials_retriever.rb
comandeo: Great idea, thank you! Changed accordingly.

.evergreen/aws
comandeo: Done

PR #2682, opened 2023-01-05
RUBY-3191 Abstract retryability of errors

jamis:
This adds a new private method, `retryable_exceptions`, that enumerates those exceptions that are (in most cases) retryable. The method then replaces the explicit lists of exceptions in various rescue statements.

~I also noticed multiple instances of duplication in the code and took the opportunity to try and tidy that up, but realized as I was doing so that this probably expands the scope of the ticket unnecessarily. I can back those changes out if needed and resubmit the refactoring as a separate PR.~

I removed the other refactorings from the original PR and will resubmit them separately.

### DIFF
--- a/lib/mongo/retryable.rb
+++ b/lib/mongo/retryable.rb
@@ -336,40 +336,88 @@ module Mongo
     # @api private
     def legacy_write_with_retry(server = nil, context:)
       session = context.session
-      attempt = attempt ? attempt + 1 : 1
 
       # This is the pre-session retry logic, and is not subject to
       # current retryable write specifications.
       # In particular it does not retry on SocketError and SocketTimeoutError.
-      server ||= select_server(cluster, ServerSelector.primary, session)
-      server.with_connection(connection_global_id: context.connection_global_id) do |connection|
-        # Legacy retries do not use txn_num
-        yield connection, nil, context.dup
+      attempt = 0
+      begin
+        attempt += 1
+        server ||= select_server(cluster, ServerSelector.primary, session)
+        server.with_connection(connection_global_id: context.connection_global_id) do |connection|
+          # Legacy retries do not use txn_num
+          yield connection, nil, context.dup
+        end
+      rescue Error::OperationFailure => e
+        e.add_note('legacy retry')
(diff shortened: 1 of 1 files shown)


lib/mongo/retryable.rb
```diff
@@ -548,5 +496,51 @@ def raise_unsupported_error(e)
       new_error.set_backtrace(e.backtrace)
       raise new_error
     end
+
+    def retryable_exceptions
```
comandeo, line 500: We try to add docstring to all methods. For private methods the docstring do not have to be long and detailed as for public ones, but a short description would be nice.

lib/mongo/retryable.rb
```diff
@@ -548,5 +496,51 @@ def raise_unsupported_error(e)
...
+        end
+      else
+        raise exception
+      end
+
+      log_retry(exception, message: 'Legacy read retry')
+      sleep(client.read_retry_interval) unless is_retryable_exception?(exception)
+    end
+
+    def prepare_to_retry_legacy_write_with_retry(exception, attempt)
```
comandeo, line 533: According to its usage - https://github.com/mongodb/mongo-ruby-driver/pull/2682/files?diff=split&w=1#diff-d01924119d9e1f082918101e83057a7fac7ff818e8126320fec1714d308b73f0R350 - this method should return `server`. However, if I am not wrong, it either return `nil`, or raise.
I believe that the idea is that returning `nil` forces a server selection; this seems to be a bit obscure.

lib/mongo/retryable.rb
jamis: Good to know! I saw many of the private methods in that file didn't have docstrings, and tried to follow that pattern. I was obviously led astray. :) I'll get that fixed.

lib/mongo/retryable.rb
jamis: Yeah, I could document this better. I considered making it return nothing, and explicitly setting `server` to `nil` at the callsite, but I figured I could kill two birds with one stone here. :) I was probably trying to be too clever. I'll clarify.

comandeo (approved):
LGTM!

PR #2683, opened 2023-01-05
Mongo::Retryable refactoring

jamis:
This refactors the `Mongo::Retryable` module with a few aims:

1. The Retryable module was responsible for both reading and writing, which are orthogonal concerns. Separating them into separate delegates allows them to be considered and maintained separately, and makes them easier to reason about.
2. Many methods were quite long. I've extracted portions of them to new methods, trying to keep each method to no more than a single screenful of code. (Naturally, this limit is subjective; I've used my own judgement and preferences here.)
3. There was significant duplication in some rescue clauses. I've combined those clauses where possible, and where it didn't obfuscate the original intent.
4. Other minor textual changes (ternary operator, safe navigation operator, etc.) that seemed appropriate to me.

None of these changes affect the public-facing API.


lib/mongo/retryable/base_worker.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+# encoding: utf-8
+
+# Copyright (C) 2015-2020 MongoDB Inc.
```
comandeo, line 4:
```suggestion
# Copyright (C) 2015-2023 MongoDB Inc.
```
Or just omit the end year.

lib/mongo/retryable/base_worker.rb
```diff
@@ -0,0 +1,90 @@
...
+# limitations under the License.
+
+module Mongo
+  module Retryable
+
+    # The abstract superclass for workers employed by Mongo::Retryable.
+    #
+    # @api private
+    #
+    # @since 2.19.0
```
comandeo, line 25: We do not use `since` tag in our new code. Sorry that I forgot to tell you earlier.

lib/mongo/retryable/base_worker.rb
```diff
@@ -0,0 +1,90 @@
...
+    class BaseWorker
```
comandeo, line 26: I assume that this class is not part of our public API. In this case we should mark it with `@api private`.

lib/mongo/retryable/base_worker.rb
```diff
@@ -0,0 +1,90 @@
...
+      #
+      # @param [ Mongo::Retryable ] retryable The client object that is using
+      #   this worker to perform a retryable operation
+      def initialize(retryable)
+        @retryable = retryable
+      end
+
+      private
+
+      # Returns a list of exception classes that are generally retryable.
```
comandeo, line 54:
```suggestion
      # @return [ Array<Mongo::Error> ] Array of exception classes that are generally retryable. 
```
I know it contradicts the Rails style, but I believe it is beneficial to add type information to the documenting comments - `@param` and `@return` in particular. While it is not required for the private methods, it is still helpfult for tools like [solargraph](https://solargraph.org/guides/yard)

lib/mongo/retryable/base_worker.rb
```diff
@@ -0,0 +1,90 @@
...
+          Error::ServerNotUsable,
+          Error::SocketError,
+          Error::SocketTimeoutError
+        ].freeze
+      end
+
+      # Tests to see if the given exception instance is of a type that can
+      # be retried.
+      #
+      # @return [ Boolean ] true if the exception is retryable.
```
comandeo, line 67:
```suggestion
      # @return [ true | false ] true if the exception is retryable.
```
We agreed not to use `Boolean` since such type does not exist in Ruby.

However, YARD (as well as Sorbet and RBS) allows and encourages using `Boolean`, so we may want to re-discuss this topic.

lib/mongo/retryable/read_worker.rb
```diff
@@ -0,0 +1,295 @@
...
+      # retry_reads: false and set :max_read_retries to 0 on the client.
+      #
+      # @api private
+      #
+      # @example Execute the read.
+      #   read_with_retry(session, server_selector) do |server|
+      #     ...
+      #   end
+      #
+      # @param [ Mongo::Session ] session The session that the operation is
```
comandeo, line 106:
```suggestion
      # @param [ Mongo::Session | nil ] session The session that the operation is
```

lib/mongo/retryable/read_worker.rb
```diff
@@ -0,0 +1,295 @@
...
+      #   being run on.
+      # @param [ Mongo::ServerSelector::Selectable ] server_selector Server
```
comandeo, line 108:
```suggestion
      # @param [ Mongo::ServerSelector::Selectable | nil ] server_selector Server
```

lib/mongo/retryable/read_worker.rb
```diff
@@ -0,0 +1,295 @@
...
+      # @api private
+      #
+      # @example Execute the read.
+      #   read_with_one_retry do
+      #     ...
+      #   end
+      #
+      # @note This only retries read operations on socket errors.
+      #
+      # @param [ Hash ] options Options.
```
comandeo, line 140:
```suggestion
      # @param [ Hash | nil ] options Options.
```

lib/mongo/retryable/read_worker.rb
```diff
@@ -0,0 +1,295 @@
...
+      # @yield Calls the provided block with no arguments
+      #
+      # @option options [ String ] :retry_message Message to log when retrying.
```
comandeo, line 143: I think it a bit easier to read when options are grouped with the option param.

lib/mongo/retryable/write_worker.rb
```diff
@@ -0,0 +1,366 @@
...
+      #
+      # @param [ nil | Hash | WriteConcern::Base ] write_concern The write concern.
+      # @param [ Context ] context The context for the operation.
+      #
+      # @yieldparam [ Connection ] connection The connection through which the
+      #   write should be sent.
+      # @yieldparam [ nil ] txn_num nil as transaction number.
+      # @yieldparam [ Operation::Context ] context The operation context.
+      #
+      # @api private
```
comandeo, line 104: This tag is unnecessary here since the whole class is marked with `api private`.

lib/mongo/retryable/base_worker.rb
jamis: I've been writing Ruby code for more than 20 years, and I *still* forget that there's no Boolean superclass in the language. :)

PR #2689, opened 2023-02-06
RUBY-3005 Support Range Indexes for auto encryption

comandeo:
This PR adds auto encryption spec tests for range indexes.
https://jira.mongodb.org/browse/RUBY-3005


lib/mongo/database/view.rb
```diff
@@ -197,7 +200,16 @@ def initial_query_op(session, options = {})
       end
 
       def send_initial_query(server, session, options = {})
```
jamis, line 202: I know this method predates your PR, but is it worth adding a comment to document it? For me, at least, it's not clear without a lot of digging what the method's intent is.

jamis (approved):

lib/mongo/database/view.rb
comandeo: Good hint, thank you! Added the comment.

PR #2696, opened 2023-03-05
[READY FOR REVIEW] Add support for Ruby 3.2

johnnyshields:
This PR adds CI coverage for Ruby 3.2. There are several places in the `/spec/` files that need to have block keyword arg usage adjusted, but nowhere in the `/lib/` dir.

It also removes the `rspec-mocks-diag` hack gem in favor of standard `rspec`. `rspec-mocks-diag` does not support Ruby 3.2 kwargs as it is based on a stale version of rpsec-mocks. See MONGOID-5581.

Spec failures appear to be at parity with the current `master` branch, possibly some transient failures happening.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -724,6 +724,10 @@ axes:
   - id: "ruby"
     display_name: Ruby Version
     values:
+      - id: "ruby-3.2"
+        display_name: ruby-3.2
+        variables:
+          RVM_RUBY: "ruby-3.2"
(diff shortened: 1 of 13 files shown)


johnnyshields:
Spec failures appear to be transient.

jamis:
This issue has been resolved as part of RUBY-3232 (a broader effort to generally modernize our build infrastructure and move away from older build hosts). Ruby 3.2 is now officially supported. Thanks for the nudge to get it done!

PR #2712, opened 2023-05-02
RUBY-3232 Switch to supported build hosts

jamis:
This bumps the spec/shared submodule to pick up the latest toolchain version, and attempts to clean up and reorganize the evergreen spec matrices to reduce duplication, clarify the purpose of the tests, make it easier to keep the tests up-to-date with newer Ruby and database versions, and (of course) to use the more recent supported build hosts.

Ref: https://jira.mongodb.org/browse/RUBY-3232


.evergreen/run-tests.sh
```diff
@@ -47,6 +47,12 @@ if test "$DOCKER_PRELOAD" != 1; then
   install_mlaunch_venv
 fi
 
+# Make sure cmake is installed (in case we need to install the libmongocrypt
+# helper)
+if test -n "$FLE"; then
```
comandeo, line 52:
```suggestion
if [ "$FLE" = "helper" ]; then
```
Strictly speaking, we need cmake only if we use the helper gem. If we use the pre-compiled library, cmake is not needed.

comandeo (approved):

PR #2717, opened 2023-05-19
RUBY-2748 Retry reads/writes on another mongos

comandeo:


lib/mongo/retryable.rb
```diff
@@ -46,8 +46,8 @@ module Retryable
     # @api private
     #
     # @return [ Mongo::Server ] A server matching the server preference.
-    def select_server(cluster, server_selector, session)
-      server_selector.select_server(cluster, nil, session)
+    def select_server(cluster, server_selector, session, failed_server = nil)
+      server_selector.select_server(cluster, nil, session, deprioritized: [failed_server].compact)
```
jamis, line 50: If you're just wanting to squeeze a possible nil out of a single element array, another idiom I find a little simpler is to use the `Array` initializer:

~~~ruby
> Array(nil)
[]
> Array(5)
[5]
> Array([5])
[5]
~~~

Ultimately it's personal preference, though -- `[failed_server].compact` is fine, too.

jamis (approved):
This looks great! Very easy to follow, and the tests are set up well, too. 👍 👍

PR #2722, opened 2023-05-31
RUBY-3152 Collect FaaS platform info in handshake

jamis:
This PR addresses DRIVERS-2209 and DRIVERS-2623. It adds a new (optional) `:env` key to the client document in the handshake, describing the current FaaS environment (if any). This PR also updates the logic to truncate the client document, to keep it under the maximum document size of 512 bytes. 

(Another PR is forthcoming to update the touched files to satisfy Rubocop.)

Ref: https://jira.mongodb.org/browse/RUBY-3152

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -64,6 +64,9 @@ RSpec/BeforeAfterAll:
 RSpec/DescribeClass:
   Enabled: false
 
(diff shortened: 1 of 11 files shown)


comandeo (approved):
I really like the Truncator refactoring!

PR #2730, opened 2023-06-12
RUBY-3252 Show user friendly error if no ffi

comandeo:
https://jira.mongodb.org/browse/RUBY-3252

### DIFF
--- a/spec/mongo/crypt_spec.rb
+++ b/spec/mongo/crypt_spec.rb
@@ -1,0 +1,27 @@
+require 'lite_spec_helper'
+
+describe Mongo::Crypt do
+  describe '.validate_ffi!' do
+    context 'when ffi is available' do
+      context 'when ffi is loaded' do
+        it 'does not raise' do
+          expect do
+            described_class.validate_ffi!
+          end.not_to raise_error
+        end
+      end
+
+      context 'when ffi is not loaded' do
+        before do
+          Object.send(:remove_const, :FFI)
+        end
+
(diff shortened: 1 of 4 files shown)


alexbevi (approved):

alexbevi:
Changes and error message lgtm

spec/mongo/crypt_spec.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Mongo::Crypt do
+  describe '.validate_ffi!' do
+    context 'when ffi is available' do
```
jamis, line 7: Should there be something here to test for the cases where FFI is not installed? Or skip when it isn't?

jamis (changes requested):
Elegant solution! Just the one comment about the specs.

spec/mongo/crypt_spec.rb
comandeo: I couldn't come up with a reasonable complex solution for testing this case. We have gem 'ffi' in our standard set of dependencies for testing. So, we would need a dedicated configuration just to test this feature; I think this is an overhead.

jamis: Ah! That makes sense. I'll trust that you've thought about this a lot more than I have. :) Thanks for the explanation. Maybe a comment in that spec, explaining the situation, would be helpful to future-us?

jamis (approved):

PR #2731, opened 2023-06-12
RUBY-3262 enable QEv2 tests on Serverless

jamis:
Syncs the fle2v2-* specs with the commit indicated in the ticket. We do not appear to be skipping serverless anywhere in the prose tests, so nothing else needed to be updated.

This also addresses RUBY-3270, effectively just removing a bandaid I added several weeks ago to `spec/support/utils.rb` to remove `nil` values from the expected values.

ref: https://jira.mongodb.org/browse/RUBY-3262, https://jira.mongodb.org/browse/RUBY-3270


comandeo (changes requested):
We should also add these specs to the set we run on serverless -- https://github.com/mongodb/mongo-ruby-driver/blob/master/.evergreen/run-tests-serverless.sh#L95

jamis:
Oh, good catch @comandeo! I totally missed that the serverless tests were run specifically. I'll get that updated.

PR #2738, opened 2023-07-05
RUBY-3205 Deprecate collStats

comandeo:


lib/mongo/collection.rb
```diff
@@ -314,7 +314,9 @@ def with(new_options)
     #
     # @since 2.0.0
     def capped?
-      database.read_command(:collstats => name).documents[0][CAPPED]
+      database.list_collections(filter: { name: name })
+        .first
+        &.dig('options', 'capped') || false
```
jamis, line 319: The `CAPPED` constant is still present; is it worth using that here? If not, it should probably be removed, too.

jamis (approved):

lib/mongo/collection.rb
comandeo: Strictly speaking, it's a part of our public API, so we cannot remove it in a minor version. Should we deprecate it and mark for removal in 3.0?

jamis: In that case, maybe we just use it here, instead of the `'capped'` literal in the `#dig` call?

```suggestion
        &.dig('options', CAPPED) || false
```

PR #2739, opened 2023-07-07
RUBY-3284 Enable max_connecting client option

comandeo:


lib/mongo/server/connection_pool.rb
```diff
@@ -36,6 +36,10 @@ class ConnectionPool
       # @since 2.9.0
       DEFAULT_MIN_SIZE = 0.freeze
 
+      # The default maximum number of connections that can be connecting at
+      # any given time.
+      DEFAULT_MAX_CONNECTING = 2.freeze
```
jamis, line 41: Freezing an immutable constant doesn't really do anything, does it? My understanding is that the primary purpose of freezing constants is to prevent them from being modified in place. Is there another reason for freezing numbers (which are already immutable)?

spec/mongo/uri/srv_protocol_spec.rb
```diff
@@ -1018,6 +1018,18 @@
         include_examples "roundtrips string"
       end
 
+      context 'when providing maxConnecting' do
+
+        let(:max_connecting) { 10 }
+        let(:options) { "maxConnecting=#{max_connecting}" }
+
+        it 'sets the max pool size option' do
```
jamis, line 1026: Should this read `'sets the max connecting option'`?

jamis (approved):
Just a question, and a copy/paste artifact. Otherwise, looks great!

lib/mongo/server/connection_pool.rb
comandeo: Good point, thank you. Fixed.

lib/mongo/client.rb
```diff
@@ -266,6 +267,12 @@ def hash
     # @option options [ String ] :log_prefix A custom log prefix to use when
     #   logging. This option is experimental and subject to change in a future
     #   version of the driver.
+    # @option options [ Integer ] :max_connecting The maximum number of
+    #  connections that can be connecting simultaneously. The default is 2.
+    #  This option should be increased if there are many threads that share
+    #  the same client and the application is experiencing timeouts
+    #  while waiting for connections to be established.
+    # selecting a server for an operation. The default is 100.
```
alexbevi, line 275: "... The default is 100"

Is that a copy/paste artifact?

lib/mongo/error/invalid_max_connecting.rb
```diff
@@ -0,0 +1,30 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module Mongo
+  class Error
+    # Exception that is raised when trying to create a client with an invalid
+    #   max_connecting option.
+    #
+    # @since 2.4.2
```
alexbevi, line 22: Should this be 2.19.0 (not 2.4.2)?

lib/mongo/client.rb
comandeo: It is indeed, good catch! Thank you, fixed.

lib/mongo/error/invalid_max_connecting.rb
comandeo: This is a copy/paste artifact, too. We do not use @since tag at all now. Removed, thank you.

PR #2768, opened 2023-08-01
RUBY-3209 Generate api docs

comandeo:


jamis (approved):
Looks good! Probably should just see why rubocop is complaining (probably because it wants the frozen_string_literal comment at the top of the file...).

comandeo:
> Looks good! Probably should just see why rubocop is complaining (probably because it wants the frozen_string_literal comment at the top of the file...).

Oh, thanks! Rubocop stopped working in RubyMine after I upgraded to the latest version, so I overlooked it.

PR #2774, opened 2023-08-22
RUBY-3316 Quality of life updates for BSON benchmarks

jamis:
The changes here are not mandated by the spec, but make the benchmarks easier to run and monitor. The two most significant changes here are:

1. **Progress reporting.** The benchmarks will (by default) write to the terminal as the benchmarks progress, allowing you to see that yes, something is indeed happening. You can disable the output by setting the `PROGRESS` environment variable to 0.
2. **Bulk execution.** You can run multiple benchmarks in a single invocation, now. For example, `rake benchmark:bson` will run all BSON benchmarks, and `rake benchmark:bson:flat` will run all flat BSON benchmarks.

This PR also refactors the Rake benchmark tasks to split them out into separate files.


alexbevi (approved):
I really like the refactor! LGTM

PR #2784, opened 2023-09-21
DOCSP-32949: docs link to the BSON 4.x Tutorial

ccho-mongodb:
Update the link BSON v4 tutorial based on the new location.

[JIRA](https://jira.mongodb.org/browse/DOCSP-32949)

Note:
Docs Platform mentioned the `:doc:` directive should work since the file is pulled in prior to the build, but not with 100% certainty. Therefore, if it does not work as intended, a followup PR will be created to change the link format used to refer to absolute URLs.


alexbevi:
> Docs Platform mentioned the :doc: directive should work since the file is pulled in prior to the build, but not with 100% certainty. Therefore, if it does not work as intended, a followup PR will be created to change the link format used to refer to absolute URLs.

@ccho-mongodb is there no way to verify this _before_ we sign off on a PR? Can the docs/docs-platform team fork the driver and validate this works so we can more confidently accept the changes?

alexbevi (approved):
ok, LGTM - if @jamis agrees we'll merge this

jamis (approved):
Looks good to me as well!

PR #2790, opened 2023-10-06
RUBY-3331: Inline server lifecycle callout

alexbevi:


alexbevi:
@ccho-mongodb I just copy/pasted from 10gen/docs-shared/blob/main/dbx/lifecycle-schedule-callout.rst

ccho-mongodb:
> @ccho-mongodb I just copy/pasted from 10gen/docs-shared/blob/main/dbx/lifecycle-schedule-callout.rst

LGTM!

jamis (approved):

PR #2793, opened 2023-10-13
RUBY-3332 Fix tailable cursors

comandeo:


spec/integration/find_options_spec.rb
```diff
@@ -187,4 +202,25 @@
...
+  describe 'cursor type' do
+    let(:collection_options) do
+      { capped: true, size: 1000 }
+    end
+
+    context 'when cursor type is :tailable' do
+      it 'sets the cursor type to tailable' do
+        collection.find({}, cursor_type: :tailable).first
+        expect(find_command.command['tailable']).to be true
+      end
```
jamis, line 215: Maybe worth checking that `awaitData` is nil/false here? Otherwise, the test would still pass even if `:tailable_await` were given, instead of `:tailable`.

jamis (approved):
Just one minor suggestion on the tests. Otherwise, LGTM!

spec/integration/find_options_spec.rb
comandeo: Good point, thank you! Added.

PR #2794, opened 2023-10-17
ruby2.7 Using the last argument as keyword parameters is deprecated

ilkarataev:
Hi I faced problem on my script I saw warnings:  
/var/lib/gems/2.7.0/gems/mongo-2.19.1/lib/mongo/server.rb:658: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/var/lib/gems/2.7.0/gems/mongo-2.19.1/lib/mongo/server/description.rb:220: warning: The called method `initialize' is defined here

This small changes fix the problem.

### DIFF
--- a/lib/mongo/server.rb
+++ b/lib/mongo/server.rb
@@ -655,7 +655,7 @@ module Mongo
     #
     # @api private
     def clear_description
-      @description = Mongo::Server::Description.new(address, {})
+      @description = Mongo::Server::Description.new(address, **{})
     end
 
     # @param [ Object ] :service_id Close connections with the specified


jamis:
Thanks for pointing this out! It looks like this is actually a problem with Ruby 2.7 and its parser; the same warning does not appear in Ruby 3.x. The call as given should be fine, because the empty hash is supposed to map to the `config` parameter, which is not a keyword argument. Ruby 2.7 is apparently trying to connect the empty hash to the subsequent keyword arguments, though.

Ultimately, the empty hash is redundant because the `Description` initializer has default values for all of those. Rather than tag the hash as keyword arguments, I think it might be better to just remove it from that call altogether:

```ruby
Mongo::Server::Description.new(address)
```

If we do want to preserve the empty hash as a hint that we're truly producing an empty server description, something like the following would be more accurate:

```ruby
Mongo::Server::Description.new(address, {}, **{})
```

/cc @comandeo -- do you have any opinions here?

comandeo:
I second what @jamis wrote above. I believe we should use just `Mongo::Server::Description.new(address)` here; it should be clear enough from the context that we create an empty description.

PR #2800, opened 2023-10-19
RUBY-3149 Test on AWS Lambda

comandeo:
https://jira.mongodb.org/browse/RUBY-3149
https://jira.mongodb.org/browse/RUBY-3283


.evergreen/run-deployed-lambda-aws-tests.sh
```diff
@@ -0,0 +1,117 @@
...
+cleanup ()
+{
+  delete_lambda_function
+}
+
+trap cleanup EXIT SIGHUP
+
+cd "${TEST_LAMBDA_DIRECTORY}"
+
+sam build --use-container
```
jamis, line 85: ❤️

spec/faas/ruby-sam-app/mongodb/app.rb
```diff
@@ -0,0 +1,115 @@
...
+def average_command_time
+  if $commands_count == 0
+    0
+  else
+    $total_command_time / $commands_count
+  end
+end
+
+def lambda_handler(event:, context:)
+  db = $client.use('lambda_test')
```
jamis, line 95: Does `$client` need to be a global, here? Is there any reason not to instantiate the client within the handler function?

spec/faas/ruby-sam-app/mongodb/app.rb
```diff
@@ -0,0 +1,115 @@
...
+  client.subscribe(Mongo::Monitoring::CONNECTION_POOL, pool_monitor)
+end
+
+$client = Mongo::Client.new(ENV['MONGODB_URI'], sdam_proc: sdam_proc)
+
+puts 'Connecting'
+$client.use('lambda_test').database.list_collections
+puts 'Connected'
+
+def reset_counters
```
jamis, line 70: The global variables make me think that maybe this should be encapsulated in an object, like a `FaaSRunner` class or something.

spec/faas/ruby-sam-app/mongodb/app.rb
comandeo: It is [recommended](https://www.mongodb.com/docs/atlas/manage-connections-aws-lambda/) to have a global client instance:
> Don't define a new MongoClient object each time you invoke your function. Doing so causes the driver to create a new database connection with each function call. This can be expensive and can result in your application exceeding database connection limits.

So, we need some kind of the global object here. We can work it around, i.e. using a singleton, but I'd keep it simple and dumb :-)

.evergreen/config/common.yml.erb
```diff
@@ -510,6 +514,52 @@ task_groups:
     tasks:
       - test-full-atlas-task
```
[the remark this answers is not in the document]
comandeo: Oh, that is a great idea! Updated accordingly.

spec/faas/ruby-sam-app/mongodb/app.rb
comandeo: Got rid of most of the global variables, except `$client` and `$stats_aggregator`. I think it is cleaner now, thank you!

spec/faas/ruby-sam-app/mongodb/app.rb
jamis: Ah! I didn't know about that constraint. Never mind me, then. 👍 Looks good!

jamis (approved):
Looks great!

PR #2821, opened 2024-01-09
RUBY-3298 Add container info to handshake metadata

jamis:
Adds a `:container` key to the environment document that is set with the handshake metadata. The container may include a `:runtime` key (for Docker) and/or an `:orchestrator` key (for Kubernetes).


lib/mongo/server/app_metadata/environment.rb
```diff
@@ -192,6 +211,32 @@ def detect_environment
...
+          if runtime || orchestrator
+            fields[:container] = {}
+            fields[:container][:runtime] = runtime if runtime
+            fields[:container][:orchestrator] = orchestrator if orchestrator
+          end
+        end
+
+        # Checks for the existence of a Dockerfile in the working directory.
+        def docker_present?
+          File.exist?('Dockerfile')
```
comandeo, line 231: I believe that we should check for the presence of `.dockerenv` here - https://github.com/mongodb/specifications/pull/1454/files#diff-0ecfac0976ff0bfe5b91b5bd52bc5a79dd1125434505ff22331c7cadf1b2f25eR379

comandeo (changes requested):

comandeo (approved):

PR #2822, opened 2024-01-11
RUBY-1791 Raise if transactions not supported

comandeo:


lib/mongo/error/transactions_not_supported.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+# rubocop:todo all
```
jamis, line 2: For new files, I think we should probably apply the linter from the beginning.

jamis (changes requested):
Looks good, just the one comment about rubocop. 🤷

lib/mongo/error/transactions_not_supported.rb
comandeo: Good catch, thanks!

jamis (approved):

PR #2825, opened 2024-01-12
RUBY-1813 Discard ServerSessions involved in network errors

jamis:
Add a `dirty` flag to `ServerSession`, and set it when a network error occurs. When checking the session back into the session pool, discard the session if it was marked dirty.

ref: RUBY-1813, DRIVERS-649, also RUBY-3000, RUBY-3031

### DIFF
--- a/lib/mongo/session/server_session/dirtyable.rb
+++ b/lib/mongo/session/server_session/dirtyable.rb
@@ -43,7 +43,6 @@ module Mongo
         #
         # @param [ true | false ] mark whether the mark the server session
         #   dirty or not.
-        #   or not.
         def dirty!(mark = true)
           @dirty = mark
         end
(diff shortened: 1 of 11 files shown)


lib/mongo/session/server_session/dirtyable.rb
```diff
@@ -0,0 +1,53 @@
...
+        #
+        # @return [ true | false ] the server session's dirty state
+        def dirty?
+          @dirty
+        end
+
+        # Mark the server session as dirty (the default) or clean.
+        #
+        # @param [ true | false ] mark whether the mark the server session
+        #   #dirty
```
comandeo, line 45:
```suggestion
        #   dirty or not.
```

lib/mongo/session/server_session/dirtyable.rb
```diff
@@ -0,0 +1,53 @@
...
+        #   or not.
```
comandeo, line 46:
```suggestion
```

comandeo (approved):

PR #2829, opened 2024-01-24
Fix failing atlas connectivity specs

jamis:
The Atlas connectivity specs were failing when testing the "free tier" atlas URI, because the "ismaster" command wasn't returning the expected results.

Looking at the Go and Python drivers, it looks like they are only testing for the absence of an error when executing these commands; the original ticket (https://jira.mongodb.org/browse/DRIVERS-382) also stipulates that drivers should merely "verify that no error occurred." I've updated the specs so they no longer check for the contents of the result, and only confirm that no error occurred.

### DIFF
--- a/spec/atlas/atlas_connectivity_spec.rb
+++ b/spec/atlas/atlas_connectivity_spec.rb
@@ -11,13 +11,13 @@ describe 'Atlas connectivity' do
 
   describe 'connection to Atlas' do
     it 'runs ismaster successfully' do
-      result = client.database.command(:ismaster => 1)
-      expect(result.documents.first['ismaster']).to be true
+      expect { client.database.command(:ismaster => 1) }
+        .not_to raise_error
     end
 
     it 'runs findOne successfully' do
-      result = client.use(:test)['test'].find.to_a
-      expect(result).to be_a(Array)
+      expect { client.use(:test)['test'].find.to_a }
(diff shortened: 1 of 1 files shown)


spec/atlas/atlas_connectivity_spec.rb
```diff
@@ -11,13 +11,13 @@
 
   describe 'connection to Atlas' do
     it 'runs ismaster successfully' do
-      result = client.database.command(:ismaster => 1)
-      expect(result.documents.first['ismaster']).to be true
+      expect { client.database.command(:ismaster => 1) }
```
comandeo, line 14:
```suggestion
      expect { client.database.command(:hello => 1) }
```

comandeo: `ismaster` is deprecated, we should use `hello` instead.

comandeo (changes requested):

spec/atlas/atlas_connectivity_spec.rb
jamis: I wasn't sure how closely we ought to mimic the other drivers; Go & Python are both still using `ismaster` here, but probably just because it's never been revisited. I'm okay switching to `hello`.

comandeo (approved):

PR #2837, opened 2024-02-14
RUBY-3368 Measure minimum RTT

comandeo:


lib/mongo/server/round_trip_time_calculator.rb
```diff
@@ -61,16 +71,21 @@ def measure
...
         @average_round_trip_time = if average_round_trip_time
           RTT_WEIGHT_FACTOR * last_round_trip_time + (1 - RTT_WEIGHT_FACTOR) * average_round_trip_time
         else
           last_round_trip_time
         end
       end
+
+      def update_minimum_round_trip_time
+        @rtts.push(@last_round_trip_time) unless @last_round_trip_time.nil?
+        @minimum_round_trip_time = 0 and return if @rtts.size < 3
```
jamis, line 84: Should this be `MIN_SAMPLES` instead of a literal `3`?

jamis (approved):
Looks great! Just one comment, but it's good to merge otherwise.

PR #2843, opened 2024-02-26
Add write timeout

comandeo-mongo:


lib/mongo/socket.rb
```diff
@@ -265,18 +263,86 @@ def connectable?
...
+          if socket_timeout <= 0
+            raise Mongo::Error::TimeoutError
+          end
+          chunk = read_from_socket(length - data.length, socket_timeout: socket_timeout, csot: true)
+          unless (chunk.length > 0 || length == 0)
+            raise IOError, "Expected to read > 0 bytes but read 0 bytes"
+          end
+          data << chunk
+        end
+        data
```
jamis, line 295: If `data` is initialized to the empty string, could we get rid of the duplication here by letting the loop do all the heavy lifting?

```suggestion
        "".tap do |data|
          while data.length < length
            socket_timeout = deadline - Utils.monotonic_time
            if socket_timeout <= 0
              raise Mongo::Error::TimeoutError
            end
            chunk = read_from_socket(length - data.length, socket_timeout: socket_timeout, csot: true)
            unless chunk.length > 0
              raise IOError, "Expected to read > 0 bytes but read 0 bytes"
            end
            data << chunk
          end
        end
```

lib/mongo/socket.rb
```diff
@@ -265,18 +263,86 @@ def connectable?
...
+          raise IOError, "Expected to read > 0 bytes but read 0 bytes"
+        end
+        while data.length < length
+          chunk = read_from_socket(length - data.length, socket_timeout: socket_timeout)
+          unless (chunk.length > 0 || length == 0)
+            raise IOError, "Expected to read > 0 bytes but read 0 bytes"
+          end
+          data << chunk
+        end
+        data
```
jamis, line 320: I know this was just copied from the original implementation of `#read`, but we can probably get rid of the duplication as was suggested above for `read_with_timeout`, by letting the loop do all the work:

```suggestion
        "".tap do |data|
          while data.length < length
            chunk = read_from_socket(length - data.length, socket_timeout: socket_timeout)
            unless chunk.length > 0
              raise IOError, "Expected to read > 0 bytes but read 0 bytes"
            end
            data << chunk
          end
        end
```

jamis (changes requested):

lib/mongo/protocol/message.rb
```diff
@@ -245,8 +245,10 @@ def self.deserialize(io,
         # method with some other IO-like object, pass options only when they
         # are not empty.
         read_options = {}
-        if timeout = options[:socket_timeout]
-          read_options[:timeout] = timeout
+        %i[timeout socket_timeout].each do |key|
+          if val = options[key]
+            read_options[key] = val
+          end
```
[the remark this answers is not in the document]
comandeo-mongo: I think it can be simplified even more: `read_options = options.slice(:timeout, :socket_timeout)`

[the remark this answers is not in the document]
jamis: Good call! I missed that `read_options` is a new Hash. 👍

lib/mongo/protocol/message.rb
```diff
@@ -244,9 +244,11 @@ def self.deserialize(io,
         # timeout option. For compatibility with whoever might call this
         # method with some other IO-like object, pass options only when they
         # are not empty.
-        read_options = {}
-        if timeout = options[:socket_timeout]
-          read_options[:timeout] = timeout
+        read_options = options.slice(:timeout, :socket_timeout)
+        %i[timeout socket_timeout].each do |key|
```
jamis, line 248: I don't think these lines are still necessary...

jamis (approved):
❤️ Very nice! Just one bit with some lines that I don't think are necessary any more, but other than that I think it's good to merge. 👍

PR #2850, opened 2024-03-10
3376 csot for sessions

comandeo-mongo:


jamis (approved):
With a possible caveat regarding the discussion we had in slack, about `operation_timeouts` and `inherited_timeout_ms`, I think this looks great. Really good work pushing through the intricacies here!

jamis (approved):
Interesting point, that we might only need the one timeout. For now, though, this all looks great. 👍

comandeo-mongo:
I am merging this as is now, and will take care of the failing tests in the immediate follow-up PR.

PR #2853, opened 2024-03-26
Fix csot test failures

comandeo-mongo:


lib/mongo/socket/ssl.rb
```diff
@@ -184,6 +187,89 @@ def readbyte
...
+          if select_timeout <= 0
+            raise Error::SocketTimeoutError, "The socket took over #{connect_timeout} seconds to connect"
+          end
+          rv = IO.select(nil, [@socket], nil, select_timeout)
+          if rv.nil?
+            raise Error::SocketTimeoutError, "The socket took over #{connect_timeout} seconds to connect"
+          else
+            retry
+          end
+          retry
```
jamis, line 263: Looks like a stray `retry` here. Also, feels like these two `rescue` blocks have a lot in common (differ only on whether the sockets are waiting for read or write). Could the logic be pulled out somewhere?

jamis (changes requested):
Great job, I'm really happy to see the switch to `IO.select` versus `Timeout.timeout`. I'm not adamant about the duplicated code--I can imagine how it might be a little tricky to refactor that out, and if it is, then don't worry about it.

Otherwise, LGTM! 👍

lib/mongo/socket/ssl.rb
comandeo-mongo: Good catch, thank you. Extracted the retry logic in a separate method.

PR #2859, opened 2024-04-11
RUBY-3379 CSOT for bulk writes

comandeo-mongo:


lib/mongo/bulk_write.rb
```diff
@@ -62,8 +62,19 @@ def execute
       validate_requests!
 
       client.send(:with_session, @options) do |session|
```
jamis, line 64: This is honestly beyond the scope of this review, but it's something I've been trying to "fix" as I run into them :) `Client#with_session` is no longer a private method (only `@api private`), so we can change these `client.send(:with_session)` calls into straight `client.with_session` calls.

jamis (approved):

PR #2863, opened 2024-04-19
RUBY-3372 CSOT Cursors

jamis:


lib/mongo/operation/context.rb
```diff
@@ -58,17 +59,42 @@ def initialize(
...
+        self.class.new(client: client,
+                       session: session,
+                       connection_global_id: connection_global_id,
+                       operation_timeouts: operation_timeouts,
+                       view: view || self.view,
+                       options: options)
+      end
+
+      def timeout_ms
+        @operation_timeouts[:inherited_timeout_ms] ||
```
comandeo-mongo, line 95: Should the `operation_timeout_ms` have the higher priority?

comandeo-mongo (approved):

lib/mongo/operation/context.rb
jamis: good catch! 👍

jamis: Actually, I don't think that method is even used anywhere...I think it was an artifact from an earlier experiment. I'll remove it.

PR #2868, opened 2024-04-25
RUBY-3375 CSOT for CSFLE

comandeo-mongo:


jamis (changes requested):
Really great! Just noted one bit of duplicated code that could be split out, but otherwise, 💯

PR #2871, opened 2024-05-13
RUBY-3379 CSOT for bulk writes

comandeo-mongo:


spec/runners/unified/ambiguous_operations.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Unified
+  module AmbiguousOperations
```
jamis, line 4: Nice. Elegant solution to an inelegant problem. 👍

spec/runners/unified/test.rb
```diff
@@ -413,6 +415,7 @@ def execute_operation(op)
           rescue Mongo::Error, bson_error, Mongo::Auth::Unauthorized, ArgumentError => e
             if expected_error.use('isTimeoutError')
               unless Mongo::Error::TimeoutError === e
+                raise e
```
jamis, line 418: Looks like this might be a debugging artifact? 👀

jamis (approved):
I love this refactoring -- splitting the timeout logic to a superclass is a great idea. 🎉

PR #2872, opened 2024-05-22
Sync CSOT spec tests

comandeo-mongo:


spec/lite_spec_helper.rb
```diff
@@ -168,6 +168,9 @@ def require_atlas
     end
   end
 
+=begin
```
jamis, line 171: What's the reasoning behind commenting this out, versus just removing it? We can always pull it from the git history if we decide we need it back.

jamis (approved):

PR #2875, opened 2024-06-10
RUBY-3488: Add empty SBOM lite file

alcaeus:
This PR adds an empty SBOM lite file so the product can be onboarded to Silk. This is necessary to then download the full SBOM upon releasing a new version.

Note: please advise if you'd rather have this file live in a different folder.


jamis (approved):
Thank you! 🙇

PR #2878, opened 2024-06-20
[RUBY-3496] Fix legacy read pool retry error

joelim41:
RUBY-3496

### DIFF
--- a/lib/mongo/retryable/base_worker.rb
+++ b/lib/mongo/retryable/base_worker.rb
@@ -59,7 +59,7 @@ module Mongo
           Error::ServerNotUsable,
           Error::SocketError,
           Error::SocketTimeoutError,
-          Error::PoolError,
+          Error::PoolClearedError,
         ].freeze
       end
 
--- a/lib/mongo/retryable/base_worker.rb
+++ b/lib/mongo/retryable/base_worker.rb
@@ -60,6 +60,7 @@ module Mongo
           Error::SocketError,
           Error::SocketTimeoutError,
           Error::PoolClearedError,
+          Error::PoolPausedError,
         ].freeze
       end
 
--- a/lib/mongo/retryable/base_worker.rb
+++ b/lib/mongo/retryable/base_worker.rb
@@ -49,7 +49,7 @@ module Mongo
 
       private
 
-      # Indicate which exception classes that are generally retryable. 
+      # Indicate which exception classes that are generally retryable.
       #
       # @return [ Array<Mongo:Error> ] Array of exception classes that are
       #   considered retryable.
@@ -58,7 +58,8 @@ module Mongo
           Error::ConnectionPerished,
           Error::ServerNotUsable,
           Error::SocketError,
-          Error::SocketTimeoutError
+          Error::SocketTimeoutError,
+          Error::PoolError,
         ].freeze
       end
 
--- a/lib/mongo/retryable/base_worker.rb
+++ b/lib/mongo/retryable/base_worker.rb
@@ -49,11 +49,26 @@ module Mongo
 
       private
 
-      # Indicate which exception classes that are generally retryable.
+      # Indicate which exception classes that are generally retryable
+      # when using modern retries mechanism.
       #
       # @return [ Array<Mongo:Error> ] Array of exception classes that are
       #   considered retryable.
       def retryable_exceptions
+        [
+          Error::ConnectionPerished,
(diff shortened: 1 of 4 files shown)


lib/mongo/retryable/base_worker.rb
```diff
@@ -58,7 +58,8 @@ def retryable_exceptions
           Error::ConnectionPerished,
           Error::ServerNotUsable,
           Error::SocketError,
-          Error::SocketTimeoutError
+          Error::SocketTimeoutError,
+          Error::PoolError,
```
comandeo-mongo, line 62:
```suggestion
          Error::PoolClearedError,
```
According to the [specification](https://github.com/mongodb/specifications/blob/663db4db13c8acf415bb97259bf7f1909b32a670/source/retryable-reads/retryable-reads.md#retryable-error) only the `PoolClearedError` should be retried, not all pool errors. So, having `PoolError` here seems to be too broad.

comandeo-mongo (changes requested):
Hi @joelim41 , thank you very much for your contribution! I have just one remark, everything else looks good!

lib/mongo/retryable/base_worker.rb
joelim41: @comandeo-mongo `Error::ServerNotUsable` and `Error::ConnectionPerished` are listed here as retryable but not in the specification, could you clarify if they are retryable? Asking because we started seeing these errors after upgrading to the latest.

spec/integration/retryable_reads_errors_spec.rb
```diff
@@ -73,31 +73,42 @@
...
-      expect(find_events.length).to eq(3)
+    end
+
+    it_behaves_like 'retries on PoolClearedError'
+
+    context 'legacy read retries' do
+
+      let(:client) { authorized_client.with(options.merge(retry_reads: false, max_read_retries: 1)) }
+
+      it_behaves_like 'retries on PoolClearedError'
```
joelim41, line 111: test is failing because it's throwing `PoolPausedError`, any idea how to make it throw `PoolClearedError` instead?

lib/mongo/retryable/base_worker.rb
joelim41: hmm, the [code](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/retryable/read_worker.rb#L216) currently seems to retry on any pool errors

lib/mongo/retryable/base_worker.rb
```diff
@@ -58,7 +58,9 @@ def retryable_exceptions
           Error::ConnectionPerished,
           Error::ServerNotUsable,
           Error::SocketError,
-          Error::SocketTimeoutError
+          Error::SocketTimeoutError,
+          Error::PoolClearedError,
+          Error::PoolPausedError,
```
joelim41, line 63: [PoolPausedError](https://github.com/joelim41/mongo-ruby-driver/blob/master/lib/mongo/error/pool_paused_error.rb#L23) looks retryable too

lib/mongo/retryable/base_worker.rb
joelim41: @comandeo-mongo I've addressed your comment, please look at it when you have the time.

comandeo-mongo: Hi @joelim41 , I am finally looking into it, sorry for the delay. Meanwhile, I would like to clarify - did you start seeing `Error::ServerNotUsable` and `Error::ConnectionPerished` only after the upgrade? It might be an indication of some other issues.

comandeo-mongo: @joelim41 one more "side" question - is there something tpreventing you from using modern retries? Legacy retry mechanism is deprecates since `2.9.0`

comandeo-mongo: Oh, I missed that you are upgrading from 2.16. So yes, you weren't able to see those errors. I think it is OK if you see them when a mongos has issues.

As to this PR - I think we should separate the list of the retryable errors for modern and legacy retries. `PoolError` has a special treatment in modern retries, so I would like to avoid just adding it to `retryable_exceptions`. I'll make a commit to your PR with the proposed solution.

joelim41: Got it, thank you!

comandeo-mongo (approved):

jamis (approved):

PR #2884, opened 2024-07-04
RUBY-3378 Document CSOT

comandeo-mongo:


lib/mongo/client.rb
```diff
@@ -942,8 +942,9 @@ def reconnect
     # @option opts [ Session ] :session The session to use.
     # @option opts [ Object ] :comment A user-provided
     #   comment to attach to this command.
-    # @option opts [ Integer | nil ] :timeout_ms Operation timeout in milliseconds.
-    #    Must a positive integer. The default value is unset which means infinite.
+    # @option options [ Integer ] :timeout_ms The operation timeout in milliseconds.
+    #    Must a non-negative integer. An explicit value of 0 means infinite.
```
alexbevi, line 946:
```suggestion
    # @option options [ Integer ] :timeout_ms The operation timeout in milliseconds.
    #    Must be a non-negative integer. An explicit value of 0 means infinite.
```

lib/mongo/client.rb
```diff
@@ -966,8 +967,9 @@ def database_names(filter = {}, opts = {})
     #   is enabled.
     # @option opts [ Object ] :comment A user-provided
     #   comment to attach to this command.
-    # @option opts [ Integer | nil ] :timeout_ms Operation timeout in milliseconds.
-    #    Must a positive integer. The default value is unset which means infinite.
+    # @option options [ Integer ] :timeout_ms The operation timeout in milliseconds.
+    #    Must a non-negative integer. An explicit value of 0 means infinite.
```
alexbevi, line 971:
```suggestion
    #    Must be a non-negative integer. An explicit value of 0 means infinite.
```

lib/mongo/client.rb
alexbevi: Can you change "Must a non-negative" to "Must be a non-negative" throughout this PR?

comandeo-mongo: Did this, good catch, thanks!

docs/reference/create-client.txt
```diff
@@ -1239,16 +1255,34 @@ round trips.
...
+The per-operation timeout value in milliseconds. The option specifies the
+best-effort maximum amount of time a single operation can take before control
+is returned to the application. This option is configurable at the level of
+a ``Mongo::Client``, database, collection, or of a single operation. Not set by
+default. An explicit value of 0 means infinite.
+
+If this option is set and the timeout expires, the driver aborts all blocking
+work and return control to the user with an error ``Mongo::Error::TimeoutError``.
+
+If this option is set, the driver ignore any deprecated timeout option -
```
jamis, line 1279:
```suggestion
If this option is set, the driver will ignore any deprecated timeout option -
```

jamis (approved):
Just one minor correction, otherwise LGTM! 👍

PR #2888, opened 2024-07-15
RUBY-3514 pointer to docs

adviti-mishra:
Modified the README.md to indicate the docs folder that was in the Ruby Driver can now be found in docs-ruby

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ To build API documentation for the master branch, check out the
 repository locally and run `rake docs`.
 
 High-level driver documentation including tutorials and the reference that were in the docs folder can now be found
-[here](https://github.com/mongodb/docs-ruby/tree/master/source)
+at the docs-ruby repository, [here](https://github.com/mongodb/docs-ruby)
 
 ## Support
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -53,6 +53,9 @@ API documentation for the most recent release can be found
 To build API documentation for the master branch, check out the
 repository locally and run `rake docs`.
 
+High-level driver documentation including tutorials and the reference that were in the docs folder can now be found
+[here](https://github.com/mongodb/docs-ruby/tree/master/source)
```
jamis, line 57: Probably personal preference, but I think it's good practice to describe what a link points to, so readers (and especially readers who are looking at the formatted text, rather than the raw markdown) can see where the link will take them.

Also, I think it's sufficient to link directly to the repository.

```suggestion
at the docs-ruby repository, [here](https://github.com/mongodb/docs-ruby)
```

jamis (changes requested):
Just the one comment -- feel free to push back if you feel strongly about it. :)

jamis (approved):

PR #2896, opened 2024-10-01
RUBY-3543 Run lambda tests on cloud-dev

alcaeus:
RUBY-3543

Evergreen config values have been updated, but the new `DRIVERS_ATLAS_BASE_URL` variable needs to be passed on to each script creating a cluster for Lambda.


comandeo-mongo (approved):
Test failures are irrelevant to this change and will be fixed in another PR.

PR #2916, opened 2025-02-04
Support the removal for Fixnum from ruby

onerinas:
Upgrading from ruby 3.1.3 to Ruby 3.2.x raises error `NameError: uninitialized constant Mongo::Support::Fixnum`

Used 1.class similar to how its used here: https://github.com/rails/rails/pull/27458/files#diff-5a9ceb7949cd32ce2be8451ae0011d4640a2a642b6bf2e5a1eb46c9bba918673L24

Unfortunately we cannot upgrade the gem to 2.x until we upgrade the database version.

### DIFF
--- a/lib/mongo/utils/support.rb
+++ b/lib/mongo/utils/support.rb
@@ -45,7 +45,7 @@ module Mongo
 
     def normalize_seeds(seeds)
       pairs = Array(seeds)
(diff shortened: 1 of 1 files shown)


onerinas:
Closing since specs fail and moving ahead with monkey patching in code rather than in older version of gem.

PR #2917, opened 2025-02-10
Add session escape functionality

johnnyshields:


johnnyshields:
@jamis @comandeo-mongo I've raised 3 semi-related PRs:
- https://github.com/mongodb/mongo-ruby-driver/pull/2917
- https://github.com/mongodb/mongo-ruby-driver/pull/2918 
- https://github.com/mongodb/mongo-ruby-driver/pull/2919 (cleanup)

As our app has grown in complexity, it is important to have greater ability to control query execution per session. Can you please review these changes and let me know if the approach is acceptable, or if you envision a different way of doing things?

Importantly, #2917 and #2918 may appear similar on the surface, but they actually address two different use cases. Our app would benefit from having them both.

Once you have reviewed, I will write the tests.

PR #2944, opened 2025-08-06
RUBY-3694 Use correct CA when verifying OCSP endpoint 

jamis:
The driver was incorrectly using the last certificate in the chain as the CA when performing OCSP verification. This worked for the case where the chain had only two elements, but for longer chains, it was causing the verification to fail.

This change searches the cert chain for the issuer of the peer certificate, and uses that as the CA for the verification request.

### DIFF
--- a/lib/mongo/socket/ssl.rb
+++ b/lib/mongo/socket/ssl.rb
@@ -505,7 +505,6 @@ module Mongo
       end
 
       # Find the issuer certificate in the chain.
-      # If the issuer is not found, raise an error.
       def find_issuer(cert, cert_chain)
         cert_chain.find { |c| c.subject == cert.issuer }
       end


lib/mongo/socket/ssl.rb
```diff
@@ -460,7 +460,7 @@ def verify_ocsp_endpoint!(socket, timeout = nil)
         end
 
         cert = socket.peer_cert
-        ca_cert = socket.peer_cert_chain.last
+        ca_cert = socket.peer_cert_chain[1]
```
alcaeus: Can we rely on this certificate always working or should we fall back to check all other certificates as well?

alexbevi (approved):
LGTM

lib/mongo/socket/ssl.rb
jamis: There are two points for why I think this is correct (but I am absolutely not well-versed in SSL lore, and may have this all wrong).

- Per the [Suggested OCSP Behavior section](https://github.com/mongodb/specifications/blob/5b15fd307b7ec06dbafc57e3c252f7ae3b8c4e46/source/ocsp-support/ocsp-support.md#suggested-ocsp-behavior) of our OCSP spec, item 8 says "If any unvalidated intermediate certificates remain and those certificates have OCSP endpoints, for each certificate, the driver SHOULD NOT reach out to the OCSP endpoint specified and attempt to validate that certificate." So, we don't try to validate the OCSP endpoints of any intermediate certs. (Not sure if that's what you were referring to, but figured I'd be thorough.)
- Given that we saw the warnings/errors happening because we were trying to validate using the last item in the chain, I'm assuming that using subsequent certificates to validate here would be similarly problematic. Or, is the correct behavior that we're supposed to try each certificate in the chain until one of them gives the green light, ignoring error messages along the way?

alexbevi (approved):
approach LGTM

PR #2948, opened 2025-08-13
RUBY-3361 Test x509 authentication on Atlas

comandeo-mongo:


alcaeus (approved):
Code LGTM from what I can see. Happy to merge if the connectivity tests pass.

PR #3033, opened 2026-04-28
RUBY-3824 Restore conditional skip logic in transactions spec runner

comandeo-mongo:
A rubocop autocorrect in commit
[8db64312b](https://github.com/mongodb/mongo-ruby-driver/commit/8db64312b8ca53d2e564e59250c079a38baf3ad3)
(\"Enable rubocop on all files, with exceptions\" — #3005) flattened
the original `if test.skip_reason` and `unless req.satisfied?` guards
in `spec/runners/transactions.rb` into bare `skip` calls inside a
single `before(:all)` block:

```ruby
before(:all) do
  ...
  skip test.skip_reason            # always skips, even when nil
  skip 'Requirements not satisfied' # always skips
  test.setup_test                   # never reached
end
```

Effect: every `transactions_spec.rb` and `transactions_api_spec.rb`
example reports as `PENDING` regardless of topology or actual
requirements. The transactions spec test suite has not run locally on
master since that commit (2026-03-27).

The fix restores the conditional checks and passes `req` through to
the block (it was renamed to `_req` because it had become unused after
the regression).

## Test plan

Against a local 3-node replica set (`replicaSet=replset`):

- [x] `bundle exec rspec spec/spec_tests/transactions_spec.rb` — 1809 examples, 0 failures, 965 pending (sharded-only). Before: all pending.
- [x] `bundle exec rspec spec/spec_tests/transactions_api_spec.rb` — 370 examples, 0 failures, 185 pending (sharded-only). Before: all pending.
- [x] `bundle exec rubocop spec/runners/transactions.rb` — clean

### DIFF
--- a/lib/mongo/session.rb
+++ b/lib/mongo/session.rb
@@ -936,6 +936,13 @@ module Mongo
     #
     # @api private
     def unpin(connection = nil)
+      # Idempotent: if there is no pinned state to clear, do nothing. Nested
+      # unpin_maybe handlers (e.g. in BulkWrite#execute_operation wrapping an
+      # OpMsg execution that already calls unpin_maybe in its own do_execute)
+      # can call this method twice for the same error; checking the connection
+      # back into the pool a second time would raise from the pool.
+      return if @pinned_server.nil? && @pinned_connection.nil? && @pinned_connection_global_id.nil?
+
       @pinned_server = nil
       @pinned_connection_global_id = nil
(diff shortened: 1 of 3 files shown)


jamis (approved):
Great catch! I was afraid something like this might slip through the cracks, after I discovered that the supposedly "safe" Rubocop autocorrect mode (`-a`) is not actually guaranteed to be safe. Thanks for fixing this. 👍

